Issue #44, opened 2021-10-25
Readability archives contain â€


lrnselfreliance:
The characters â€ (or some variation) are present in some Readability documents.  This appears to be an encoding problem.

PR #61, opened 2021-11-12
Using charset utf-8 and including mime.types

lrnselfreliance:
Using charset utf-8 and including mime.types so HTML files and images are displayed correctly.


lrnselfreliance:
This is a fix for #44

PR #156, opened 2023-02-02
Upgrading to Node 18.

lrnselfreliance:
Upgrading app docker container to use Node 18.  This requires the removal of several npm packages.  File browser will require re-write.  It now pulls fonts from Google, this will need to be fixed.


lrnselfreliance:
Merged https://github.com/lrnselfreliance/wrolpi/commit/81d80238bed05130272b90ec49816b0da83b69d6

PR #251, opened 2024-09-09
Feature/auto background task

lrnselfreliance:
Slow tasks (like writing config files) are automatically run in background using `@auto_background_task`


lrnselfreliance:
This is abandoned in favor of `wrolpi.switches`

PR #496, opened 2026-06-22
PWA: installable app, offline calculators, read-only Inventory

lrnselfreliance:
## Summary

Makes the React app an **installable PWA** whose app shell, **calculators** (including the One Time Pad), and **Inventory** work fully **offline**. Driving use case: reference an inventory (e.g. Fasteners) from an iPhone in a store aisle with no connection.

No backend changes. The service worker only emits in a production build, so `npm start` dev is unaffected.

## What's included

**Phase 1 — Service worker** (Workbox InjectManifest, production build only)
- `service-worker.js`: precache the shell (calculators work offline), `NavigationRoute` for react-router deep links, **NetworkFirst** cache of `GET /api/inventory` + `/api/inventory/catalog` (`wrolpi-inventory`), **NetworkOnly** `/api/status` (so the apiDown indicator stays truthful).
- `serviceWorkerRegistration.js` + registration in `index.js`.
- `workbox-*` runtime deps pinned to **6.5.4** to match CRA's bundled plugin.

**Phase 2 — Offline hardening + read-only Inventory**
- `api.js` maps offline `fetch` failures (TypeError / timeout / `!navigator.onLine`) → `ApiDownError`, so the existing nav apiDown icon lights up.
- `toast.js` wrapper swallows **error** toasts while offline; the `toast` import is repointed in the **13** files that raise them (the `App.js` container is left on the original package).
- `PwaContext` (`usePwa()` → `{offline, isStandalone}`), provider mounted in `App.js` inside `StatusProvider`. Default `offline: false` so components render without the provider in unit tests.
- Inventory goes **read-only offline**: `InventoryRoute` disables the edit buttons and shows a banner; `InventoryTable` hides the new-item row, checkboxes and delete, and disables click-to-edit. Item values stay visible for reference. Writes are simply never issued (no API write-guards needed).

**Phase 3 — iOS** Add-to-Home-Screen meta tags in `public/index.html`.

## Testing

- **676 Jest tests pass**, including 2 new `InventoryTable` offline tests.
- **Verified end-to-end against the real API** (production build served via Caddy at `:8443`): the SW caches the real Fasteners inventory, then with the origin stopped the cached inventory still renders **read-only** — both reactively (no reload) and on a cold offline reload.

## Deferred (not in this PR)

Media "Save for offline" (Cache Storage + IndexedDB, iOS Range/206, quota), a Downloads list, and offline inventory **edits**/sync.

## Remaining manual verification

Definitive on-device test: Add to Home Screen on an iPhone against `10.0.0.5` — requires the WROLPi Root CA trusted on the device or the SW won't register.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -33,7 +33,12 @@
         "semantic-ui-offline": "^2.5.0",
         "semantic-ui-react": "^2.1.5",
         "three": "^0.184",
-        "web-vitals": "^5.3.0"
+        "web-vitals": "^5.3.0",
+        "workbox-cacheable-response": "^6.5.4",
+        "workbox-core": "^6.5.4",
+        "workbox-precaching": "^6.5.4",
+        "workbox-routing": "^6.5.4",
+        "workbox-strategies": "^6.5.4"
       },
       "devDependencies": {
         "@testing-library/jest-dom": "^6.9.1",
@@ -106,6 +111,7 @@
       "version": "7.20.12",
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.20.12.tgz",
       "integrity": "sha512-XsMfHovsUYHFMdrIHkZphTN/2Hzzi78R08NuHfDBehym2VsPDL6Zn/JAD/JQdnRvbSsbQc4mVaU1m6JgtTEElg==",
+      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.1.0",
         "@babel/code-frame": "^7.18.6",
@@ -902,6 +908,7 @@
       "version": "7.18.6",
       "resolved": "https://registry.npmjs.org/@babel/plugin-syntax-flow/-/plugin-syntax-flow-7.18.6.tgz",
       "integrity": "sha512-LUbR+KNTBWCUAqRG9ex5Gnzu2IOkt8jRJbHHXFT9q+L9zm7M/QQbEqXyw1n1pohYvOyWC8CjeyjrSaIwiYjK7A==",
+      "peer": true,
       "dependencies": {
         "@babel/helper-plugin-utils": "^7.18.6"
       },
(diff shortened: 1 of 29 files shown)


lrnselfreliance:
## Follow-up: Docker deployment served the dev server (no service worker)

On-device testing surfaced that the **Docker** deployment ran the CRA dev server (`npm run start`), which never emits the Workbox service worker — so the installed PWA had no offline cache (infinite spinner offline). Native (`wrolpi-app.service` → `serve build`) was already fine.

Fixed in `f87921e2`:
- `docker/app/Dockerfile` now builds the bundle and serves it (`serve build -c /app/serve.json`), matching native.
- `docker-compose.yml` base now serves the **production build** (what end users get).
- New `docker-compose.dev.yml` opt-in override restores the dev server + hot reload for local development.
- `docker/README.md` documents dev vs. production run commands.

Verified locally through Caddy: `/service-worker.js` now returns real `application/javascript` (was the SPA `index.html` fallback); SW registers, caches inventory, and serves it offline.

lrnselfreliance:
## Follow-up: iOS first-install offline (infinite spinner)

On a fresh iOS *Add to Home Screen* install, the inventory still spun offline even though the shell loaded. Root cause (confirmed via research + local repro):

- The just-installed service worker isn't controlling the page yet when React first fetches `/api/inventory`, so the NetworkFirst route never stores it → offline = empty cache → spinner.
- iOS compounds this: Safari and the standalone PWA have **isolated storage**, so priming in Safari doesn't help the installed app; the standalone app must prime its own cache while open online.

Fix (`201df63e`): the service worker now **warms the inventory cache in its `activate` handler** (`cache.add(new Request('/api/inventory', {cache:'reload'}))` + catalog), so the first online launch guarantees an offline copy regardless of fetch/control timing.

Verified locally: a clean first load caches the inventory with no manual reload; cold offline reload renders the inventory read-only instead of spinning. Deployed and confirmed served on the QA device.

lrnselfreliance:
Requires too much fiddling, for not enough benifit.

PR #520, opened 2026-07-03
Global search: fast ABC mode by default, opt-in deep search

lrnselfreliance:
## Motivation

`file_group.textsearch` indexes all four weighted text columns into one tsvector. On a large library (156K file_groups), d_text (captions, document text, zip contents) is ~90% of the text data, producing a ~1 GB GIN index and ~3.3s queries — `ts_rank` must detoast every matching tsvector. 60–90% of matches come only from d_text and are low-relevance.

## Changes

**Database** (migration `e5f7a9c1d3b6`)
- New `textsearch_abc` generated column: title/author/description only (weights A/B/C), ~8x smaller than `textsearch`, plus a GIN index. Adds ~270 MB, existing column unchanged.
- Note: adding a STORED generated column rewrites the table — a few minutes on large libraries.

**Backend**
- `FilesSearchRequest.deep: bool = False`; `search_files()` picks `textsearch_abc` (default) or `textsearch` (deep).
- `/api/search_file_estimates` returns both counts from one index-bound query (`COUNT(*) FILTER` on the abc subset over the deep superset): `file_groups` (fast) and `file_groups_deep`. Verified the estimate never reads tsvector values (bitmap index scan, `width=0`): ~23ms vs ~211ms+ for the ranked query at 18K matches.

**Frontend** — one coherent story across the three count surfaces:
- The typing **dropdown** always suggests the deep count — what's *findable*.
- The **tab badge** shows the active mode's count, so it always matches the result list.
- The **hint** reconciles the difference: "No results in fast search. N results available in deep search." (or "N more results available…") with a one-click "Try deep search" button.
- "Search Depth" toggle in the search filter modal (censored-filter idiom), persisted as `?deep=true`.

**Tests**
- New: `test_files_search_deep` (fast vs deep results, ranking, headlines, both estimate counts), deep-toggle cases in `SearchFilterModal.test.js`.
- Updated: tests that relied on d_text matches (archive contents, captions, zip contents) now pass `deep=True`; estimate tests assert `file_groups_deep`.

**Follow-up (out of scope):** videos/archives/docs module searches still rank on the full `textsearch` column; they'll adopt the same pattern next.

## Verification

- `pytest -nauto`: 1550 passed. Jest: 729 passed. Cypress: 43/43 against the docker stack.
- `alembic check` clean; migration downgrade/upgrade round-trip tested.
- Manually verified all three states against the docker stack (fast-zero + hint, fast-partial + "N more" hint, deep-on) and the dropdown→search→deep click-through flow.

### DIFF
--- a/alembic/versions/e5f7a9c1d3b6_file_group_textsearch_abc.py
+++ b/alembic/versions/e5f7a9c1d3b6_file_group_textsearch_abc.py
@@ -1,0 +1,43 @@
+"""file_group_textsearch_abc
+
+Add the `textsearch_abc` generated tsvector column to `file_group`.  It indexes only a_text (title),
+b_text (author), and c_text (description) — excluding d_text (captions/document body), which is ~90% of all
+text data.  Searches default to this much smaller column; the existing `textsearch` column remains for
+opt-in "deep" search.
+
+Adding a STORED generated column rewrites the table; expect a few minutes on large libraries.
+
+Revision ID: e5f7a9c1d3b6
+Revises: d4e6f8a0c2b5
+Create Date: 2026-07-03
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = 'e5f7a9c1d3b6'
+down_revision = 'd4e6f8a0c2b5'
+branch_labels = None
+depends_on = None
+
+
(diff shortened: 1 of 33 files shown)


lrnselfreliance:
@greptile

PR #543, opened 2026-07-16
videos: center YouTube auto-generated captions

lrnselfreliance:
## Problem

Captions always render at the **bottom-left** of the video player, making them hard to read.

YouTube auto-generated captions (downloaded via yt-dlp) stamp **every** cue with `align:start position:0%`:

```
00:00:00.400 --> 00:00:02.230 align:start position:0%
```

That cue setting pins the text to the left edge. Properly-authored VTT files have no cue settings and already render centered by default. Native `::cue` CSS **cannot** override cue position/align — those are per-cue properties, not stylable.

## Fix

Purely client-side, no VTT rewriting:

- **`centerAutoCaptions(track)`** — scans a track's parsed cues and, for only those matching the auto-caption signature (`align === 'start' && position === 0`), resets them to the browser's centered default (`align:center / position:auto / line:auto`). Cues deliberately positioned any other way are left untouched.
- **`handleCaptionTrackLoad`** — wired to each `<track>`'s `onLoad`. The default English track has cues ready at load; non-default language tracks are `disabled` (no cues parsed yet), so they are normalized lazily on the first `cuechange` when the user selects them.

This mutates only the in-memory `VTTCue` objects the browser parsed — **the `.vtt` file on disk is never modified**, and no backend/Docker changes are needed (captions are served statically).

## Verification

- **Live, against the real player** (video 13 on a QA RPi): running the mutation normalized 707 cues and the captions moved from bottom-left to centered.
- **Jest**: new `VideoPlayer.test.js` — 4 passing tests covering the auto-caption case, mixed tracks, preserved intentional positioning, and null-cue tracks.

### DIFF
--- a/app/src/components/VideoPlayer.js
+++ b/app/src/components/VideoPlayer.js
@@ -37,17 +37,18 @@ const MEDIA_PATH = '/media';
 // which pins the text to the bottom-left of the player and makes it hard to read.  Detect that signature and
 // reset those cues to the browser's centered default.  This mutates only the in-memory VTTCue objects the
 // browser parsed; the .vtt file on disk is never modified.  Cues that were deliberately positioned (anything
-// other than the auto-caption signature) are left untouched.
+// other than the auto-caption signature) are left untouched.  YouTube auto-captions never set `line`, so we
+// require `line === 'auto'` too; that leaves a deliberately placed cue (e.g. `line:0` for top-edge) alone even
+// if it happens to also be `align:start position:0%`.
 export function centerAutoCaptions(track) {
     if (!track || !track.cues) {
         return 0;
     }
     let changed = 0;
     for (const cue of track.cues) {
-        if (cue.align === 'start' && cue.position === 0) {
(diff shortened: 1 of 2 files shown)


lrnselfreliance:
Thanks @greptile-apps — triaged both:

**P1 (Explicit Cue Line Gets Reset) — fixed** in 96b8c1de. Added `cue.line === 'auto'` to the signature. YouTube auto-captions never set `line`, so this leaves a deliberately placed cue (e.g. `line:0`) alone even if it's also `align:start position:0%`. Added a regression test.

**P2 (First Cue Change Ends Normalization) — hardened.** In practice the browser parses a local static WebVTT file in full when the track leaves `disabled`, so `track.cues` is already complete at the first `cuechange` and `centerAutoCaptions` iterates the whole list (not just active cues). The one real gap was detaching too early if a `cuechange` somehow fires before cues load — the listener now only removes itself once `track.cues.length > 0`.

PR #544, opened 2026-07-16
videos: center YouTube auto-captions at download time

lrnselfreliance:
## Problem

YouTube auto-generated captions (downloaded via yt-dlp) stamp **every** cue timing line with `align:start position:0%`, which pins the on-screen text to the **bottom-left** of the player and makes it hard to read:

```
00:00:00.400 --> 00:00:02.230 align:start position:0%
```

Across the entire test corpus, **every** cue setting is that exact string (23k+ occurrences) — there is no legitimate positioning variety in YouTube captions.

## Fix

Strip that exact cue setting from the downloaded `.vtt` files so browsers render the captions centered by default.

- **`strip_youtube_caption_positioning()`** in `wrolpi/captions.py` — a targeted text replacement (matched to the exact `align:start position:0%` signature on a cue timing line), *not* a `webvtt` parse→`.save()` round-trip, so the `<c>` word-timing tags inside the cue text are preserved byte-for-byte. Idempotent; rewrites in place only when something changed. Deliberately positioned cues (e.g. ones that also carry a `line` setting) are left untouched.
- Called from `VideoDownloader` right after `normalize_video_file_names`, **gated on a `youtube.com` / `youtu.be` source URL**. It runs *before* the `FileGroup` is created, so the first index reads the corrected file — no re-index plumbing needed. Re-downloads (which WROLPi does to refresh metadata) just re-strip harmlessly.

## Relationship to #543

Complementary — keep both:

- **#543 (frontend)** centers captions in WROLPi's own player for **any source and existing/already-downloaded** files, purely in-memory.
- **This PR (backend)** makes **new YouTube downloads** correct at the source, so the files themselves are right for any consumer and we don't rely solely on a client-side mutation.

No retroactive migration of existing files (those are covered by #543).

## Tests

`wrolpi/test/test_captions.py` — new cases: strips the signature + preserves timestamps/`<c>` tags/parseability, is idempotent, leaves clean (no-setting) files untouched, and preserves a deliberately `line`-positioned cue while still centering a plain auto-caption cue. Full `test_captions.py` suite passes (7).

### DIFF
--- a/modules/videos/downloader.py
+++ b/modules/videos/downloader.py
@@ -21,6 +21,7 @@ from sqlalchemy.orm import Session
 from yt_dlp import YoutubeDL, DownloadError
 from yt_dlp.extractor import YoutubeTabIE  # noqa
 
+from wrolpi.captions import strip_youtube_caption_positioning
 from wrolpi.cmd import YT_DLP_BIN
 from wrolpi.common import logger, get_media_directory, escape_file_name, resolve_generators, background_task, \
     trim_file_name, cached_multiprocessing_result, get_absolute_media_path, aiohttp_post, normalize_domain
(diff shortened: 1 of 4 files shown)


lrnselfreliance:
Thanks @greptile-apps — both valid, both fixed in 24b935d5:

**P1 (Failed Write Corrupts Caption File) — fixed.** `strip_youtube_caption_positioning` now writes to a temp file in the same directory and `os.replace()`s the original (atomic on the same filesystem), cleaning up the temp on failure. A partial/failed write can no longer truncate the existing caption.

**P2 (Substring Gate Misclassifies) — fixed.** Replaced `'youtube.com' in url` with `is_youtube_url()`, which parses the hostname via `urlparse` and matches `youtube.com` / `youtu.be` and their subdomains, rejecting lookalikes like `youtube.com.example.org` and `notyoutube.com`. Added a parametrized test covering those cases.

PR #550, opened 2026-07-18
upgrade: hard-cutoff Debian 12 at the debian12-final tag

lrnselfreliance:
## Why

Debian 12 (bookworm) cannot run the Postgres→SQLite move or the Debian 13 rebase, and there is **no supported in-place 12→13 upgrade**. Without a guard, a Debian 12 box tracking `release` would eventually pull SQLite/Debian 13 code and break mid-upgrade (services are already stopped by that point). This pins Debian 12 to a final supported commit and tells the user to reinstall on Debian 13.

## How it works

The guard lives at the top of `scripts/upgrade.sh` (the outer `upgrade.sh` is unchanged — it still stops services, fetches, and checks out the latest release). By the time `scripts/upgrade.sh` runs, the latest release is already checked out, so the guard runs before any dependency work:

- **Debian 13+** → no-op; normal upgrade proceeds.
- **Debian 12, HEAD past the cutoff** → verify the `debian12-final` commit signature → `git checkout -f -B release debian12-final` → `exec` that version's script (which warns and runs its normal upgrade, so services restart via `repair.sh`).
- **Debian 12, already at `debian12-final`** → print the "reinstall on Debian 13" warning, fall through to the normal upgrade.
- **Tag missing or unsigned** → abort (exit 10/11); never runs newer code on Debian 12.

`scripts/upgrade.sh` is deliberately the enforcement point because it is the one file that runs the *new* version even for users still on the *old* `upgrade.sh` — the only hook into an existing box's transition upgrade.

## Testing

Sandbox harness (throwaway git repo + dedicated GPG signing key, exercising the **real** guard block via path substitution), 14/14 assertions pass:
- Debian 13 pass-through (stays on latest, no warning)
- Debian 12 at-tag: warn + proceed, stays on tag
- Debian 12 past-tag: verify → rollback → re-exec → settle on the tag, **no infinite loop**
- Missing/unreachable tag: refuses with exit 10

The harness caught a real bug: `git rev-parse "tag^{commit}"` echoes the unresolved arg on failure, so a missing tag skipped the clean abort path — fixed with `--verify --quiet`.

## ⚠️ Required follow-up before this reaches `release`

1. **Create + sign-tag `debian12-final`** on the last Postgres/Debian 12 release commit (`origin/release` + this guard). Until the tag exists, a Debian 12 box hitting this code refuses with exit 10 (safe, but not the intended pin).
2. **Sequencing:** the guard **and** the tag must land on `release` *before* any SQLite/Debian 13 code is promoted to `release`, or an interim upgrade strands Debian 12 users.
3. `debian12-final` is off the master/release line (Postgres vs SQLite diverge at the cutover), so the guard fetches the tag explicitly (`git fetch origin tag debian12-final`).

## Verification plan

On a real Debian 12 box (10.0.0.8), point `upgrade.sh -b feature/debian12-upgrade-cutoff` and confirm it refuses to advance to latest master. Note: **without the tag** you'll see the exit-10 refusal (services left stopped, since it aborts before `repair.sh`); **with the tag** you'll see the full rollback-to-`debian12-final` + restart.

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -12,6 +12,110 @@ cleanup() {
 }
 trap cleanup EXIT
 
+# --- Debian 12 (bookworm) hard cutoff --------------------------------------
+# WROLPi's Debian 12 line ends at the signed tag `debian12-final`.  Everything
+# after it (the Postgres->SQLite move and the Debian 13 rebase) requires a
+# fresh Debian 13 (trixie) system; there is no supported in-place 12 -> 13
+# upgrade.  ./upgrade.sh has already fetched and checked out the latest
+# release by the time we run, so on a Debian 12 machine we pin back to
+# `debian12-final` and never advance past it.  On Debian 13+ this is a no-op
+# and the normal upgrade proceeds.
+GUARD_TAG="debian12-final"
+GUARD_KEY="/opt/wrolpi/wrolpi/roland@learningselfreliance.com.gpg"
+
+# True only on Debian 13 (trixie) or newer.  A missing or non-numeric
+# VERSION_ID is treated as "older" so the frozen line fails safe (pins).
+os_is_debian_13_or_newer() {
+    local vid=""
+    [ -r /etc/os-release ] && vid=$(awk -F= '$1=="VERSION_ID"{gsub(/"/,"",$2);print $2}' /etc/os-release)
+    case "${vid}" in
+        '' | *[!0-9]*) return 1 ;;
+    esac
+    [ "${vid}" -ge 13 ]
+}
+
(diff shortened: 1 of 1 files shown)


lrnselfreliance:
Created and pushed the signed pin tag `debian12-final` → commit `ac69ec10` (`origin/release` 7da89d88 + this guard, i.e. the last Postgres/Debian 12 release). This is the commit Debian 12 machines pin to. Note: only the **tag** is pushed; `origin/release` is unchanged. Moving `origin/release` to `debian12-final` is the deliberate publish step and should happen (before any SQLite/Debian 13 code is promoted to release) once verified.

PR #554, opened 2026-07-19
Speed up tags config import for un-modeled files

lrnselfreliance:
## Problem

`Tags.import_config()` crawled (~a dozen rows/second) when importing hundreds of `tag_files` for files that exist on disk but have **no FileGroup in the DB yet** — e.g. a fresh config import before a file refresh has modeled the files.

In that case every tagged file fell into the on-disk fallback, which per row did:

1. `glob_shared_stem()` → a full **scan of the parent directory** (O(files²) for a large Channel/Domain directory),
2. `get_mimetype()` → a libmagic **byte-read** of the file,
3. `session.flush([file_group])` → a **per-row DB round-trip**.

The other path (FileGroup already in the DB) was already fast — flat ~8 statements regardless of row count — so this only bit fresh/un-modeled imports.

## Fix

- **Directory-listing cache** — each directory is scanned once via `iterdir()` and reused, replacing the per-file glob. O(n²) → O(n), the main win for large directories.
- **Two-pass batch flush** — FileGroups are created under `session.no_autoflush`, then flushed once (`session.flush(new_file_groups)`) instead of once per row; TagFiles are created in a second pass and inserted as a single batch.
- **Path-cache dedup** — a file tagged by multiple tags reuses one FileGroup instead of risking duplicate creation.

The per-file `get_mimetype` read and `from_paths` existence-check remain — they're inherent to modeling a new file (this keeps the create-on-import fallback rather than deferring to a refresh).

## Tests

- New regression test `test_import_tags_config_creates_missing_file_groups_efficiently`: fails on old code (20 per-row inserts), passes on the fix; asserts 10 FileGroups + 20 TagFiles + batched inserts.
- Full backend suite: **1647 passed, 5 skipped**. Tags + config/DR suites green.

### DIFF
--- a/wrolpi/tags.py
+++ b/wrolpi/tags.py
@@ -388,7 +388,7 @@ class TagsConfig(ConfigFile):
     def import_config(self, file: pathlib.Path = None, send_events=False):
         from modules.zim import lib as zim_lib
         from modules.zim.models import Zim, TagZimEntry
-        from wrolpi.files.lib import glob_shared_stem
+        from wrolpi.files.lib import split_path_stem_and_suffix
 
         if PYTEST and not TEST_TAGS_CONFIG:
             logger.warning('Refusing to import tags without test tags config.  '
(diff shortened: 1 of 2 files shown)


lrnselfreliance:
Good catch, @greptile-apps — confirmed and fixed in 646e6965.

I reproduced it with a TDD regression test (`test_import_tags_config_shared_stem_members`): two config entries for the same shared-stem group (`video.mp4` + `video.info.json`) triggered `UNIQUE constraint failed: file_group.primary_path` and aborted the whole import. This was a regression from the pre-PR per-row flush, which made the first group visible to `from_paths`.

Fix caches **every** member path of a newly-created FileGroup (not just the primary) so any later entry for the group reuses it. Test fails before the fix, passes after.

PR #563, opened 2026-07-21
zim: ship a default fallback Zim and honor the configured Zim directory

lrnselfreliance:
## Context

Two related gaps in how WROLPi serves Kiwix content:

1. **No default Zim.** A fresh WROLPi — or one whose Zim files are all corrupt/absent — showed a blank Kiwix with no explanation. Every image now ships a small built-in default Zim that Kiwix falls back to, telling the user to download real content.
2. **The configured Zim directory was ignored by the startup scripts.** `wrolpi.yaml`'s `zims_destination` (honored by the Python side via `get_zim_directory()`) was hardcoded to `/zims` in `scripts/start_kiwix_serve.sh`.

## Changes

**Reusable config reader** — `wrolpi/scripts/read_config_value.sh`: a zero-dependency (`awk`) reader so non-Python scripts can reliably read top-level `wrolpi.yaml` scalars. `jq`/`yq` were rejected (JSON-only / packaging/dep cost); PyYAML would have added deps to the minimal Zim image. Plus a `config_value()` wrapper in `wrolpi/scripts/lib.sh`.

**Default fallback Zim** — `wrolpi/scripts/default_zim/`:
- `src/{index.html,favicon.png}` — a friendly "you have no Zim files, download some" page.
- `build_default_zim.sh` — reproducible build via `ghcr.io/openzim/zim-tools`, output committed to `wrolpi/blobs/default.zim` (71 KB, passes `zimcheck`).

**`wrolpi/blobs/` shipping mechanism** — any committed blob in this directory is copied to `/opt/wrolpi-blobs` by the pi-gen (`04-run-chroot.sh`) and debian-live (`9999-wrolpi.hook.chroot`) builds, and baked into the Docker `zim` image (`docker/zim/Dockerfile`).

**Startup scripts**:
- `scripts/start_kiwix_serve.sh` (RPi/Debian): honors `zims_destination`; falls back to the default Zim when no functioning zim is found.
- `docker/zim/entrypoint.sh` (Docker): same default-Zim fallback. **No config read needed** — it already scans the whole mounted media directory, and `zims_destination` is always a subdirectory of that mount (relative-only, `root_api.py`), so there is no API→container communication and no new volume.

**Apply directory change without reboot** — `wrolpi/root_api.py`: changing `zims_destination` via the settings API now activates the `restart_kiwix` switch so Kiwix re-reads the new directory.

## Tests

- New `modules/zim/test/test_read_config_value.py` (6 cases).
- Extended `modules/zim/test/test_start_kiwix_serve.py` — configured-directory + default-Zim fallback (6 total).
- New `test_root_api.py` case: changing `zims_destination` restarts Kiwix; unrelated changes do not.
- **Full backend suite: 1665 passed.** (One unrelated pre-existing parallel-ordering flake, `test_search_suggestions`, passes in isolation.) No frontend code changed.

## Verification on hardware

To confirm on a device: set a custom `zims_destination`, restart `wrolpi-kiwix`, verify `curl http://127.0.0.1:9085/` serves from the new dir; then move/rename all zims and confirm the default Zim is served. Rebuild `wrolpi/blobs/default.zim` only via `wrolpi/scripts/default_zim/build_default_zim.sh` if `src/` changes.


lrnselfreliance:
## Triage of review findings (fixed in 386851e7)

All four addressed:

- **P1 — Restart can read stale config** (`root_api.py`): confirmed real — `update()` only queues an async `background_save`, and the switch worker pops LIFO with no ordering guarantee vs the restart switch, so the restart often fired first. Now the config is saved **synchronously** before the restart switch is activated (failure is non-fatal; `background_save` still runs).
- **P1 — YAML null becomes directory name** (`read_config_value.sh`): the reader now treats an unquoted `null`/`Null`/`NULL`/`~` as unset → returns the default, matching PyYAML/the Python layer. A quoted `"null"` stays a literal string.
- **P1 / security — Relative path escapes media root** (`start_kiwix_serve.sh`): the settings API now **rejects** a `zims_destination` containing `..`, and the startup script defends in depth by falling back to `zims` for any absolute/traversal/empty value — so Kiwix can never scan or rewrite `library.xml` outside the media root.
- **P2 — Floating toolchain** (`build_default_zim.sh`): pinned `ghcr.io/openzim/zim-tools:3.7.0` (the version that produced the committed `default.zim`) instead of `:latest`.

Added tests: YAML null handling, API traversal rejection, and the startup script's traversal fallback. Note the `..`/absolute validation gap also exists for the other `*_destination` keys on the Python side (pre-existing); left out of scope here, worth a follow-up.

PR #584, opened 2026-07-30
ui: Phase 2 part 1 — extension pages and Configs off Semantic

lrnselfreliance:
First slice of Phase 2 (call-site migration). Targets `feature/custom-theme`, not master.

Three files leave Semantic UI — `admin/ExtensionInstallSuggestion.js`, `admin/ExtensionPage.js`, `admin/Configs.js`. Baseline **75 → 72**.

## Components these call sites needed

All added to `src/components/ui`, all in the `/theme-sample` gallery, all tested:

| Component | Why |
|---|---|
| `Header` | Semantic's shape (`as`/`icon`/`subheader`/`dividing`). 47 files use `<Header>`. The level now picks the size, so the type scale stays a scale rather than each call site choosing a font size. |
| `Message` `icon` + `onDismiss` | 7 sites dismiss a message; the extension nudge is one. |
| `ActionInput` | Input and button sharing one outline — Semantic's `<Input action={...}/>`, 5 sites. |
| `Loading` | Centered loader with a caption, replacing `<Loader active inline='centered'>` and its wrapper div in 19 places. |
| `Modal` (compound) | `Modal.Header` / `.Content` / `.Actions`, plus Semantic's `open`, `closeIcon`, and size names. 34 files are written against that shape, so they migrate by import. |

## Two things reviewers should know

**`test-utils` now mounts `MantineProvider`**, with ThemeProvider's exact configuration. Migrated components render Mantine internals; without this every migrated component's test file would need its own provider.

**Semantic's global CSS does bleed**, as the plan anticipated. First real case: `semantic.min.css` sets `h1,h2,h3,h4,h5 { font-family: Lato }` and loads after Mantine's element rule, so headings ignored the theme's body face — only visible in amber, where the body is mono and headings were not. Fixed by stating `font-family` on `.wrolpi-header-text`; a class outranks the element selector, and the rule comes out when Semantic does. Expect more of these.

## Verification

- 925 jest tests pass; `npm run build` clean (no new warnings).
- Night-mode audit of every painted pixel in the gallery: red-only, with the one known false positive (`color` on an `appearance:none` checkbox input, which paints nothing).
- Visual pass in light, dark, night, and amber.
- Configs backup/restore exercised against the real API in the dev stack — list, merge preview, and diff view — without applying.


lrnselfreliance:
## Added: media filtering is now a per-theme setting

Filtering was keyed on `data-theme="night"`, which made it a property of the theme that no user could turn off. A theme now *declares* a filter and whether it starts on; the user decides.

| Theme | Filter | Default | Why |
|---|---|---|---|
| night | `night-red` | **on** | An unfiltered thumbnail undoes dark adaptation — the whole point of the mode. |
| amber | `amber-mono` | off | Amber is a look, not a night-vision aid, so media keeps its real colors until asked. |
| light / dark | — | — | No filter offered; no toggle shown. |

The choice is stored **per theme**, so wanting amber's tint says nothing about wanting night's. The toggle lives in the theme picker, which puts it on Settings and in the gallery at once.

**The rules now key on `data-media-filter`, never `data-theme`** — a rule keyed on the theme could not be turned off. `index.html` stamps that attribute before first paint alongside `data-theme`; without it night mode paints a screen of unfiltered thumbnails while the bundle arrives over a hotspot. Tests enforce both the attribute rule and that the inline script's duplicated ids/defaults still agree with `names.ts`.

Amber's matrix is the same luminance collapse tinted to `#ff9500` (red full, green 0.584, blue zero) with a gentler gamma — amber carries far more luminance than red and clips to a flat wash otherwise.

The gallery gains a **Media filtering** section with the WROLPi logo filtered beside an unfiltered copy that opts out via `night-unfiltered`. That copy sits behind a click on purpose: an unfiltered image is a bright patch, and night mode should not spring one on someone reading in the dark.

`ui-design.md` gains the media-filtering contract it never got in Phase 0, including that a Phase 3 custom theme names an existing filter id and cannot ship its own matrix — that would be code, not data.

Verified in the browser: night on by default → toggled off → amber off by default → toggled on, with the two choices independent; and the pre-paint script stamping both attributes while `#root` is still empty, in both directions. 944 jest tests, clean build.

PR #585, opened 2026-07-30
Strip whitespace from inventory item string values on save

### LINKED ISSUE #52: Inventory categories can contain whitespace
Inventory (sub)categories can have whitespace around them.  Strip whitespace from around all string values on save.

brisolt:
## What changed

modules/inventory/common.py: `_normalize_items` now strips leading/trailing whitespace from any string-valued item field before persisting an inventory (name, category, subcategory, location, etc.). Non-string values (numbers, None) are left untouched.

## Why

Inventory items are saved to disk exactly as typed, with no trimming, either client-side or server-side. Since inventory summaries and grouping (see InventorySummary/summarize.js) group items by exact string match on fields like category, a value saved as "grains " would silently form its own group separate from "grains". This has been open as #52 since 2021; the underlying inventory storage has since moved to the config-only YAML model, but the same gap exists in the current `_normalize_items`.

## Testing

Added `test_save_inventory_strips_whitespace` in modules/inventory/test/test_inventory.py, following the existing `test_save_inventory_items` pattern: saves an item with leading/trailing spaces in `name` and `category`, and asserts both come back trimmed.

Closes #52

### DIFF
--- a/modules/inventory/common.py
+++ b/modules/inventory/common.py
@@ -55,7 +55,7 @@ def _allowed_item_keys(fields_: list) -> set:
 
 
 def _normalize_items(items: list, fields_: list) -> list:
-    """Filter each item to the schema's allowed keys and assign stable ids (fresh id for any missing/duplicate)."""
+    """Filter each item to the schema's allowed keys, strip whitespace from string values, and assign stable ids (fresh id for any missing/duplicate)."""
     if not isinstance(items, list):
         raise ValidationError('items must be a list')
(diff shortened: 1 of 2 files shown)


lrnselfreliance:
@brisolt please squash your commits. You are bringing in some other commit in this PR.

brian-soltani:
Good catch, thanks. Squashed it down to one commit rebased on current master, that stray commit is gone now. Should be ready for another look.

PR #586, opened 2026-07-30
ui: Phase 2 parts 2–3 — Common.js and 22 more files off Semantic

lrnselfreliance:
Continues Phase 2 on `feature/custom-theme`. **Baseline 72 → 50.**

The important one is `Common.js`: 51 files import it, so its Semantic usage was propping up everything downstream. It's fully migrated, along with 22 other files.

## New in the library

All in the `/theme-sample` gallery, all tested.

| Component | Notes |
|---|---|
| `SearchBox` | Replaces `<Search category>`, **keeping its grouped `results` shape** — four call sites already build that structure. A real combobox: arrows move, Enter takes the highlighted suggestion or submits the raw text, Escape closes without clearing. |
| `Pagination` | Semantic's prop names; `onPageChange` receives the page number, not `(event, data)`. |
| `TabBar` + `tabClassName` | The strip and its house style; routing stays at the call site, which is what knows what's active. |
| `IconStack` | Semantic's `IconGroup`, named once for assistive tech. |
| `Table.HeaderCell` `sorted`/`onSort` | A real `<button>` carrying `aria-sort`. |
| `Button`/`IconButton` size mapping | Semantic size names → Mantine's scale. |

That last one is defensive: unmigrated callers still pass `size='tiny'`, and **Mantine drops a size it doesn't recognise without any warning**, so the button quietly renders at the default size. Mapping keeps the mixed state honest.

## Three real bugs, found and fixed

Not cosmetics — each was user-visible:

1. **`fileSuffixIconName` returned fourteen icon names the map didn't have.** Opening any folder containing a PDF, Word document, spreadsheet or archive logged a console error and drew a fallback glyph. All fourteen mapped, plus a test that walks the mapper's *own return values* and another asserting every mapped name resolves to a real Tabler component.
2. **`Statistic` depended on Semantic's `Statistic.Group` for spacing.** On the dashboard the load averages ran together as `1 MIN. LOAD5 MIN. LOAD15 MIN. LOAD`. It carries its own spacing now, so it doesn't rely on a parent's stylesheet.
3. **Semantic's inverted stylesheet colors every descendant of an inverted segment**, so `Toggle` labels inherited `#dddddd` — white pixels in night mode, on the Settings page, which is precisely where someone adjusts their WROLPi in the dark.

That third one is the same class of failure as the Lato heading bleed in the last PR, and it will keep happening while both stylesheets are loaded. Each fix is a token restatement with a comment saying it comes out when Semantic does.

## Ratchet improvement

The baseline test gained a third assertion. A file can stop *importing* `semantic-ui-react` and still depend on its stylesheet via `className="ui stackable grid"` — invisible to the import check, and it would collapse silently when Phase 4 deletes the CSS. There's one such bridge today (`CollectionEditForm`, kept deliberately because three unmigrated callers pass Semantic `Grid.Row` children) and it's now declared rather than incidental.

## Verification

- 962 jest tests, 56 suites, clean `npm run build`.
- Migrating `Common.js` broke 35 tests across 12 suites — all of them asserting Semantic DOM or rendering without Mantine's provider. Repaired without weakening a single assertion; no behavior regressions surfaced.
- Visual pass in the dev stack: dashboard, Files browser, Settings, and the gallery. Night-mode pixel audit of the migrated components on the Settings page — clean after fix #3.

Reviewers: `Common.js` is the file to read closely. The rest is mechanical by comparison.


lrnselfreliance:
## Triage: all three findings were real, all fixed in 37973559

**Greptile P1 — result callback shape.** Correct, and it was the worst of the three: Semantic handed callers `{result}`, and `Search.js:443` destructures exactly that. Passing the bare result left `result.location` undefined, so **selecting a search suggestion navigated nowhere** — both the dashboard search and the Ctrl+K modal. Restored the `{result}` shape, with a comment saying why it stays until the call sites can change together.

**Greptile P2 — `resultRenderer` dropped.** Also correct. Callers use it to render a tag suggestion as a chip instead of a line of text; my wrapper forwarded it into the surrounding `div`'s props rather than to `SearchBox`, so every suggestion silently fell back to generic title/description markup. `SearchBox` accepts it now.

**cypress `forms/directory-search.cy.js` — not a stale selector.** The failing assertion was `result descriptions appear correctly`, and the message was precise: *"not visible because its content is being clipped by one of its parent elements."* With directories + channels + domains + New Directory, the mock's eight suggestions across four group headings run to roughly 376px, past the 320px cap I'd set on the results list — so the last group was genuinely cut off. Eight grouped suggestions is an ordinary result set, so the cap was just too low. It's `min(60vh, 480px)` now, still bounded so a long list can't cover the page.

Worth saying plainly: **the cypress failure was a real UX bug, not test friction.** I'd have shipped a suggestion list that hides its last group.

Both regressions are now pinned by tests — one asserting the callback shape a caller actually destructures, one asserting the custom renderer *replaces* the default markup rather than wrapping it.

965 jest tests, clean build.

PR #587, opened 2026-07-30
ui: Phase 2 part 4 — thirty-five files off Semantic, leaving fifteen

lrnselfreliance:
**Baseline 50 → 15.** Continues Phase 2 on `feature/custom-theme`.

Migrated: all eight calculators, the rest of inventory and collections, the Map/Zim/Docs cluster, Videos, Files, Search, Archive, DashboardPage, DatesSelector, HelpModal, Events, Tags, admin Settings and Status, Download, App.

Plus the three shared modules everything imports — `api.js` and `customHooks.js` swap the Semantic toast library for the library shim (identical call shape), and `useForm.js` renders the library's `Message`. `useForm` accepts **both** the old `{content, header, positive|negative}` shape and the new `{kind, title}`, because the pages building those objects migrate on their own schedule.

## Two library gaps closed

Both found by call sites that needed them, not by design review:

- **`ActionInput` is a `forwardRef` now**, and the ref reaches the *input*, not the wrapper — callers use it to focus or select the field. One call site had resorted to hand-copying the component's markup to get a usable ref.
- **`Header` takes a `color` token name.** Three separate call sites had reached for an inline style to colour a heading that carries meaning ("3 items to add" vs "2 to remove"). That's how a hex ends up in markup.

## Please read: a process failure in this batch

I ran ten agents in parallel on this tree. One of them ran `git stash`, which swept **37 files of in-flight work from ten agents** into a stash, then restored only its own four files and treated the rest as unrelated leftovers.

I found it because an agent reported five files migrated and the files were untouched — my first reading was a fabricated report. The reflog showed otherwise: the work had been done and then deleted underneath it.

Recovery: stopped all ten agents, saved both the stash and the working tree as patches before touching anything, diffed the two to separate work that was *lost* (13 files, never redone) from work agents had already *rewritten* (24 files), and restored only the former.

**Two files then needed repair from being interrupted mid-edit** — `MapViewer` had a `Segment` left in its error branch, `VideoPlayer` had lost the `useContext` import its own code still used. Neither jest nor `tsc` caught either; only `npm run build` did.

The root cause is mine: I told agents not to edit shared *files* but never not to run git commands against the shared *tree*. Future agent prompts carry that prohibition explicitly, and real parallelism here wants `isolation: "worktree"` so agents cannot reach each other's work at all.

Reviewers may reasonably want to look harder at this diff than the last one, given how it was produced. The verification below is what I'd lean on.

## Verification

- 965 jest tests, 56 suites, all passing.
- `npm run build` compiles; **zero** undefined-symbol or missing-export warnings across the 35 files.
- `tsc --noEmit` clean.
- No app file imports `@mantine/core` directly (only the stylesheet import in `App.js` and the provider in `Theme.tsx`).
- The baseline ratchet's three assertions pass, including the Semantic-CSS-class check.

## What's left

`Archive`, `BulkTagModal`, `Channels`, `Docs`, `FileBrowser`, `FilePreview`, `Flasher`, `Nav`, `Playlists`, `Theme.tsx`, `Upload`, `VideoPlayer`, `Zim`, `ControllerPage`, `admin/Downloads`.

`Theme.tsx` is deliberately last: the unmigrated files still import its Semantic-backed wrappers, so it can only go once they have. `ControllerPage` needs parity with the controller's own HTML fallback UI.


lrnselfreliance:
## Greptile's P1 was real, and it was a silent failure

`App.js` migrated and took `<SemanticToastContainer/>` with it — but seven pages (Archive, Channels, Docs, FilePreview, Playlists, Zim, ControllerPage) still raised their toasts through the Semantic library. With no container mounted, **those notifications went nowhere**: no error, no console warning. A user deletes a document and is simply never told whether it worked.

Fixed by pointing those seven at the library's `toast` — same options — rather than mounting a second toast system. Nothing calls the Semantic toast library now, and there's a test asserting that **at zero**, because a file that quietly goes back to it would produce notifications that never appear, which no reviewer would catch by reading a diff.

## And it exposed something worse that Greptile didn't ask about

Editing `Archive.js` for that fix made the build fail with `'Panel' is not defined` / `'Tabs' is not defined` — errors that had been there all along. **react-scripts caches ESLint results per file**, so my earlier "clean build" was partly cache-blessed; touching the file invalidated its entry and the truth surfaced.

The underlying state: an interrupted agent had migrated part of `Archive.js` (Panel, Tabs, Mantine's `Grid.Col`) while the rest still used Semantic's `Segment`, `Tab` and `Grid.Row` — so one `Grid` identifier needed to be two different components at once.

I reverted `Archive.js` to its last good state and put it back on the baseline. A half-migrated 1400-line file is worse than an unmigrated one; it gets done in one piece later. **Baseline is 15 files, not 14** — `Archive` rejoins the list, and I'd rather the number be honest.

Then I cleared `node_modules/.cache` and rebuilt from scratch. That build is clean — no undefined symbols, no missing exports anywhere — so `Archive.js` was the only file hiding breakage.

**Lesson for the rest of this migration:** after a wide change, `npm run build` only means something with the ESLint cache cleared. I'll do that before every future migration PR.

966 tests, 56 suites, clean cache-busted build.

lrnselfreliance:
## Cypress: 36/43 → 43/43. Five stale selectors, **two real defects**.

### The two real ones, both mine

**The search modal stopped focusing its input.** Mantine mounts modal content on a later tick, so the effect that focused it ran while the ref was still `null`. Semantic's modal mounted synchronously, which is why focusing there had always worked. Fixed declaratively — the field asks for focus itself, and `SearchBox` also sets `data-autofocus`, which is what Mantine's focus trap honours when it takes focus away from a plain `autoFocus`.

**Escape closed the suggestion list *and* the enclosing modal together**, so a user correcting a mistyped query lost the whole modal. Escape now dismisses the suggestions and stops there; a second Escape, with the list already closed, still closes the modal. Standard combobox behaviour, and the test was poking right at it.

### Five stale selectors

Semantic's icon font (`i.search.icon`) is SVG now; `.ui.modal`/`.ui.dimmer` are the library's modal and overlay. One test asserted a button carried `ui button secondary` — I rewrote that to check what a user can actually perceive (Edit is a real link pointing at the edit route), because framework class names say nothing and break on every library change.

### Two accessibility gaps the tests walked into

- The nav's **search control** was an `<a>` with an `onClick`, **no accessible name and no keyboard path**. It now has a name, a role, and Enter/Space activation — which is also what let the spec stop matching on icon-font classes.
- The **search dialog had no accessible name** at all. It does now.

### One `{force: true}`, with its reason recorded

Cypress judges that field covered because it sits in a `position: fixed` modal under a sticky header. I verified by hand at Cypress's own viewport (1000×660) that the input is the topmost element at its own centre, unobscured, and focused on open — so the check is misfiring, not the component. The comment in the spec says exactly that.

### A hole in the ratchet, closed

This uncovered something worth knowing: **a file could drop its `semantic-ui-react` imports, leave the baseline, and still render Semantic components through `Theme.tsx`'s wrappers.** That is precisely how a Semantic modal survived in the search shortcut long after that file looked migrated — my "15 files remaining" was measuring imports, not markup.

A test now forbids it, and the three files doing it are fixed: `Calculators` and `KeyboardShortcutsProvider` import from the library, and `Common.js` takes theme *names* from `themes/names` (data, not markup) instead of through `Theme`.

### Verification

- **43/43 cypress specs**, run locally against the dev stack.
- 967 jest tests, 56 suites.
- Cache-busted `npm run build` (`rm -rf node_modules/.cache` first) — clean.

PR #588, opened 2026-07-30
ui: Phase 2 complete — Semantic UI is gone

lrnselfreliance:
**Baseline 15 → 0.** This finishes Phase 2: no file imports Semantic UI, and `semantic.min.css` is no longer loaded.

## How it was done

Nine agents in **isolated git worktrees**, one cluster each, merged branch by branch. That isolation was the fix for the previous round, where one agent's `git stash` in a shared tree destroyed 37 files of other agents' work. This time: zero cross-agent damage, and each agent could run its own cache-busted build. The only merge friction was a `package-lock.json` that npm rewrote inside a worktree, discarded after confirming the merged lock matched the branch point.

`Theme.tsx` I kept for myself and did last, because it could only go once nothing imported its wrappers. It drops **861 → 338 lines**: just the provider that resolves the theme, stamps `data-theme`/`data-media-filter`, persists both, and hands Mantine our tokens. The dark-mode compat props (`i`/`s`/`t`/`inverted`) went with the wrappers — they carried hardcoded `#1B1C1D`/`#dddddd`/`#eeeeee`, which is exactly what the tokens replaced, and three call sites in `Common.js` were still injecting one into links.

## Removing the stylesheet took more with it than expected

Tests and the build were both happy. **Only looking at the running app caught these:**

- **Every link lost its styling** — the dashboard's load averages and section headings rendered underlined and visited-purple. Links now take `--blue` (remaps per theme) and underline only on hover, since so many wrap cards, posters and icons.
- **`box-sizing: border-box` disappeared.** Semantic's reset had been providing it for the whole app; losing it silently changes the size of every padded element.
- **The nav bar overflowed 1000px and clipped "Admin"** — the one link someone needs to reach the settings that fix a broken WROLPi. Semantic's menu wrapped; ours does now too.

Gzipped CSS bundle: **47.5 kB**.

## Library defects the agents surfaced

Each was reported by an agent *working around* it rather than patching one file, so each is fixed once for every call site:

| Defect | What a user saw |
|---|---|
| `Button` accepted `href` and dropped it | A download button that downloads nothing |
| Media filter applied twice to nested media | Audio poster markedly darker than its surroundings |
| `Progress` had no indeterminate state | A pending upload reads as *stalled*; screen readers told 0% |
| Inline `--label-color` survived night mode | The Flasher's 12-colour chip badges put green and blue on a dark-adapted screen |

Plus two pre-existing bugs found and fixed: the **memory warning indicator never changed colour** (its colour was passed to react-router's `Link`, which ignores it — 81% and 95% RAM looked identical), and `useForm` was importing `Form` from `Theme` after it was deleted, which **the build reports as "was not found", not "is not exported"** — a phrasing my own grep had been missing.

## Ratchet kept, not deleted

At zero it asserts four things: nothing imports Semantic UI, nothing calls its toast library, nothing leans on its stylesheet, and no file reaches back through `Theme.tsx`. It goes when the packages are uninstalled.

`Theme.test.js` is deleted, but only after porting the two behaviours it protected: that `Button` forwards a ref to the real DOM element (a class-component ref once crashed the Control page, since portals call `node.contains`), and that the progress percentage stays readable on every bar colour.

## Deliberate visual changes (reviewed with Roland)

Accepted: restart buttons green → amber per the status vocabulary; Zim unsubscribe gained a confirmation; the Flasher's chip badges lose their palette in night/amber. **Restored on request:** the card mimetype-colour accent, which two agents had independently dropped — now a `color` prop on `Card`, drawn as a 3px top edge from a token so it remaps per theme.

## Verification

- **43/43 cypress specs**, run locally against the dev stack with the stylesheet gone.
- 948 jest tests, 55 suites.
- Cache-busted `npm run build` (`rm -rf node_modules/.cache` first) — clean, no undefined symbols, no missing exports.
- Visual pass on the dashboard and Files browser across themes.

## Not in this PR

Uninstalling `semantic-ui-react`, `react-semantic-toasts-2` and `semantic-ui-offline`, and the follow-ups the agents reported: `APIButton` has no icon-only variant so some icon-only API buttons lack accessible names; `Table` has no `warning` row or `disabled` row treatment; a pre-existing `path['path']` on a string in `FilePreview`; and dead `data` destructures in `Archive`.


lrnselfreliance:
## Card mimetype accent: restored, moved, and completed

Three commits since the PR opened, all on Roland's feedback:

1. **Restored** the accent as a `color` prop on `Card` — a token colour name drawn as a 3px edge rather than a tinted surface, so surfaces stay flat per the design rules and the accent remaps per theme (four hues in light and dark, one in night and amber).
2. **Moved it to the bottom edge.** Roland's call and the better one: along the top it sat immediately against the poster's own top border and the two edges competed. On the bottom it closes the card and reads as a property of the whole thing.
3. **Video cards were still missing it.** `Videos.js` has its own `VideoCard`, separate from the file cards in `Files.js`, so the first pass missed it — archives and documents had the edge and videos didn't.

On that last one I audited **every** remaining `<Card>` rather than fixing just the reported one. The only two without an accent are the map's place card and the loading placeholder, and neither has a mimetype to colour — both correct.

One thing worth stating so it doesn't read as a bug later: on the Videos page some accents are violet rather than blue. Those are audio-only file groups, and violet is `audio/*` in `mimetypeColor` — the same helper that colours the file icons, so a card and its icon can never disagree.

Verification after each change: 948 jest tests, 55 suites, cache-busted build clean, and **43/43 cypress specs** against the dev stack.

PR #590, opened 2026-07-30
ui: PathInput, and a browser layer that can see layout

lrnselfreliance:
Two commits: the bug the user reported, and the test layer that would have caught it — which found a second, older bug on its first run.

## 1. The reported bug

Five fields on `/admin/settings` passed **`leftSectionWidth='auto'`**. Mantine forwards that to the `--input-padding-inline-start` custom property, and `padding-inline-start: auto` is not valid CSS — it resolves to **zero**. The input reserved no room for the absolutely-positioned `leftSection` on top of it, so Archive Directory rendered as:

```
a/media/wrolpi/hain_tag)s/%(domain)s
```

It failed only in CSS, only at runtime, with no warning from webpack, `tsc`, or jest.

**The fix is structural, not a width.** `PathInput` makes the prefix and the input **flex siblings** sharing one outline, rather than one painted over the other. There is no width to compute and nothing to overlap, so the failure is impossible rather than corrected. It also sizes to any prefix, which matters: the media directory is `/media/wrolpi` on a Pi and differs under Docker, so any hardcoded width would have been wrong somewhere.

Six behaviour tests, no mocks beyond the provider: prefix is a sibling not a child; the value never contains the prefix; `onChange` reports only typed text; the prefix is the accessible description; `disabled` reaches the input; the ref forwards to the `INPUT`.

## 2. The layer that should have caught it

jsdom computes no layout — `getBoundingClientRect()` returns zeros, `var()` never resolves, no filter is ever applied. So the only guard available was a grep for one spelling of one mistake. `ui-layout.cy.js` adds the missing layer: `cy.mountUI()` mounts a library component in a real browser with Mantine's provider and a stamped `data-theme`, and **nothing else**.

That spartan setup is the argument. These components read no ambient state, so a spec can isolate one and assert real geometry, resolved tokens, and applied filters across **all four themes** — which until now was verified only by me looking at a browser.

It also **reproduces the `leftSectionWidth='auto'` overlap deliberately and asserts it still overlaps**, because a geometry check that has never once gone red is indistinguishable from one that cannot fail.

## 3. The bug that layer found

The rule resetting `filter` on leaf media inside a `.media` wrapper scored **(0,2,2)** against the **(0,3,1)** rule it was correcting. It lost every time. **It has never applied to anything.**

So every image inside a filtered wrapper has been through the luminance matrix *twice* since that rule landed — the second pass takes 0.2126 of an already-red pixel, leaving it markedly darker than its surroundings, which is precisely what the rule was written to prevent.

`tokens.test.js` asserted only that the rule's **text was present**, and so stayed green the whole time. Presence is not effect. It now compares the two selectors' computed specificity — catchable in jest, no browser needed — and goes red on the old selector.

## 4. Keeping the library mock-free by construction

`ui.test.js` runs 80 tests with **zero `jest.mock`**; `DomainEditPage.test.js` needs seven mocks before it renders at all. That gap is the reason the Settings inputs had no test when they broke. Three guards make the property structural rather than habitual: no `jest.mock` in the library's own tests, no `fetch`/`apiCall` in any library file, and no `useContext` outside `ThemePicker`, whose job is to read the theme.

## Verification

- 959 jest tests, 55 suites
- 14 new browser component tests
- 43 e2e tests
- clean `npm run build` after `rm -rf node_modules/.cache`

Every new guard was verified **red against a planted violation** before being trusted — the three mock guards, the specificity check, and the overlap check.

**Honest limits:** the `.media` fix is confirmed by measuring computed `filter` in a browser, not by eyeballing an audio poster. And the source-grep guard still only catches its one spelling; the browser layer is what generalizes.

### DIFF
--- a/app/cypress/e2e/settings/special-directories.cy.js
+++ b/app/cypress/e2e/settings/special-directories.cy.js
@@ -1,0 +1,117 @@
+/*
+ * The five Special Directories fields on /admin/settings.
+ *
+ * These are the fields that shipped printing their own prefix over their own value, and the
+ * ones whose visible labels were never associated with their inputs.  Both faults were
+ * invisible to the jest suite: the overlap because jsdom computes no layout, and the naming
+ * because the page needs enough provider and API mocking that it had no test at all.
+ *
+ * Intercepting the API is not mocking the component -- the page really renders, so the
+ * accessible names and the real geometry are the ones a user gets.
+ */
+describe('Special Directories settings', () => {
+    const mockSettings = {
+        archive_destination: 'archive/%(domain_tag)s/%(domain)s',
+        download_manager_disabled: false,
+        download_manager_stopped: false,
+        download_on_startup: true,
+        download_timeout: 0,
+        download_wait: 20,
+        download_window_start: null,
+        download_window_end: null,
+        hotspot_device: 'wlan0',
+        hotspot_on_startup: true,
+        hotspot_password: 'wrolpi hotspot',
+        hotspot_ssid: 'WROLPi',
+        check_for_upgrades: true,
+        ignore_outdated_zims: false,
+        log_level: 'info',
+        map_destination: 'map',
+        nav_color: 'violet',
+        media_directory: '/media/wrolpi',
+        playlists_destination: 'playlists',
+        tags_directory: true,
+        throttle_on_startup: false,
+        version: '1.0.0',
+        videos_destination: 'videos/%(channel_tag)s/%(channel_name)s',
+        wrol_mode: false,
+        zims_destination: 'zims',
+        save_ffprobe_json: true,
+    };
+
+    const directories = [
+        ['Archive Directory', 'archive/%(domain_tag)s/%(domain)s'],
+        ['Videos Directory', 'videos/%(channel_tag)s/%(channel_name)s'],
+        ['Map Directory', 'map'],
+        ['Zims Directory', 'zims'],
+        ['Playlists Directory', 'playlists'],
+    ];
+
+    beforeEach(() => {
+        cy.intercept('GET', '/api/status', {
+            statusCode: 200,
+            body: {
+                version: '1.0.0',
+                flags: {},
+                cpu_percent: 10,
+                memory_percent: 30,
+                downloads: {
+                    pending: 0, recurring: 0, disabled: false, stopped: false,
+                    outside_download_window: false,
+                },
+            },
+        }).as('getStatus');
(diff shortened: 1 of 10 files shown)


lrnselfreliance:
Triaged all three findings. **All three were real**, and one was worse than reported. Fixed in fe05bb6f.

### P1 — Directory inputs lose accessible names ✅ fixed

Confirmed: `<label>Map Directory</label>` with no `htmlFor`, and no `label` prop on the input. Five fields in one grid holding similar-looking relative paths, all sharing the prefix — so a screen reader announced five fields described only as "/media/wrolpi/" with no way to tell which one saves videos.

The labels now belong to the inputs, `InfoPopup` content included.

One correction to the framing: this **predates** `PathInput`. The old `TextInput` call sites had the same bare labels; the migration neither introduced the gap nor fixed it. Moving to a component that supports `label` is what made it fixable.

### P1 — Descriptions remain unassociated ✅ fixed

Confirmed. `aria-describedby` referenced only the prefix, so the help text listing the variables a path accepts, and the reason a path was rejected, were both rendered on screen and announced to nobody. Both are now in the accessible description, and an errored field sets `aria-invalid`.

### P2 — Long prefixes lack overflow handling ✅ fixed, and worse than described

Measured in a browser: at `flex: none` a long prefix took the whole row and left the input **18 pixels wide**. Not "consumes the editable area" — a field nobody can type into.

Worth stating why this was reachable: the prefix is `settings.media_directory`, whatever the API reports. It's `/media/wrolpi` on a Pi, but a development box or an unusual Docker mount can report far longer, so its length was never ours to assume. The prefix now shrinks and ellipsises at half the control, with the full path on `title` and in full in the accessible description.

### Tests

Each fix was verified **red against the original code** before being trusted:

- reverting one field to a bare `<label>` fails the new e2e spec (`expected undefined to be a string`)
- a long prefix fails the browser spec at `expected 18 to be above 60`
- the two association tests failed on the old `aria-describedby`

The naming fault deserves a note on *where* it's testable. jsdom computes no layout, and `Settings.js` needs enough provider and API mocking that it had no test at all — which is exactly why these five fields were uncovered when they broke. `cypress/e2e/settings/special-directories.cy.js` intercepts the API and lets the real page render, so it checks the names and geometry a user actually gets.

962 jest, 15 browser component, 46 e2e, clean cache-busted build, `tsc` clean.

Note: this review covered f8d7972f only. Two later commits add the browser test layer, three guards keeping the library mock-free, and a fix for a `.media` filter reset that lost on CSS specificity and had never once applied — a re-trigger would be welcome.

PR #591, opened 2026-07-30
test: one place for fixtures, real providers instead of mocked context modules

lrnselfreliance:
Answering "can we create a system so UI components can reliably use contexts and states?" — the emphasis on **reliably** turned out to be the important word.

## What was actually wrong

Specs had no shared vocabulary for the app's data or its contexts, so each one rebuilt both.

**Duplication:** a settings object was written by hand wherever one was needed, twenty-odd keys at a time, so a spec asserting one field still had to know the other nineteen. I contributed a copy of that myself last PR.

**Mocked context modules — the worse problem.** To render a consumer, a spec mocked the whole context module. That replaces *every* export, providers included, so each spec reinvented the pieces it still needed:

- one supplies a `Media` that always reports desktop, quietly leaving that component's mobile branch untested
- one hand-writes a context as a bare `_currentValue` object with no `.Provider`, working only because `useContext` reads that React internal

## The drift was not hypothetical

The theme fixture supplied `i`, `s` and `t` holding hardcoded Semantic greys for **months** after `ThemeProvider` stopped supplying them, and `ThemeContext` still declared all four. So the fixture agreed with the tests and with nothing else.

`FileBrowser` was still reading `inverted` off the context and pasting the resulting `undefined` into a className. Its footer — near-white `#f9fafb`, with a `.inverted` dark variant nothing could select any more — **has been light in dark, night and amber**. A white bar pinned to the bottom of the screen in night mode costs the user precisely the dark adaptation the theme exists to protect. It takes tokens now.

That is what an unchecked fixture buys you: a test that passes while the component is broken in a browser, and the fixture is the last place anyone looks.

## The system

`test-fixtures.js` is one definition of both the API shapes and the context values, shared by jest, the Cypress component specs, and e2e `cy.intercept` bodies — so the three cannot disagree about what the API returns. Both settings specs now use it. Every fixture is a function taking overrides; a shared mutable object lets one test change what the next one sees.

`test-fixtures.test.js` holds each fixture to its context's declared shape **in both directions**, so a property added to a context and forgotten in a fixture fails, and so does one deleted from a context and left behind.

`test-utils.js` now has three tiers: `renderUI` (Mantine only, for the library), `render` (real providers holding fixtures), and `mockModule` (always spreads `requireActual`, which hand-written factories forget — one blanks six hooks out of `customHooks` and every other export with them).

## Two failed attempts, recorded because they were instructive

Having the harness import `Tags.js` and `FileWorkerStatusContext.js` to supply those providers **broke eight suites**. It put both into the module graph of every spec ahead of that spec's own mocks, and components bound to a *second instance* of a mocked module — a different `jest.fn()` than the test had configured, so a hook mocked to return a value returned `undefined`, in a file no test mentions.

Requiring them lazily fixed that and not its mirror image: `useContext(TagsContext)` in a component read a different `TagsContext` than the harness had required, so the provider was ignored and the context default (`SingleTag: null`) won.

**A test harness must not change what a spec imports.** So it owns only the four contexts in `contexts/contexts.js`, and a spec passes any other context itself — resolving it exactly as the component does.

## Scope

`DomainTagging` is converted off the `_currentValue` fake as proof, and now asserts the tag name a user sees rather than a test id its own mock invented. **Five specs still mock a context module** and are listed in a ratchet so the count can only fall; converting them is follow-up work, and the snag is real — removing a wholesale module mock changes what else loads.

## Verification

974 jest / 15 browser component / 46 e2e, clean `npm run build` after `rm -rf node_modules/.cache`, `tsc` clean. Fixtures confirmed absent from the shipped bundle. No new dependency — `msw` would have been the obvious reach, and AGENTS.md rules out testing-only dependencies.

**Noted, not fixed:** 26 further dead `.inverted` rules remain in `App.css`, now unselectable. Worth its own sweep.

### DIFF
--- a/app/cypress/e2e/settings/download-window.cy.js
+++ b/app/cypress/e2e/settings/download-window.cy.js
@@ -1,43 +1,9 @@
+import {settingsFixture, statusFixture} from '../../../src/test-fixtures';
+
 describe('Download Window Settings', () => {
-    const mockSettings = {
-        archive_destination: 'archive/%(domain_tag)s/%(domain)s',
-        download_manager_disabled: false,
-        download_manager_stopped: false,
-        download_on_startup: true,
-        download_timeout: 0,
-        download_wait: 20,
-        download_window_start: null,
-        download_window_end: null,
-        hotspot_device: 'wlan0',
-        hotspot_on_startup: true,
-        hotspot_password: 'wrolpi hotspot',
-        hotspot_ssid: 'WROLPi',
-        check_for_upgrades: true,
-        ignore_outdated_zims: false,
-        log_level: 'info',
-        map_destination: 'map',
-        nav_color: 'violet',
-        media_directory: '/media/wrolpi',
-        tags_directory: true,
-        throttle_on_startup: false,
-        version: '1.0.0',
-        videos_destination: 'videos/%(channel_tag)s/%(channel_name)s',
-        wrol_mode: false,
-        zims_destination: 'zims',
-        save_ffprobe_json: true,
-    };
 
     beforeEach(() => {
-        cy.intercept('GET', '/api/status', {
-            statusCode: 200,
-            body: {
-                version: '1.0.0',
-                flags: {},
-                cpu_percent: 10,
-                memory_percent: 30,
-                downloads: {pending: 0, recurring: 0, disabled: false, stopped: false, outside_download_window: false},
-            }
-        }).as('getStatus');
+        cy.intercept('GET', '/api/status', {statusCode: 200, body: statusFixture()}).as('getStatus');
 
         cy.intercept('GET', '/api/tags', {
             statusCode: 200,
@@ -58,7 +24,7 @@ describe('Download Window Settings', () => {
     it('time inputs are empty when no window is configured', () => {
         cy.intercept('GET', '/api/settings', {
             statusCode: 200,
-            body: {...mockSettings},
+            body: settingsFixture(),
         }).as('getSettings');
 
         cy.visit('/admin/settings');
@@ -73,7 +39,7 @@ describe('Download Window Settings', () => {
         cy.intercept('GET', '/api/settings', {
             statusCode: 200,
             body: {
-                ...mockSettings,
+                ...settingsFixture(),
                 download_window_start: '08:00',
                 download_window_end: '17:00',
             },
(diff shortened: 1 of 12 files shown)


lrnselfreliance:
Both findings were real. Fixed in da414eeb.

### P2 — `mockModule` relative paths ✅ fixed

Correct, and this one is a plain miss on my part. `mockModule` shipped **with no caller at all**, so the usage in its own docstring had never once been executed. It called `jest.requireActual` on a path handed in by the spec, which resolves relative to `test-utils.js` — so for any spec in a subdirectory the documented `'../hooks/customHooks'` pointed outside `src` entirely. A helper whose stated purpose is preventing a mocking footgun, shipped with a mocking footgun.

It now takes the module the spec has already resolved:

```js
jest.mock('../hooks/customHooks', () => require('../test-utils').mockModule(
    jest.requireActual('../hooks/customHooks'),
    {useDomains: () => mockUseDomains()},
));
```

It **throws on a string** rather than trying to be clever. Silently resolving to the wrong module is the failure that was there to begin with.

And it has a caller now. `FileBrowser.test.js` listed six hooks in a hand-written factory and returned only those, so every other export of `customHooks` was undefined for anything it rendered — precisely the landmine the helper exists to remove. 31 tests still pass with the real module spread underneath.

### P2 — theme and colour scheme diverge ✅ fixed

Also correct. `inverted` drove `forceColorScheme` on its own, so `render(ui, {theme: {theme: 'night'}})` produced a tree that **reported night and rendered Mantine's light scheme**.

Worth being precise about why that matters more than it looks: production cannot reach that combination, and a harness inventing states the app never has is worse than one that is merely incomplete — whatever a test proves about that tree is true of nothing. It is the one thing a test harness must not do.

The scheme is now derived from the theme name through the existing `isDarkTheme`, which covers the case that made this reachable: `night` and `amber` are explicit-choice-only themes, so nothing else would ever put Mantine's own components on a dark scheme for them.

While there I also stamp `data-theme` on `<html>`, without which a component test cannot exercise **any** token table or the night-mode treatment — every one of those rules keys on that attribute. `isDark` stays overridable, so a spec deliberately checking what a component does when theme and `isDark` disagree can still say so.

### Tests

`test-utils.test.js` is new — the harness had no tests of its own, which is how both of these got in. **Eight of its twelve assertions were red** against the code as reviewed.

986 jest / 46 e2e, clean cache-busted build, `tsc` clean.

PR #592, opened 2026-07-30
test: hold the frontend fixtures to the API's own OpenAPI spec

lrnselfreliance:
Your idea, and it worked better than the source-parsing approach I was heading toward — the spec is already generated from the `@openapi.definition` response dataclasses, so pytest can read the real thing in-process off `/docs/openapi.json`.

One wrinkle first: **the spec was not trustworthy either.** Asserting fixtures against it as-found would have pinned them to a contract the API had already stopped honouring. So the chain is guarded in two links:

```
handler  <->  OpenAPI spec  <->  test-fixtures.js
```

The first keeps `/docs` honest about the endpoint — worth having on its own account, since that page is a promise to callers. The second keeps the fixtures honest about the spec.

## What the guards found on their first run

**`/api/settings`** — the published spec declared `hotspot_status` and `throttle_status`, both of which moved to the Controller and which the handler had long since stopped returning, while omitting four fields it does return. So `/docs` was advertising two fields the endpoint never sends.

`settingsFixture` was missing **seven** real fields (`download_daily_limit_global`, `download_daily_limit_per_domain`, `ignored_directories`, `map_default_location`, `require_cookies_unlocked`, `require_media_mounted`, `timezone`). Every spec intercepting that endpoint handed the app a body with seven holes in it, and every assertion about the rest passed regardless.

**`/api/status`** was further gone — spec declared seven fields the handler doesn't return, omitted seven it does. `statusFixture` was missing **seventeen** and **invented two**: `cpu_percent` and `memory_percent`, which no endpoint has ever returned and no component has ever read. They came from a hand-written cypress body and had been copied onward from there — including by me, last week.

## The status endpoint was declarable after all

Half of `/api/status` is spread from `shared_ctx.status`, which is why it looks open-ended. But that dict is seeded with a fixed set of keys in `contexts.py` and the worker only ever *updates* those — so the key set is stable, and it is now declared. That is recorded on the dataclass, since the next person will wonder the same thing.

## Gaps named rather than hidden

`domainFixture` and `tagFixture` stand in for endpoints whose `@openapi.definition` carries a description and **no `response=`**, so the spec promises nothing about their bodies — they remain hand-maintained guesses. `fileWorkerStatusFixture` has no declared shape at either end: that context's default is `null` *by design*, because `useFileWorkerStatus` relies on it to detect a missing provider.

All three are listed in a ratchet, because an uncovered fixture sitting quietly next to covered ones reads as coverage. Declaring a response on one of those endpoints is how its line comes off.

## Verification

Both links were confirmed by adding a field to the handler and watching them fire **in sequence** — first the spec check (`handler returns, spec omits: ['probe_new_field']`), then, once the schema was updated, the fixture check.

The reader itself had a bug the guards caught: its regex excluded digits, so it reported `is_rpi4`/`is_rpi5` as missing from a fixture that had them. Its own guard now covers that case.

**1706 pytest, 986 jest, 46 e2e.**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,33 @@ jobs:
           command: './venv/bin/pytest -nauto --junitxml=test-results.xml'
       - store_test_results:
           path: test-results.xml
+  # Do the API and the frontend's test fixtures still agree?
+  #
+  # `api-tests-3-11` runs these too, as part of the whole suite.  This job exists for the
+  # signal: fixture drift is silent by nature -- the frontend suite passes against a body the
+  # API stopped sending -- so it is worth a check whose name says what broke, failing in under a
+  # minute rather than somewhere inside seventeen hundred results.
+  #
+  # It needs no ffmpeg or catdoc (verified), and shares the dependency cache key with the other
+  # Python jobs, so it costs a venv restore and one test file.
+  api-contract-tests:
+    docker:
+      - image: cimg/python:3.11
+    resource_class: medium
+    steps:
+      - checkout
+      - restore_cache:
+          key: deps-3.11-{{ checksum "requirements.txt" }}-5
+      - run:
+          name: Install Requirements
+          command: |
+            python3 -m venv venv
+            . venv/bin/activate
+            pip install -r requirements.txt
+      - run:
+          name: Fixtures match the OpenAPI spec, and the spec matches the handlers
+          command: './venv/bin/pytest wrolpi/test/test_api_fixtures.py -v'
+
   api-alembic-migrations:
     docker:
       - image: cimg/python:3.11
@@ -180,6 +207,7 @@ workflows:
     jobs:
       - sanic-api-start
       - api-tests-3-11
+      - api-contract-tests
       - api-alembic-migrations
   #      - api-tests-3-12 Sanic does not yet support 3.12.
   wrolpi-app-test:
--- a/.github/workflows/jest-tests.yml
+++ b/.github/workflows/jest-tests.yml
@@ -1,10 +1,16 @@
 name: Jest Tests
 
+# `pull_request` is deliberately unfiltered.  It was `branches: [master]`, which meant a pull
+# request into anything else ran no jest and no webpack build -- so the entire Semantic UI to
+# Mantine migration, seventy-five files of it, was reviewed and merged across nine pull requests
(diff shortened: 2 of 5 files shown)


lrnselfreliance:
Right, and it generalized further than the case you raised. Fixed in f9fbde61.

### P1 — response types unchecked ✅ fixed

Correct on both counts: `StatusResponse.downloads` was `int` while the handler has always sent the download manager's summary object, and my guard compared property *names* only, so it passed straight over it. The spec was telling a generated client to expect a number where it receives a dict.

`downloads` is now `DownloadsSummaryResponse`, with the six fields `get_summary()` actually returns.

### Two more the same check then found

Once the comparison covered types I ran it against **both** endpoints rather than only `/api/status`, which turned up two more pre-existing mismatches:

- **`StatusResponse.wrol_mode: str`** — `wrol_mode_enabled()` returns a bool.
- **`SettingsResponse.log_level: str`** — the handler sends a Python logging integer. `SettingsRequest.log_level` already said `int`, and the frontend converts it numerically via `fromApiLogLevel()`.

That last one had a consequence worth spelling out. `settingsFixture` carried `log_level: 'info'`, and `fromApiLogLevel('info')` returns `undefined` — so **every Settings test has been rendering a blank log-level label** where production shows "Info". Same provenance as the `cpu_percent` fiction from the last round: a hand-written cypress body, copied onward. The fixture now carries `20`.

### Nested shapes

Also real, and it caught one: `statusFixture.downloads` was missing `daily_limit_reached`, a field the API has always sent.

`/api/status` returns an empty `downloads` object while the database is down, which it is for much of the suite, so the endpoint cannot vouch for its own nested shape. `DownloadManager.get_summary()` is the authority there and is now compared against the fixture directly.

### What is still name-only, and why

Fixture *values* are compared by name only. They are read from JS source rather than executed, so their types aren't reliably available without running node. The type check applies to the pair where real values exist on both sides — spec vs handler — which is where all three bugs lived. I've said that plainly in the test rather than leaving it to look like coverage.

Comparison is limited to properties the spec gives a `type` for and the handler actually sent, since a `null` from an optional field says nothing about its declared type.

**1709 pytest, 986 jest, 46 e2e.** One unrelated flake in one backend run — the known filename-formatting tests under `-nauto` — clean across two further full runs.

PR #606, opened 2026-08-01
ui: toasts — top right, readable titles, the clickable one restored, and 24 tests

lrnselfreliance:
`toast` had thirteen call sites and no test. It has already failed once app-wide — `App.js` dropped Semantic's container while seven pages still imported the old helper, and every notification silently went nowhere.

## Title was unreadable in dark, night and amber

Mantine's dark-scheme notification title is `color: var(--mantine-color-white)`, and the bridge aliases that to `--btn-text` — deliberately, because it is the text drawn on a filled button, which is near-black in those three themes. Measured: **`rgb(28,31,33)` on `rgb(38,41,43)`, about 1.1:1.** Present and invisible.

Fixing the alias isn't an option; it's correct for the buttons it exists for. So notifications take their surface, border and text from our tokens instead. After: **14.9 / 11.9 / 5.0 / 8.6** to one.

I also caught a regression I introduced doing it: I set the description to `--muted`, which is **2.0:1 in night** and 3.1 in amber. The description is the message, so night and amber drop the muting and let the title's weight (500 vs 400) carry the hierarchy.

Mantine writes its rule through `:where(...)`, which has zero specificity, so an `html`-prefixed single-class override outranks it.

## Moved to the top right

Bottom right put toasts over the file browser's sticky footer and a video's player controls — the two places a background task is most likely to report while the user is doing something else.

**One thing for review:** at scroll-top the first toast overlaps the nav bar. The nav is `position: static`, so it scrolls away and the overlap only exists at the top of a page. Left as-is rather than guessing at an offset — say if you want the container pushed clear of the header.

## The clickable toast was dropped by the migration

Semantic's `toast` took an `onClick` in its options and made the whole toast activate it. The migrated helper destructured four keys and dropped the rest, silently. `Events.js` still passes one — it did before the migration and was never changed — for the three events that arrive with a URL:

| event | what it announced |
|---|---|
| `user_notify_message` | a page shared from the browser extension |
| `upload_archive` | a finished archive upload |
| `screenshot_generated` | a generated screenshot |

The first renders the description **"Click here to view the shared page"**, which it was not. All three used to `window.open(url, '_self')`.

`onClick` is passed through again, with a pointer cursor. Two additions beyond parity:

- **Dismissing does not follow it.** The close button is inside the toast, so its click reaches the root; without a guard, closing an upload toast would navigate you away.
- **It is reachable from the keyboard**, which it never was even under Semantic. WROLPi runs on devices with a keyboard and no mouse. No ARIA `role` — `button` and `link` may not contain interactive content, and the dismiss button is in there.

Mantine also leaves that dismiss button unnamed, so a screen reader read it as "button". It has a label now, which is how the dismiss-doesn't-follow test finds it.

## Gallery

A Toasts section: one button per type, one that stacks a numbered toast per press, one that fills the container to its limit of five, one that never expires, one with no title, one long enough to wrap, one clickable, and clear-all. A toast is the one component that cannot be seen by loading a page.

## Testing

- jest: **1032 passing**, 59 suites. 18 new in `toast.test.js`.
- `ui-layout.cy.js`: **54 passing**, 6 new.
- `npm run build` clean. Checked by eye in all four themes, and the clickable toast verified in a real browser rather than against a spy.

**Eleven of the new tests were verified red against planted defects** — a shared `id` (which makes Mantine replace instead of stack), one colour for every type, `autoClose` taking the 0 literally, and the notification overrides removed.

### Four test-authoring corrections, all mine

Worth reading, because three of them are the kind that leave a green suite over a broken page:

1. **`cy.mountUI` never set `forceColorScheme`**, so every theme rendered against Mantine's *light* scheme. Any per-theme browser test could pass on a page that was broken in the app. This is the same harness the tag-contrast tests use.
2. **The support file never imported `@mantine/notifications/styles.css`**, which `App.js` does — so a Notification rendered unstyled and had no Mantine colour to be wrong about.
3. **The spec's own helper mounted twice**, and the second mount reset the theme to the default. All four per-theme cases ran in light.
4. **The `time: 0` test passed against a planted `autoClose: 0` twice.** Fake timers never complete the exit transition, so the element stays in the DOM regardless — first a synchronous assertion, then a control that could not fail either. It now uses real timers and waits for a normally-expiring toast to disappear first, so dismissal is demonstrably working before the survivor is asserted.

Mantine does treat a literal `0` as "close immediately" (`getAutoClose` returns the number, then `setTimeout(hide, 0)`), so that guard in `toast.ts` is load-bearing.

Pre-existing and unrelated: `Tags.cy.js`, `Channels.cy.js`, `Download.cy.js`, `Inventory.cy.js` fail with the same 11 Semantic-era selector errors, unchanged by this branch.

## Follow-up

In night and amber all four toast types are the same colour — deliberate per the token tables, but it means an error toast reads like an info toast. Belongs with the semantic-role-token pass (`--good`/`--warning`/`--bad`), along with the `--orange` problem.

`--muted` measuring 2.0:1 on a panel in night is not only a toast issue: statistic labels and inactive tabs use it too. Left alone here.

### DIFF
--- a/app/src/components/ThemeSamplePage.js
+++ b/app/src/components/ThemeSamplePage.js
@@ -33,6 +33,7 @@ import {
     ThemePicker,
     Toggle,
     Tooltip,
+    clearToasts,
     toast,
 } from './ui';
 import {semanticColorNames} from '../themes/mantine';
@@ -60,6 +61,10 @@ const Section = ({label, children}) => <section style={{marginBottom: 34}}>
 const Row = ({children}) => <div style={{display: 'flex', gap: 10, flexWrap: 'wrap', alignItems: 'center'}}>
     {children}
 </div>;
+
+// Numbers the stacked toasts so it is obvious which press produced which, and that a second
+// press adds a toast rather than replacing the first.  Module scope, so it survives re-renders.
+let toastCounter = 0;
 
 export function ThemeSamplePage() {
     const {theme} = useContext(ThemeContext);
@@ -208,6 +213,77 @@ export function ThemeSamplePage() {
                 <p style={{fontSize: 12, color: 'var(--muted)', marginBottom: 0}}>
                     In night mode Delete drops its fill for a dashed red outline; dashed means
                     destructive or failed, and nothing else uses it.
+                </p>
+            </Panel>
+        </Section>
+
+        <Section label='Toasts'>
+            <Panel>
+                {/*
+                  * `toast` has thirteen call sites and no test, and it has already failed once
+                  * app-wide: App.js dropped Semantic's toast container while seven pages were
+                  * still importing the old helper, and every notification silently vanished.
+                  * A toast is also the one component you cannot see by loading a page -- it has
+                  * to be provoked -- which is why it needs buttons here rather than a sample.
+                  */}
+                <Row>
+                    <Button role='primary' onClick={() => toast({
+                        type: 'info', title: 'Refresh started',
+                        description: 'Indexing the media directory.',
+                    })}>Info</Button>
+                    <Button role='save' onClick={() => toast({
+                        type: 'success', title: 'Saved',
+                        description: 'Channel settings written to config/channels.yaml.',
+                    })}>Success</Button>
+                    <Button role='cancel' onClick={() => toast({
+                        type: 'warning', title: 'Daily download limit reached',
+                        description: 'Downloads will resume tomorrow at 00:00.',
+                    })}>Warning</Button>
+                    <Button role='danger' onClick={() => toast({
+                        type: 'error', title: 'Download failed',
+                        description: 'HTTP 403 from the remote server after 3 attempts.',
+                    })}>Error</Button>
+                </Row>
+
+                <Row>
+                    {/* Click this repeatedly: each press is a separate toast, and the
+                        container holds five at once. */}
+                    <Button role='cancel' onClick={() => toast({
+                        type: 'info', title: `Stacked toast ${++toastCounter}`,
+                        description: 'Press again before this one expires.',
+                    })}>Stack one more</Button>
+                    <Button role='cancel' onClick={() => {
+                        // All five at once, to see the container at its limit.
+                        for (let i = 1; i <= 5; i += 1) {
+                            toast({
+                                type: 'info', title: `Toast ${i} of 5`,
+                                description: 'Filling the container to its limit.',
+                            });
+                        }
+                    }}>Fill the stack</Button>
+                    <Button role='cancel' onClick={() => toast({
+                        type: 'error', title: 'This one stays',
+                        description: 'time: 0 means it waits to be dismissed. Close it yourself.',
+                        time: 0,
+                    })}>Never expires</Button>
+                    <Button role='cancel' onClick={() => toast({
+                        type: 'warning',
+                        description: 'No title, only a description — some call sites send just this.',
+                    })}>No title</Button>
+                    <Button role='cancel' onClick={() => toast({
+                        type: 'error', title: 'A long one, to see how it wraps',
+                        description: 'Failed to fetch https://example.com/a/very/long/path/that/'
+                            + 'keeps/going/for/a/while.html: the remote server closed the '
+                            + 'connection after sending part of the response body.',
+                    })}>Long text</Button>
+                    <Button role='danger' onClick={() => clearToasts()}>Clear all</Button>
+                </Row>
+
+                <p style={{fontSize: 12, color: 'var(--muted)', marginBottom: 0}}>
+                    Bottom right, five at a time, five seconds each unless the caller says
+                    otherwise. Check each type in all four themes: night has no second hue to
+                    spend, so info, success, warning and error cannot be told apart by colour
+                    there.
                 </p>
             </Panel>
         </Section>
--- a/app/src/components/ThemeSamplePage.js
+++ b/app/src/components/ThemeSamplePage.js
@@ -276,6 +276,14 @@ export function ThemeSamplePage() {
                             + 'keeps/going/for/a/while.html: the remote server closed the '
                             + 'connection after sending part of the response body.',
                     })}>Long text</Button>
+                    {/* Events.js sends three of these, each opening a URL.  Click the toast
+                        body, or Tab to it and press Enter; the dismiss button does not follow. */}
+                    <Button role='cancel' onClick={() => toast({
+                        type: 'info', title: 'Screenshot Generated',
+                        description: 'Click here to view it.',
+                        time: 10000,
+                        onClick: () => toast({type: 'success', title: 'Followed the toast'}),
+                    })}>Clickable</Button>
                     <Button role='danger' onClick={() => clearToasts()}>Clear all</Button>
                 </Row>
 
--- a/app/src/components/ThemeSamplePage.js
+++ b/app/src/components/ThemeSamplePage.js
@@ -288,7 +288,7 @@ export function ThemeSamplePage() {
                 </Row>
 
                 <p style={{fontSize: 12, color: 'var(--muted)', marginBottom: 0}}>
-                    Bottom right, five at a time, five seconds each unless the caller says
+                    Top right, five at a time, five seconds each unless the caller says
                     otherwise. Check each type in all four themes: night has no second hue to
                     spend, so info, success, warning and error cannot be told apart by colour
                     there.
--- a/app/src/components/ui/toast.test.js
+++ b/app/src/components/ui/toast.test.js
@@ -4,6 +4,7 @@ import userEvent from '@testing-library/user-event';
 import {MantineProvider} from '@mantine/core';
 import {Notifications} from '@mantine/notifications';
 import {cssVariablesResolver, mantineTheme} from '../../themes/mantine';
+import {notificationsStore} from '@mantine/notifications';
 import {clearToasts, toast} from './toast';
 
 /*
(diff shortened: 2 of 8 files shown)


app/src/components/ThemeSamplePage.js
```diff
@@ -212,6 +217,85 @@ export function ThemeSamplePage() {
...
+                        type: 'info', title: 'Screenshot Generated',
+                        description: 'Click here to view it.',
+                        time: 10000,
+                        onClick: () => toast({type: 'success', title: 'Followed the toast'}),
+                    })}>Clickable</Button>
+                    <Button role='danger' onClick={() => clearToasts()}>Clear all</Button>
+                </Row>
+
+                <p style={{fontSize: 12, color: 'var(--muted)', marginBottom: 0}}>
+                    Bottom right, five at a time, five seconds each unless the caller says
```
lrnselfreliance: **[bug]** The new Toasts section help text still says “Bottom right…”, but `ThemeProvider` was deliberately changed to `position='top-right'`. Anyone using the theme sample page as a visual checklist will look in the wrong corner and may conclude toasts are broken (or that the PR did not land).

**Suggestion:** Update the copy to match the new placement, e.g. “Top right, five at a time, five seconds each unless the caller says otherwise.”

app/src/components/ui/toast.test.js
```diff
@@ -0,0 +1,278 @@
...
+ * were still importing the old helper, so every notification in the app silently went
+ * nowhere.  Nothing threw, and nothing looked wrong until somebody expected a message.
+ *
+ * The container is mounted here the way ThemeProvider mounts it, because a toast with no
+ * container is exactly the failure worth covering.
+ */
+
+const withToasts = () => render(
+    <MantineProvider theme={mantineTheme} cssVariablesResolver={cssVariablesResolver}>
+        <Notifications position='top-right' limit={5}/>
```
lrnselfreliance, line 24: **[suggestion]** Both Jest (`withToasts`) and Cypress (`showToast`) mount their own `<Notifications position='top-right' limit={5}/>` rather than rendering through `ThemeProvider`. That means a regression that moves Theme.tsx back to `bottom-right`, drops the container entirely from ThemeProvider, or changes `limit` will not fail these suites — yet the comments correctly call out “container missing while call sites still fire” as the failure mode that already shipped once. The harnesses re-encode the intended wiring instead of observing it.

**Suggestion:** Add at least one integration-style test that mounts `ThemeProvider` (or the same Notifications mount path the app uses) and asserts user-visible placement, for example:

```js
// After rendering ThemeProvider + a toast trigger:
const root = document.querySelector('[class*="mantine-Notifications-root"]');
expect(root).toHaveAttribute('data-position', 'top-right');
// and/or geometry:
const rect = root.getBoundingClientRect();
expect(rect.top).toBeLessThan(window.innerHeight / 2);
expect(rect.right).toBeGreaterThan(window.innerWidth / 2);
```

Keep the isolated `toast()` unit tests, but do not let them be the only guard for container position/limit.

app/src/components/ui/toast.test.js
```diff
@@ -0,0 +1,278 @@
...
+     */
+
+    it('goes away on its own', async () => {
+        withToasts();
+        showToast({title: 'Saved', time: 60});
+
+        await waitFor(() => expect(screen.queryByText('Saved')).not.toBeInTheDocument());
+    });
+
+    it('is still up a moment after being shown, when no time is given', async () => {
```
lrnselfreliance: **[suggestion]** “is still up a moment after being shown, when no time is given” only waits 150ms and asserts the toast is still present. That catches “vanishes on the next tick” (`autoClose: 0` style bugs) but would also pass for many wrong defaults (e.g. 200ms, 1000ms, `false`/never-closes, or Mantine’s container default of 4000ms). No test observes that the helper’s documented default of 5000ms is what actually gets applied, so the default duration is effectively untested.

**Suggestion:** Prefer an assertion that does not require sleeping five seconds. Either (a) spy on `window.setTimeout` once and assert a 5000ms hide timer is scheduled when `time` is omitted, or (b) export/test the `autoClose` mapping in a tiny pure helper (`autoCloseFor(time)`) and unit-test `undefined → 5000`, `0 → false`, `60 → 60`. Keep the existing real-timer expiry tests for the paths that already use short explicit `time` values.

app/src/components/ui/toast.test.js
```diff
@@ -0,0 +1,278 @@
...
+        // WROLPi runs on devices with a keyboard and no mouse.  A toast that only a pointer can
+        // follow is a feature half the deployments cannot use.
+        const onClick = jest.fn();
+        withToasts();
+
+        showToast({title: 'Screenshot Generated', onClick});
+        await waitFor(() => expect(visibleToasts()).toHaveLength(1));
+        visibleToasts()[0].focus();
+        expect(visibleToasts()[0]).toHaveFocus();
+        await userEvent.keyboard('{Enter}');
```
lrnselfreliance, line 204: **[suggestion]** Production code activates followable toasts on both `Enter` and `Space` (`toast.ts` `onKeyDown`), and the comments emphasize keyboard-only deployments. The keyboard test only presses `{Enter}`. A regression that drops the Space branch (or only handles `event.code === 'Enter'`) would stay green.

**Suggestion:** Add a sibling case:

```js
it('can be activated with Space', async () => {
    const onClick = jest.fn();
    withToasts();
    showToast({title: 'Screenshot Generated', onClick});
    await waitFor(() => expect(visibleToasts()).toHaveLength(1));
    visibleToasts()[0].focus();
    await userEvent.keyboard(' ');
    expect(onClick).toHaveBeenCalledTimes(1);
});
```

The existing `jest.fn()` usage here is fine: the toast, focus, and key event are real; the spy only records the user-visible outcome (handler ran).

app/src/components/ui/toast.test.js
```diff
@@ -0,0 +1,278 @@
...
+
+    it('carries the colour its type maps to', async () => {
+        withToasts();
+
+        showToast({type: 'error', title: 'Download failed'});
+
+        await waitFor(() => expect(visibleToasts()).toHaveLength(1));
+        // Mantine writes the resolved colour onto the element as a custom property.
+        const notification = visibleToasts()[0];
+        expect(notification.getAttribute('style')).toContain('red');
```
lrnselfreliance: **[nit]** Colour assertions use `style` string includes (`'red'`, `'blue'`) and a regex on `--notification-color`. That is better than pure mock theater — it reads the rendered DOM — but it is coupled to Mantine writing the palette name into an inline custom property. A future Mantine change to resolved hex/rgb would break these without any product regression, and `toContain('red')` is a soft match.

**Suggestion:** Prefer asserting the full custom property when present (`expect(style).toMatch(/--notification-color:\s*var\(--red\)/)` or `toContain('--notification-color: var(--mantine-color-red')`) so the type→colour map is checked explicitly, and leave cross-type uniqueness to the existing Set-size test.

lrnselfreliance (commented):
## Summary

This PR does three real things well: moves the Mantine notifications container to `top-right`, restyles toast title/description/surface onto theme tokens (fixing the dark-scheme invisible-title bug), and restores Semantic-era `onClick` followability that Events.js still relies on. The Jest suite is largely behavioral (real `Notifications` mount, real DOM, real timers for auto-close), and the Cypress contrast suite is exactly the right place to catch CSS-variable contrast failures jsdom cannot see. Dominant residual risks are incomplete wiring of the sample-page docs and tests that re-hardcode the container instead of asserting what `ThemeProvider` actually mounts, so the stated position change and the historic “container missing” failure mode are only partially locked down.

## Issue counts by severity

- bugs: 1
- suggestions: 3
- nits: 1

app/src/components/ThemeSamplePage.js
lrnselfreliance: Fixed. Straightforward miss — I moved the container and left the copy describing where it used to be, on the one page whose whole job is being a visual checklist.

app/src/components/ui/toast.test.js
lrnselfreliance: Right, and this is the sharpest of the five. Both suites mounted their own `<Notifications position='top-right' limit={5}/>`, so they re-encoded the wiring instead of observing it — while the comments in the same file cite "container missing while call sites still fire" as the failure that already shipped. Moving `Theme.tsx` back to bottom-right would have failed nothing.

Three tests in `ThemeProvider.test.js` now raise a real toast through the provider: that it appears at all, that it lands in the top right, and that five show at once.

Two things I got wrong on the way, worth recording:

- Querying for the container the way your snippet suggests returns `top-center` regardless of configuration — Mantine renders an empty container for **every** position it supports. The first version of the test failed for that reason rather than the intended one. It now asserts on `notification.closest('[class*="mantine-Notifications-root"]')`, which is where a real toast actually went.
- The five-at-once test does **not** guard the prop: Mantine's own default limit is 5, so deleting `limit={5}` leaves it green. I planted that regression to check, and it was the one of four that did not go red. The comment now says it catches the limit being *raised* rather than claiming to prove the prop is present.

Position and container-presence do go red when planted.

app/src/components/ui/toast.test.js
lrnselfreliance: Agreed, that test ruled out "vanishes on the next tick" and nothing else — it would have passed for a 200ms default, for `false`, and for Mantine's own 4000ms.

Took a third option over your two: `notificationsStore.getState()` holds exactly what `toast()` handed Mantine, so the real call path is exercised without a new export or a `setTimeout` spy. Three assertions — `undefined → 5000`, `250 → 250`, `0 → false`. Verified red by changing the default to 4000.

The real-timer expiry tests stay for the paths with short explicit times, as you suggested.

app/src/components/ui/toast.test.js
lrnselfreliance: Correct — `toast.ts` handles Enter and Space and only Enter was covered, so dropping the Space branch stayed green. Added, and confirmed red against `if (event.key === 'Enter')`.

app/src/components/ui/toast.test.js
lrnselfreliance: Fair. Both now match the whole property (`/--notification-color:\s*var\(--mantine-color-red-/`) rather than a substring — `toContain('red')` is satisfied by any declaration carrying those three letters.

On the coupling: it is real, and I have kept it deliberately. The alternative is asserting a resolved rgb, which pins the palette rather than the mapping, and the thing worth protecting here is that `type` reaches Mantine as the right colour name. If Mantine ever changes what it writes, these fail loudly and get updated — that is a better failure than the mapping silently going unchecked. Cross-type uniqueness stays with the Set-size test as you say.

PR #607, opened 2026-08-01
ui: test the last four components, and put the last four in the gallery

lrnselfreliance:
Closes the tail of the component library.

`Loader`, `Loading`, `Placeholder` and `Checkbox` had no test of their own. `IconStack`, `MediaFilterToggle`, `MultiSelect` and `NumberInput` were not in `/theme-sample`, which the project rule says every component must be.

Putting them in the gallery is what found the two bugs below.

## The icon stack was on the wrong surface everywhere but one

The corner glyph paints a disc of the surface behind itself so the two sets of strokes do not cross. That surface was hardcoded to `--bg`, which is right in exactly one of the three places a stack appears:

- the nav bar showed a **white disc on the red bar**
- the file browser's "New Folder" button showed one **punched through the blue fill**

It is now `--icon-stack-bg`, still defaulting to the page. The file browser passes the button's fill; the nav bar sets it on the `<nav>` next to the background it already writes there and lets it inherit — the bar's colour is a user setting, so it cannot be named by a token.

## MultiSelect pills were rounded

Mantine's `Pill` is the one component that ignores the radius theme: it writes `--pill-radius: rem(1000)` onto the element itself, which zeroing `--mantine-radius-xl` cannot reach. The chosen tags were lozenges on a page of square boxes.

## Tests

Twelve jest tests and sixteen browser tests. The browser half covers what jsdom cannot judge:

- a `Skeleton` is a different colour from the panel it covers, **in all four themes** — its greys are remapped to our tokens in `themes/mantine.ts`, and neither `--border` nor `--bg` was chosen with a panel behind it
- the loader colour resolves through the token rather than falling back to `currentColor`
- the corner disc matches its surface, both when the stack sets it and when an ancestor does (the nav bar's mechanism, which a test on the stack itself would not catch)

Every guard was checked **red** against a planted regression before being trusted — the CSS revert, a nonexistent token, an off-by-one on the ragged line, a skeleton forced to the panel colour, `style={style}` dropping the checkbox tokens, and the pill radius restored.

## Verified

- jest: **1051 passing**, 59 suites
- cypress component `ui-layout.cy.js`: **70 passing** (was 54)
- `npm run build`: clean
- all four themes checked in Chrome

The 11 other cypress component failures (`Tags.cy.js`, `Channels.cy.js`, `Download.cy.js`, `Inventory.cy.js`) are pre-existing Semantic-era selector assertions — same count and same messages as before this branch.

### DIFF
--- a/app/src/components/FileBrowser.js
+++ b/app/src/components/FileBrowser.js
@@ -437,7 +437,10 @@ export function FileBrowser() {
             disabled={wrolModeEnabled}
             style={{paddingLeft: '1em', paddingRight: showLabels ? '1em' : '0.8em'}}
         >
-            <IconStack corner={<Icon name='add'/>} label='New folder'>
+            {/* On a filled button, so the corner disc must be the button's fill and not
+                the page background, which would read as a hole punched in the button. */}
+            <IconStack corner={<Icon name='add'/>} label='New folder'
+                       style={{'--icon-stack-bg': 'var(--blue)'}}>
                 <Icon name='folder'/>
             </IconStack>
             {showLabels ? <>&nbsp;New Folder</> : null}
--- a/app/src/components/FileBrowser.js
+++ b/app/src/components/FileBrowser.js
@@ -437,10 +437,9 @@ export function FileBrowser() {
             disabled={wrolModeEnabled}
             style={{paddingLeft: '1em', paddingRight: showLabels ? '1em' : '0.8em'}}
         >
-            {/* On a filled button, so the corner disc must be the button's fill and not
-                the page background, which would read as a hole punched in the button. */}
-            <IconStack corner={<Icon name='add'/>} label='New folder'
-                       style={{'--icon-stack-bg': 'var(--blue)'}}>
+            {/* The corner disc follows the button's fill on its own -- including when WROL
+                mode disables it and Mantine repaints it grey.  See ui.css. */}
+            <IconStack corner={<Icon name='add'/>} label='New folder'>
                 <Icon name='folder'/>
             </IconStack>
             {showLabels ? <>&nbsp;New Folder</> : null}
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import {
-    ActionInput, Button, Header, PathInput, Statistic, StatisticGroup, TabBar, tabClassName, TextInput,
+    ActionInput, Button, Header, Icon, IconStack, Loading, MultiSelect, Panel, PathInput,
+    Placeholder, Statistic, StatisticGroup, TabBar, tabClassName, TextInput,
 } from './index';
 import {contrastingColor} from '../Common';
 import {Notifications} from '@mantine/notifications';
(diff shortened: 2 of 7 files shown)


app/src/components/ui/ui-layout.cy.js
```diff
@@ -619,6 +620,164 @@ describe('a toast is readable in every theme', () => {
     });
 });
 
+/*
+ * The colour a box actually shows: its own background, or the first pseudo-element that
+ * paints one.  Mantine's Skeleton draws itself entirely from ::before and ::after, so
+ * reading the root alone reports `rgba(0, 0, 0, 0)` and any test built on it is vacuous.
+ */
+const paintedBackground = (el) => {
```
lrnselfreliance: **[bug]** `paintedBackground` walks `null` → `::before` → `::after` and returns the first non-transparent paint. Mantine Skeleton always paints an opaque `::before` with `var(--mantine-color-body)` (`--bg` in this app) and puts the actual bar colour on `::after` (`--mantine-color-gray-3` / `--mantine-color-dark-4`, with z-index above `::before`). The new “separates the skeleton from the panel” cases therefore compare **page background vs panel**, not **skeleton bar vs panel**. Those two tokens differ in every theme by design, so the assertion stays green even if the visible bars collapse into the panel. The comment also misstates the light token as `gray-1` (`--bg`); the stylesheet uses `gray-3` (`--head`).

**Suggestion:** Sample the topmost painted layer (or `::after` specifically) for the skeleton, and keep comparing that to the panel’s own background. For example:

```js
const paintedBackground = (el, order = [null, '::before', '::after']) => {
    for (const pseudo of order) {
        const colour = getComputedStyle(el, pseudo).backgroundColor;
        if (colour && !/^rgba\(.*,\s*0(\.0+)?\)$/.test(colour) && colour !== 'transparent') {
            return colour;
        }
    }
    return null;
};

// skeleton bar is ::after; panel is the element itself
const line = paintedBackground($line[0], ['::after', '::before', null]);
const panel = paintedBackground($line[0].closest('.wrolpi-panel'), [null]);
```

Also fix the comment to say `gray-3` / `dark-4`.

app/src/components/FileBrowser.js
```diff
@@ -437,7 +437,10 @@ export function FileBrowser() {
             disabled={wrolModeEnabled}
             style={{paddingLeft: '1em', paddingRight: showLabels ? '1em' : '0.8em'}}
         >
-            <IconStack corner={<Icon name='add'/>} label='New folder'>
+            {/* On a filled button, so the corner disc must be the button's fill and not
+                the page background, which would read as a hole punched in the button. */}
+            <IconStack corner={<Icon name='add'/>} label='New folder'
+                       style={{'--icon-stack-bg': 'var(--blue)'}}>
```
lrnselfreliance: **[suggestion]** The New Folder stack hardcodes `'--icon-stack-bg': 'var(--blue)'` on the IconStack. That matches the filled blue button when enabled (the bug this PR fixes), but Mantine’s disabled button switches to `background: var(--mantine-color-disabled)` (`--head` here) while the disc stays `--blue`. With WROL mode on, the corner disc becomes a blue chip on a disabled surface—the inverse of the original hole. The Cypress happy-path only mounts an enabled button with the same hardcoded style, so it never catches this.

**Suggestion:** Put the surface token on the button (or drive it from the same cascade the button uses) and override for disabled, e.g. set `--icon-stack-bg` on the `Button` and add a small rule such as:

```css
.wrolpi-button:disabled .wrolpi-icon-stack-corner,
.wrolpi-button[data-disabled] .wrolpi-icon-stack-corner {
    background: var(--mantine-color-disabled);
}
```

Alternatively, inherit a variable the button already updates. Extend the Cypress case with a `disabled` button if you keep a fixed token.

app/src/components/ui/ui.test.js
```diff
@@ -880,6 +884,119 @@ describe('Toggle', () => {
     });
 });
 
+describe('Checkbox', () => {
+    it('reports and changes its state', async () => {
```
lrnselfreliance, line 888: **[suggestion]** “reports and changes its state” only asserts `onChange` was called. The checkbox is controlled with `checked={false}`, so the DOM never becomes checked and the handler payload is never inspected. That is weak coverage for the one behaviour test on Checkbox (the other two correctly lock the token-wrapper purpose).

**Suggestion:** Prefer an uncontrolled box (`defaultChecked={false}`) and assert `toBeChecked()` after click, or keep controlled and assert the event (`expect(onChange).toHaveBeenCalledWith(...)` with the checked payload Mantine emits). Rename if the test only claims “fires onChange”.

app/src/components/ui/ui.test.js
```diff
@@ -880,6 +884,119 @@ describe('Toggle', () => {
...
+
+        expect(screen.getByLabelText('Loading')).toBeInTheDocument();
+    });
+
+    it('renders no caption when there is nothing to say', () => {
+        const {container} = renderUI(<Loading/>);
+
+        expect(container.querySelector('.mantine-Loader-root')).toBeInTheDocument();
+        // An empty caption div would still take its line-height, moving the spinner off centre.
+        expect(container.querySelectorAll('div').length).toBe(1);
```
lrnselfreliance: **[nit]** “renders no caption” asserts `container.querySelectorAll('div').length === 1`. That works today because the oval Loader is span-based and MantineProvider does not wrap with a div, but any extra wrapper (provider, Loader type change to a div-based variant, test utility change) fails the test without a caption regression. The intent—no empty caption node stealing layout—is right; the probe is global.

**Suggestion:** Assert relative to the Loading root, e.g. only one child under the flex wrapper, or `expect(screen.queryByText(/.+/)).not.toBeInTheDocument()` / no second text node beside the loader.

app/src/components/ui/ui.css
```diff
@@ -357,6 +362,17 @@ html[data-theme="amber"] .wrolpi-searchbox-control:focus-within {
 }
 
 /* ------------------------------------------------------------- Form fields */
+/*
+ * A Pill is what MultiSelect renders for each chosen option, and it is the one place
+ * Mantine does not honour the radius theme: `Pill` writes `--pill-radius: rem(1000)`
```
lrnselfreliance: **[nit]** Comment claims Pill “writes `--pill-radius: rem(1000)` on the element itself”. In Mantine 8, MultiSelect’s pills leave `radius` undefined, `varsResolver` omits `--pill-radius`, and the **stylesheet** falls back with `border-radius: var(--pill-radius, 1000rem)`. The override (setting `--pill-radius: 0` on `.mantine-Pill-root`) is still the right fix; the comment overstates the mechanism and could mislead the next person into hunting an inline style that is not there.

**Suggestion:** Reword to: when `--pill-radius` is unset, Pill’s CSS falls back to `1000rem`; zero the variable (and/or `border-radius`) on `.mantine-Pill-root` so that fallback never wins.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -619,6 +620,164 @@ describe('a toast is readable in every theme', () => {
...
+        );
+
+        cy.get('.wrolpi-icon-stack-corner').should(($corner) => {
+            const red = getComputedStyle(document.documentElement).getPropertyValue('--red').trim();
+            expect(getComputedStyle($corner[0]).backgroundColor).to.equal(hexToRgb(red));
+        });
+    });
+
+    it('defaults to the page background when nothing says otherwise', () => {
+        // The other half of the claim: the default still has to be right for the nav bar,
```
lrnselfreliance: **[nit]** Default IconStack surface test comment says the default must stay right “for the nav bar, which is where every other stack in the app lives.” Nav now sets `--icon-stack-bg` to the user nav colour on both mobile and desktop bars; Hotspot stacks inherit that. The default `--bg` path is still worth testing for free-standing stacks, but the comment no longer matches the nav contract.

**Suggestion:** Describe the default as the fallback for stacks not under a surface that sets `--icon-stack-bg`, and point the nav case at the ancestor-inheritance test above it.

lrnselfreliance (commented):
## Summary

This PR finishes the UI library’s remaining surface area: unit tests for Checkbox / Loader / Loading / Placeholder, Cypress layout/theme checks for placeholders, loaders, IconStack surface colour, and MultiSelect pill radius, plus gallery entries and the real fix that IconStack’s corner disc is surface-aware via `--icon-stack-bg`. The product changes (FileBrowser, Nav, CSS) look correct for the enabled/nav cases they target, and most of the new tests assert meaningful DOM or cascade behaviour rather than pure mocks. The main risk is a vacuous Cypress skeleton contrast check that never samples the visible bar layer, plus a few incomplete edge cases around disabled buttons and brittle assertions.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 3

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Confirmed and fixed in e560a03b — you are right, and the token correction is right too.

Measured in Chrome: the element paints nothing, `::before` is `rgb(28,31,33)` = `--bg` at z-index 10, `::after` is `rgb(74,80,82)` = `--border` at z-index 11. Light is `--head` (`gray-3`), exactly as you said, not `gray-1`. So all four cases were comparing page background to panel — two tokens that differ by design in every theme. The bars could have collapsed into the panel entirely and nothing would have failed.

What is worth recording is why my planted regression missed it: **the plant was broader than the claim.** I forced `::before`, `::after` and the root all to the panel colour, so the test went red for a reason the fix did not depend on. There are now two plants — one setting only `::after` (goes red) and one setting only `::before` (stays green) — and the pair is what actually demonstrates the helper reads the visible layer.

Took the order-parameter suggestion as written.

app/src/components/FileBrowser.js
lrnselfreliance: Real, and fixed — but the fix went further than the suggestion, because the call site should not be answering this question at all.

`--icon-stack-bg` is gone from FileBrowser. The surface now comes from the cascade:

```css
.mantine-Button-root:not([data-variant]) .wrolpi-icon-stack-corner,
.mantine-Button-root[data-variant="filled"] .wrolpi-icon-stack-corner {
    background: var(--button-bg, var(--icon-stack-bg, var(--bg)));
}
.mantine-Button-root:disabled .wrolpi-icon-stack-corner,
.mantine-Button-root[data-disabled] .wrolpi-icon-stack-corner {
    background: var(--mantine-color-disabled);
}
```

A call site cannot know it is about to be disabled and repainted, so a fixed token there is wrong by construction.

One correction to your suggested selector, which I hit: `[data-variant="filled"]` alone matches **none** of our filled buttons. Mantine emits `data-variant` from the resolved variant, and `<Button color="blue">` leaves it undefined — the attribute is absent and the filled styling comes from the default. Every filled button in this app is written that way. Hence `:not([data-variant])`. Verified on the real `/files` page: the New Folder disc is `rgb(32,98,156)` against a `rgb(32,98,156)` button, with no style prop.

New cypress case mounts a `disabled` button, and it goes red with the disabled rule removed.

app/src/components/ui/ui.test.js
lrnselfreliance: Agreed — it only ever proved a handler fired. Split in e560a03b:

- "reports and changes its state" is now uncontrolled (`defaultChecked={false}`) and asserts `toBeChecked()` after the click.
- "hands the caller the new state" keeps it controlled and asserts the payload, since call sites read `event.currentTarget.checked` and a handler firing with the *old* value would have every one of them writing the state back unchanged.

Worth noting for the next person: the payload has to be read **inside** the handler. `currentTarget` is only set while the event is dispatching, so `onChange.mock.calls[0][0].currentTarget` is null by the time the mock is inspected — my first attempt at your suggestion failed exactly that way.

app/src/components/ui/ui.test.js
lrnselfreliance: Right, the probe was global. Now counted against the loader’s own parent:

```js
expect(loader.parentElement.children).toHaveLength(1);
```

Same intent — no empty caption node stealing a line-height — without depending on how many wrappers the provider adds.

app/src/components/ui/ui.css
lrnselfreliance: Correct, checked against the installed source. `Pill.mjs` has `"--pill-radius": radius === void 0 ? void 0 : getRadius(radius)`, so with `radius` undefined the variable is never emitted, and `styles.css:4686` is `border-radius: var(--pill-radius, 1000rem)`. The literal is in Mantine’s own stylesheet, not on the element.

Comment reworded to say that, and to say that defining the variable is what stops the fallback from applying. The fix is unchanged.

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Fair — I wrote that comment before the nav fix in the same PR, and never went back to it. It now describes the default as the fallback for a stack standing on the page, and points at the two tests that cover the real call sites: the nav bar by inheritance, the file browser by the button cascade.

PR #609, opened 2026-08-01
ui: give tables a real header, drop the column grid, quiet the row rules

lrnselfreliance:
Three issues, all visible on the file browser.

## The header was not a surface

Mantine gives `thead` no background at all, so the header painted whatever was behind the table — which is exactly what an unstriped body row paints. Against a striped body it read as **one more stripe**, with bold type the only thing distinguishing it.

It now carries `--head`, the token that already means "recessed surface", with a full-strength 2px rule beneath. The header/body division is the one line in a table that carries meaning — everything above it names, everything below is data — so it is the one allowed to be loud.

## Column borders are gone

A line around every cell is a lot of ink to repeat what the column headings already say, and it was half the reason the header disappeared: a bordered header cell looks like a cell in a grid.

## Row rules are quieter

They were `--border` — the same weight as the outline around the whole table — so a long table read as a stack of boxes. They now use `--table-line`, a new token per theme sitting between the surface and `--border`. The table outline is *not* a row rule and keeps `--border`; Mantine draws both from `--table-border-color`, so that had to be restated.

## File browser icon alignment

An icon and a filename are inline siblings in one cell. An inline SVG sits on the text baseline, so a 24px glyph beside 13px text hung above it — the icon looked centred in the row and the name looked like it was resting on the icon's floor. `vertical-align: middle` centres it.

The margin is a gap that was never there at all: the folder icon touched the first letter of its own name, and a folder and its "ignored" marker touched each other.

## Tests

One jest test (the column-border choice is a DOM attribute) and eleven browser tests — header distinct from *every* body row in all four themes, header rule heavier than a row rule, row rule quieter than the outline, no rule between cells, icon centred, gap present. Each was checked **red** against a planted regression.

Two traps worth recording:

- **`html .mantine-Table-table` is specificity, not decoration.** Mantine declares `--table-border-color` on its own hashed class, which ties with `.mantine-Table-table` and wins on source order.
- **The "quieter than the outline" test measures both lines against `--panel`.** Nothing paints `<html>` under a mounted component, so measuring against the root background compares both to black — which ranks the *paler* line as the louder one. The assertion passed backwards until it was pointed at the surface the rules actually sit on.

Cypress now loads `App.css` as the app does; without it a spec covering file browser rows measures a component the app never renders.

## Verified

- jest **1054 passing** / 59 suites
- `ui-layout.cy.js` **86 passing** (was 71)
- `npm run build` clean
- all four themes checked in Chrome on `/files`

The other cypress component failures (`Tags`, `Channels`, `Download`, `Inventory`) are pre-existing Semantic-era assertions. Correcting my own earlier reports: the count is **12**, not 11 — I confirmed against an unmodified `feature/custom-theme` worktree that `Inventory.cy.js` fails 3/3 there as well. One of its tests is flaky and I had caught it passing.

### DIFF
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import {
     ActionInput, Button, Header, Icon, IconStack, Loading, MultiSelect, Panel, PathInput,
-    Placeholder, Statistic, StatisticGroup, TabBar, tabClassName, TextInput,
+    Placeholder, Statistic, StatisticGroup, TabBar, Table, tabClassName, TextInput,
 } from './index';
 import {contrastingColor} from '../Common';
 import {Notifications} from '@mantine/notifications';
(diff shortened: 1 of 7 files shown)


app/src/components/ui/ui.test.js
```diff
@@ -737,6 +737,25 @@ describe('Table', () => {
...
+         * table, so the absence is assertable without a browser.
+         */
+        const {container} = renderUI(<Table>
+            <Table.Body>
+                <Table.Row><Table.Cell>a</Table.Cell><Table.Cell>b</Table.Cell></Table.Row>
+            </Table.Body>
+        </Table>);
+
+        const table = container.querySelector('table');
+        expect(table).not.toHaveAttribute('data-with-column-border');
```
lrnselfreliance: **[suggestion]** The new unit test claims “Mantine marks the [column-border] choice on the table,” then asserts `table` does not have `data-with-column-border`. Mantine never puts that attribute on `<table>` — only on `td`/`th` when `withColumnBorders` is true (`Table.components.mjs` sets `data-with-column-border` on cells). This expectation was already true when column borders were enabled, so the test would not catch a regression that re-adds `withColumnBorders`.

**Suggestion:** Assert on cells instead, e.g. `expect(container.querySelector('[data-with-column-border]')).toBeNull()` or `expect(container.querySelector('td')).not.toHaveAttribute('data-with-column-border')`. Keep the `data-with-table-border` check on the table; that one is real. The Cypress `borderRightWidth === 0` check already covers the painted outcome.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -146,6 +146,177 @@ describe('ActionInput', () => {
...
+     * resting on the icon's floor.
+     *
+     * Purely a question of layout: jsdom returns zeros from getBoundingClientRect, so
+     * nothing there can tell a centred glyph from one hanging off a baseline.
+     */
+    const fileRow = <Table>
+        <Table.Body>
+            <Table.Row>
+                <Table.Cell className='file-path'>
+                    <Icon name='file audio'/>
```
lrnselfreliance: **[suggestion]** The file-path layout fixture uses `<Icon name='file audio'/>` with the default size (`small` = 16px), but the CSS comment and the failure mode it documents are about a ~24px glyph beside body text. File browser file rows pass `size={null}` into `FileIcon`, which resolves to Tabler’s default 24px — the harder centering case. The current fixture can pass while a 24px icon is still optically off.

**Suggestion:** Match production: render the icon at 24 (or use `FileIcon` with the same `size={null}` path) so the centre-vs-text assertion exercises the layout that users actually see. Optionally also mount a two-icon row (`folder` + `eye slash`) since the CSS comment specifically cites icons touching each other.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -146,6 +146,177 @@ describe('ActionInput', () => {
...
+            cy.get('thead th').first().should(($th) => {
+                const header = getComputedStyle($th[0]).backgroundColor;
+                expect(header, 'the header paints a surface of its own')
+                    .to.not.match(/^rgba\(.*,\s*0(\.0+)?\)$/);
+
+                const rows = [...Cypress.$('tbody tr')].map(row => {
+                    // An unstriped row paints nothing, so what shows is the page behind it.
+                    const own = getComputedStyle(row).backgroundColor;
+                    return /^rgba\(.*,\s*0(\.0+)?\)$/.test(own)
+                        ? getComputedStyle(document.documentElement).backgroundColor
```
lrnselfreliance: **[nit]** For unstriped body rows the header-surface test falls back to `document.documentElement`’s background. Under `mountUI`, `<html>` usually paints transparent (`rgba(0,0,0,0)`); only `body` gets `background: var(--bg)`. That makes `header !== unstripedRow` a free pass whenever the header is any opaque colour — the real painted surface behind an unstriped row is the page/`body`, not the root element.

**Suggestion:** Walk to the first ancestor with a non-transparent background (as `effectiveBackground` already does elsewhere in `component.js`), or compare against `document.body` / `token('--bg')` so the unstriped case is a real colour inequality.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -146,6 +146,177 @@ describe('ActionInput', () => {
...
+            });
+        });
+    });
+
+    it('rules off the header with a heavier line than it puts between rows', () => {
+        /*
+         * The header/body division is the one that carries meaning -- everything above it
+         * names, everything below is data -- so it is the one line allowed to be loud.
+         */
+        cy.mountUI(sampleTable, {theme: 'light'});
```
lrnselfreliance: **[nit]** The “heavier line under the header” assertion runs only for `theme: 'light'`. Widths are theme-independent today (2px vs 1px), so this is low risk, but the rest of the table suite already loops `themeNames` and a future per-theme border tweak would slip past.

**Suggestion:** Fold it into the existing `themeNames.forEach` block next to the header-surface and quiet-rule tests, or at least run once on a dark scheme theme as well.

lrnselfreliance (commented):
## Summary

Solid, well-motivated table restyle: drop column borders, give `thead` a real `--head` surface, keep the outer outline at full `--border`, and quiet body row rules via a new `--table-line` token. CSS specificity work against Mantine’s post-imported stylesheet is careful and correct, tokens are defined in all four themes, and the Cypress suite asserts real computed styles rather than class-name theater. A few test assertions are weaker than the comments claim (especially the jsdom column-border check), but nothing looks like a functional regression in the DataTable/CSS/token path.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 2

app/src/components/ui/ui.test.js
lrnselfreliance: Correct — verified in the installed source. `Table.components.mjs:25` is the only place `data-with-column-border` is written, and it goes on the cell. The table never carries it, so my expectation held whether column borders were on or off. Fixed to `expect(container.querySelector("[data-with-column-border]")).toBeNull()`, which now goes red when `withColumnBorders` is restored.

The part worth recording is how I missed it: **I did plant that exact regression.** I restored `withColumnBorders` and saw red — but I ran the plant against cypress only, where `draws nothing between the cells of a row` caught it. The plant was right and the suite was wrong. Both suites now run for every plant.

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Right on the size, and I had the evidence in hand and did not use it — I read `width="24" height="24"` off the real `/files` DOM earlier in this same session, then wrote a fixture at the 16px default.

The mechanism, for the record: a file row calls `<FileIcon size={null}/>`; `Icon` defaults `size` to `small` only for `undefined`, so an explicit `null` falls through to `sizes[null]` → `undefined` → Tabler's own 24. Fixture is now 24 with that noted.

Took the two-icon suggestion too — `folder` + `eye slash`, which is what an ignored folder draws. It asserts the two glyphs are apart, that the second is clear of the name, and that the *second* glyph is centred as well, since only the first was being checked.

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Confirmed: `tokens.css:188` puts `background: var(--bg)` on `body`, and nothing sets one on `html`. So the fallback read `rgba(0, 0, 0, 0)` and any opaque header passed the inequality.

Now compares against the `--bg` token — what actually shows through an unstriped row — plus an assertion that the body really is striped, so both row kinds are covered rather than whichever happened to be sampled.

Planted the precise case the old form could not see (header painted `--bg`, opaque and identical to an unstriped row): red in all four themes.

Second time this exact trap has bitten in this PR — the `--panel` note in the commit message is the same root cause in the row-rule test. Both are now written against a token rather than a computed root background.

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Done — folded into the `themeNames.forEach` block beside the header-surface test, so it runs in all four.

PR #610, opened 2026-08-01
ui: semantic roles, so night and amber can tell severity apart at all

lrnselfreliance:
A component now asks for what a colour **means** — `danger` — and the theme decides how that looks. Five roles: `neutral`, `info`, `success`, `warning`, `danger`.

## Why this was worth doing

The token set is the easy half. Light and dark spend a hue per role. Night is red and nothing else, amber is amber and nothing else — and there the hue names are **synonyms**:

```
night   --orange  is byte-identical to  --text
        --yellow  is byte-identical to  --amber
amber   the same collisions, one hue over
```

A call site picking between hue names in those themes is picking between the same pixel. That is why all four toast kinds looked identical in night, and why an error message was indistinguishable from an informational one.

**Correction to my original claim in this PR.** I wrote that all four toast kinds were "the same pixel" in night. They were not — planting a reversion to hue names left the new distinctness test green, because in night `blue/green/yellow/red` are four different values. What they were is *cramped and badly ranked*, with info at **2.29:1** against the panel — below the 3:1 floor, present and invisible.

The byte-identical pairs are real but bite elsewhere: a call site reaching for `--orange` or `--yellow` in night gets `--text` or `--amber`. That is the synonym problem; the toast mapping's problem was legibility and ranking.

After the change, night's toasts are a deliberate ramp:

```
info #c02929 (3.41:1)   success #e02323 (4.21:1)
warning #ee3434 (4.91:1)   danger #ff3e3e (5.71:1)
```

## The ramps are built, not copied

Every role clears 3:1 against the panel (`neutral` is allowed lower — for `pending` and disabled, *dim is the signal*), severity is strictly increasing in contrast, and `danger` reaches that theme's own `--text` brightness so the loudest role is never quieter than prose.

Amber needed its own scale rather than night's: its `--text` is **8.6:1** where night's is **5.0:1**, so night's ramp would have left every amber role dimmer than ordinary text. Saturation climbs with severity too, so `danger` is the most saturated colour on screen and not merely the palest.

**Fixed in passing:** night's info was **2.29:1** — under the 3:1 floor.

## Naming

These are the names already in the tree. `--primary/--success/--warning/--danger` were declared in `tokens.css` with **zero consumers**, and `MessageKind` was already `info|success|warning|error`. This collapses three vocabularies into one.

`error` stays a valid kind — 26 call sites write it, and renaming those is app-level churn for no behaviour change. `danger` is the same thing under the token's name, for new code.

`--primary` is deliberately **not** a role: recolouring your primary button shouldn't recolour your messages.

## Status loses its per-theme overrides

`Status` was the one component already encoding severity by brightness, by hand, with three `html[data-theme="night"]` rules. The roles are that generalised, so those are gone and one rule serves four states in four themes.

## Scope

Tokens plus the library (`Message`, `Status`, `Progress`, `Label`, `toast`). The ~24 app-level `color='...'` sites are a follow-up — each needs a judgement, and two already disagree: `Downloads.js` colours the same "Download error" label `orange` on line 241 and `red` on line 456. The 22 `violet` uses are not semantic at all and stay hue-named.

## Tests

The ordering and danger-vs-text claims run **only for the monochrome themes**, via a new `isMonochromeTheme` predicate. In light, blue is *darker* than green on white — asserting a brightness ramp there fails a design that is perfectly fine. Distinctness and the 3:1 floor are asserted everywhere.

Roles were folded into each theme's own block rather than a standalone one, so the existing "same tokens in every theme" and "only red hues in night" guards cover them for free.

Every guard checked red against a planted regression, including the original bug (warning and danger collapsed to one value in night) and night's info back at its old 2.29:1. One plant of mine came back green and was **my error, not the test's** — I set `failed: 'warning'`, which is still four distinct roles; re-planted as a real collision.

## Verified

- jest **1055 passing** / 59 suites
- `ui-layout.cy.js` **107 passing** (was 90)
- `npm run build` clean
- all four themes checked in Chrome, plus the new gallery section

Pre-existing cypress failures unchanged at 12.

### DIFF
--- a/app/src/components/ui/Feedback.tsx
+++ b/app/src/components/ui/Feedback.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import {Loader as MLoader, Skeleton} from '@mantine/core';
 import {Icon} from './Icon';
-import {SemanticColorName} from '../../themes/mantine';
+import {RoleName, SemanticColorName} from '../../themes/mantine';
 
 /*
  * Feedback: messages, labels, progress, status text, loaders.
@@ -12,13 +12,20 @@ import {SemanticColorName} from '../../themes/mantine';
 
 // ------------------------------------------------------------------ Messages
 
-export type MessageKind = 'info' | 'success' | 'warning' | 'error';
+/*
+ * `error` is the original spelling and stays: 26 call sites use it, and renaming those
+ * is app-level churn for no behaviour change.  `danger` is the same thing under the
+ * name the token uses, and is what new code should write.
+ */
+export type MessageKind = 'info' | 'success' | 'warning' | 'error' | 'danger';
 
-const messageColors: Record<MessageKind, string> = {
-    info: 'var(--blue)',
-    success: 'var(--green)',
-    warning: 'var(--amber)',
-    error: 'var(--red)',
+/** The role each kind means.  Never a hue -- see the note on roles in tokens.css. */
+const messageRoles: Record<MessageKind, RoleName> = {
+    info: 'info',
+    success: 'success',
+    warning: 'warning',
+    error: 'danger',
+    danger: 'danger',
 };
 
 export interface MessageProps {
@@ -35,10 +42,10 @@ export interface MessageProps {
 export function Message({kind = 'info', title, children, icon, onDismiss, className}: MessageProps) {
     return <div
         // Errors and warnings interrupt; info and success are announced politely.
-        role={kind === 'error' ? 'alert' : 'status'}
-        className={['wrolpi-message', kind === 'error' ? 'wrolpi-message-error' : '', className]
+        role={messageRoles[kind] === 'danger' ? 'alert' : 'status'}
+        className={['wrolpi-message', messageRoles[kind] === 'danger' ? 'wrolpi-message-error' : '', className]
             .filter(Boolean).join(' ')}
-        style={{['--message-color' as string]: messageColors[kind]}}
+        style={{['--message-color' as string]: `var(--${messageRoles[kind]})`}}
     >
         {icon && <div className='wrolpi-message-icon'>
             {typeof icon === 'string' ? <Icon name={icon} size={20}/> : <Icon component={icon} size={20}/>}
@@ -61,7 +68,7 @@ export function Message({kind = 'info', title, children, icon, onDismiss, classN
 // -------------------------------------------------------------------- Labels
 
 export interface LabelProps {
-    color?: SemanticColorName | 'black' | 'white';
+    color?: SemanticColorName | RoleName | 'black' | 'white';
     icon?: string | React.ComponentType<any>;
     /**
      * Draw it as a physical tag — pointed left edge and an eyelet — rather than a plain
@@ -100,7 +107,7 @@ export interface ProgressProps {
     indeterminate?: boolean;
     /** Show the percentage inside the bar. */
     showPercent?: boolean;
-    color?: SemanticColorName;
+    color?: SemanticColorName | RoleName;
     /** Replaces the percentage with arbitrary text (e.g. "2.1 GB / 5.3 GB"). */
     label?: React.ReactNode;
     className?: string;
@@ -137,6 +144,19 @@ export function Progress({
 
 export type StatusKind = 'complete' | 'active' | 'pending' | 'failed';
 
+/*
+ * Status was the one component already doing this correctly: it had per-theme overrides
+ * in ui.css because night has no hues to distinguish four states with.  Those are gone --
+ * the roles carry the brightness ramp now, so this table is the whole of it and every
+ * theme is served by the same rule.
+ */
+const statusRoles: Record<StatusKind, RoleName> = {
+    complete: 'success',
+    active: 'info',
+    pending: 'neutral',
+    failed: 'danger',
+};
+
 const statusIcons: Record<StatusKind, string> = {
     complete: 'check',
     active: 'circle notch',
@@ -157,7 +177,10 @@ export interface StatusProps {
  * the color.
  */
 export function Status({kind, children, plain}: StatusProps) {
-    return <span className={`wrolpi-status wrolpi-status-${kind}`}>
+    return <span
+        className={`wrolpi-status wrolpi-status-${kind}`}
+        style={{['--status-color' as string]: `var(--${statusRoles[kind]})`}}
+    >
         {!plain && <Icon name={statusIcons[kind]} size={14} loading={kind === 'active'}/>}
         {children}
     </span>
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -1,11 +1,12 @@
 import React from 'react';
 import {
     ActionInput, Button, Header, Icon, IconStack, Loading, MultiSelect, Panel, PathInput,
-    Placeholder, Statistic, StatisticGroup, Status, TabBar, Table, tabClassName, TextInput,
+    Message, Placeholder, Statistic, StatisticGroup, Status, TabBar, Table, tabClassName,
+    TextInput,
 } from './index';
 import {contrastingColor} from '../Common';
 import {Notifications} from '@mantine/notifications';
-import {toast} from './toast';
+import {clearToasts, toast} from './toast';
 import {monochromeThemes, themeNames} from '../../themes/names';
 
 /* `--panel` is authored as a hex; computed backgrounds come back as rgb(). */
@@ -13,6 +14,24 @@ const hexToRgb = (hex) => {
     const value = hex.replace('#', '');
     const [r, g, b] = [0, 2, 4].map(i => parseInt(value.slice(i, i + 2), 16));
     return `rgb(${r}, ${g}, ${b})`;
+};
+
+/*
+ * Normalise a colour to rgb(), and REFUSE anything that is neither a hex nor already
+ * rgb().  Strictness is the point: feeding an unresolved `var(--blue)` to a luminance
+ * calculation yields NaN, and `expect(NaN).to.not.equal(x)` passes -- a test that cannot
+ * fail.  Better to blow up naming the value than to quietly measure nothing.
+ *
+ * Tokens do resolve: `getPropertyValue('--info')` returns `#20629c` in light, because
+ * custom properties are substituted at computed-value time.  This guards the day that
+ * stops being true on some engine, and catches the hex/rgb mismatch that is easy to hit
+ * when reading a Mantine variable rather than a painted property.
+ */
+const toRgb = (value) => {
+    const trimmed = String(value).trim();
+    if (/^rgba?\(/.test(trimmed)) return trimmed;
+    if (/^#[0-9a-fA-F]{6}$/.test(trimmed)) return hexToRgb(trimmed);
+    throw new Error(`Not a resolved colour: "${trimmed}"`);
 };
 
 /*
@@ -192,13 +211,11 @@ describe('semantic roles read as severity in every theme', () => {
 
     const roleColours = () => {
         const root = getComputedStyle(document.documentElement);
-        const panel = hexToRgb(root.getPropertyValue('--panel').trim());
+        const panel = toRgb(root.getPropertyValue('--panel'));
         return {
             panel,
-            text: hexToRgb(root.getPropertyValue('--text').trim()),
-            roles: ROLES.map(role => ({
-                role, colour: hexToRgb(root.getPropertyValue(`--${role}`).trim()),
-            })),
+            text: toRgb(root.getPropertyValue('--text')),
+            roles: ROLES.map(role => ({role, colour: toRgb(root.getPropertyValue(`--${role}`))})),
         };
     };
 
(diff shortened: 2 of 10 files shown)


app/src/components/ui/ui-layout.cy.js
```diff
@@ -175,6 +175,147 @@ const sampleTable = <Table>
...
+    const ROLES = ['neutral', 'info', 'success', 'warning', 'danger'];
+
+    const roleColours = () => {
+        const root = getComputedStyle(document.documentElement);
+        const panel = hexToRgb(root.getPropertyValue('--panel').trim());
+        return {
+            panel,
+            text: hexToRgb(root.getPropertyValue('--text').trim()),
+            roles: ROLES.map(role => ({
+                role, colour: hexToRgb(root.getPropertyValue(`--${role}`).trim()),
```
lrnselfreliance: **[bug]** `roleColours()` reads roles with `getPropertyValue(\`--${role}\`)` and feeds the result to `hexToRgb`. In light and dark the roles are authored as aliases (`--info: var(--blue)`, etc.), and custom-property `getPropertyValue` returns that token stream — not a resolved hex. `hexToRgb('var(--blue)')` becomes `rgb(NaN, NaN, NaN)`, so every role collapses to the same string. The “five distinct roles” / “legible against the panel” cases under `themeNames.forEach` are therefore wrong for light/dark (they either fail spuriously or assert NaNs). The Status test a few lines later already does this correctly by reading `getComputedStyle(el).color`, which forces real substitution.

**Suggestion:** Resolve roles the same way Status is measured — paint a throwaway element with `color: var(--info)` (or set a real property to `var(--${role})`) and read the computed color. Keep `hexToRgb` only for tokens that are authored as bare hex (`--panel`, `--text`), or teach it to refuse non-hex input so a regression fails loudly.

app/src/themes/tokens.css
```diff
@@ -50,17 +50,45 @@ html[data-theme="light"] {
...
+ * for `--danger`, not `--red`, and each theme decides how danger looks.
+ *
+ * This matters because two of the four themes have no hues to spend.  Night is red
+ * and nothing else; amber is amber and nothing else.  There, a role is a step on a
+ * BRIGHTNESS ramp -- and it has to be, because `--orange` in night is byte-identical
+ * to `--text` and `--yellow` is byte-identical to `--amber`, so a call site choosing
+ * between hue names there is choosing between synonyms.  That is why all four toast
+ * kinds looked the same in night, and why a warning and an error were the same pixel.
+ *
+ * The ramps below are built so that in every theme:
```
lrnselfreliance: **[suggestion]** The roles note claims that “in every theme” the order `neutral < info < success < warning < danger` is strictly increasing in contrast against the panel, and that `ui-layout.cy.js` asserts “all of that, per theme.” The Cypress suite correctly limits the brightness-order and “danger ≥ text” checks to `monochromeThemes` only, and even comments that light would fail a ramp assertion on a fine design (blue darker than green on white). The token comment over-promises and will mislead the next person who “fixes” light to match the prose.

**Suggestion:** Narrow the prose to monochrome for the ramp / danger-vs-text rules; keep the “distinct + legible” claims for all four themes, matching what the tests actually enforce.

app/src/components/ui/Feedback.tsx
```diff
@@ -35,10 +42,10 @@ export interface MessageProps {
 export function Message({kind = 'info', title, children, icon, onDismiss, className}: MessageProps) {
     return <div
         // Errors and warnings interrupt; info and success are announced politely.
-        role={kind === 'error' ? 'alert' : 'status'}
-        className={['wrolpi-message', kind === 'error' ? 'wrolpi-message-error' : '', className]
+        role={messageRoles[kind] === 'danger' ? 'alert' : 'status'}
+        className={['wrolpi-message', messageRoles[kind] === 'danger' ? 'wrolpi-message-error' : '', className]
             .filter(Boolean).join(' ')}
-        style={{['--message-color' as string]: messageColors[kind]}}
+        style={{['--message-color' as string]: `var(--${messageRoles[kind]})`}}
```
lrnselfreliance, line 48: **[suggestion]** Message now sets `--message-color` from roles, but `.wrolpi-message-error` (unchanged companion rule in `ui.css`) still forces `border: 1.5px dashed var(--red)`. The left accent and icon follow the role; the dashed danger frame still names a hue. In night that is almost the same pixel, so it will not show up in the new distinctness tests, but it breaks the “call sites name meaning, not hue” rule the PR is establishing.

**Suggestion:** Use `var(--message-color)` or `var(--danger)` for the dashed border so the error treatment tracks the role token in every theme.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -175,6 +175,147 @@ const sampleTable = <Table>
...
+                     * being dim IS the signal -- but it still has to be visible at all.
+                     */
+                    expect(contrast(colour, panel), `${role} against the panel`)
+                        .to.be.at.least(role === 'neutral' ? 2.5 : 3);
+                });
+            });
+        });
+
+
+        it(`gives each Status kind a colour of its own in ${theme}`, () => {
```
lrnselfreliance, line 256: **[nit]** The suite proves Status kinds paint four distinct colors and that monochrome role tokens form a ramp, but never mounts a Message row or a set of toasts under night/amber to assert the end of the consumer chain for those two APIs. Unit tests only check Mantine variable *names* on the style attribute. A future toast regression that mapped every type to `danger` while still emitting four distinct CSS variable strings is unlikely, but a regression that reintroduced hue names (`color: 'red'`) would still pass the name-uniqueness test if the four hues differed as strings — and only fail where it matters if someone re-ran visual QA.

**Suggestion:** Optional follow-up: one monochrome case that shows four toast types (or four Messages) and asserts four distinct computed border/indicator colors, mirroring the Status test.

app/src/components/ui/Feedback.tsx
```diff
@@ -137,6 +144,19 @@ export function Progress({
 
 export type StatusKind = 'complete' | 'active' | 'pending' | 'failed';
 
+/*
+ * Status was the one component already doing this correctly: it had per-theme overrides
+ * in ui.css because night has no hues to distinguish four states with.  Those are gone --
+ * the roles carry the brightness ramp now, so this table is the whole of it and every
+ * theme is served by the same rule.
+ */
+const statusRoles: Record<StatusKind, RoleName> = {
```
lrnselfreliance, line 153: **[nit]** Residual risk outside the diff, but relevant to the PR’s goal: several app call sites still hard-code severity hues (`var(--green)` / `var(--amber)` / `var(--red)` / `var(--blue)`) for status-like meaning — e.g. `ControllerPage` SMART health colors and `Files.js` refresh-step icons. Those will still collapse or mis-rank on night/amber the way toasts used to. Not introduced here, but the new roles are exactly the fix those sites need.

**Suggestion:** Track a follow-up to route severity UI through `Status` / role tokens (`var(--success)`, `var(--warning)`, `var(--danger)`, `var(--info)`) rather than palette hues. Button action roles (`primary`/`save`/`retry`/`danger`) can stay on a separate track if product meaning differs from severity.

lrnselfreliance (commented):
## Summary

Solid, well-motivated change: severity goes through role tokens (`neutral` / `info` / `success` / `warning` / `danger`) so night and amber can encode meaning as a brightness ramp instead of collapsing hue synonyms. Toast and Message correctly alias legacy `error` to `danger`, Status drops the old night-only CSS overrides, and the monochrome Cypress assertions (ordering + danger loudness + Status distinctness via computed `color`) match the design claim. The main problems are a test helper that cannot resolve light/dark role aliases, a sample-page caption that still claims the bug this PR fixes, and a few leftover hue call sites / docs that undercut the new contract.

## Issue counts by severity

- bugs: 2
- suggestions: 5
- nits: 2

## Issues outside the diff

- **[bug]** The Toasts panel still tells the reader that “night has no second hue to spend, so info, success, warning and error cannot be told apart by colour there.” That is the pre-fix failure mode. Immediately below, the new Semantic roles section explains that those kinds are now distinct brightness steps. The sample page is the manual QA surface for this change; leaving the old caption will train people to expect the bug and ignore a regression.
  **Suggestion:** Rewrite the toast caption to match the new model, e.g. that night/amber no longer share one pixel across kinds because toast types map to roles, and point at the roles strip for the ramp.

- **[suggestion]** `Status`’s public JSDoc still says night encodes status by brightness “see ui.css”. The night class overrides were removed in this PR; encoding now lives entirely in `statusRoles` + role tokens. Callers reading the doc will go looking for CSS that no longer exists.
  **Suggestion:** Point at `statusRoles` / the roles note in `tokens.css`, and say every theme uses the same rule.

- **[suggestion]** The “gives each type a colour of its own” comment still frames night and amber as a deliberate exception where kinds may collapse, “decided by the token tables, not here.” After this PR the token tables are exactly what prevent collapse; the unit test only checks that four different Mantine color *names* are set, which is still useful, but the comment re-teaches the old bug as design.
  **Suggestion:** Say the unit test locks the type→role name mapping, and that resolved luminance / ramp checks live in `ui-layout.cy.js` for monochrome themes.

- **[suggestion]** Mantine’s form/error chrome still bridges through `--mantine-color-error: var(--red)`. Role tokens now exist and are registered as Mantine colors, so validation and severity can diverge slightly in night/amber (`--red` vs `--danger` are close but not the same step on the ramp). Any Mantine path that paints “error” will skip the role table this PR introduces.
  **Suggestion:** Point `--mantine-color-error` at `var(--danger)` (or register and use the `danger` color name) so built-in error UI rides the same ramp as toasts and messages.

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Checked this directly, and the premise does not hold: custom properties are substituted at **computed-value time**, so `getPropertyValue` returns the resolved hex, not the token stream. Measured in Chrome on the running app:

```
light  ["#6a7072","#20629c","#3f7a2e","#95661a","#b23a3a"]
dark   ["#93999b","#5c9dd6","#7fb069","#d6a656","#e06c6c"]
night  ["#992c2c","#c02929","#e02323","#ee3434","#ff3e3e"]
```

Corroborated by the tests themselves: had every role collapsed to `rgb(NaN, NaN, NaN)`, `new Set(colours).size` would be 1 and the distinctness case would fail in light and dark. It passes with 5.

The defensive half of the suggestion is worth taking though, and I did — reads now go through a strict `toRgb` that refuses anything not already resolved. The reason is your NaN observation generalised: `expect(NaN).to.not.equal(x)` **passes**, so a test built on an unresolved value would not fail, it would silently assert nothing. Better to throw naming the value. It earned its keep immediately, catching a hex-vs-rgb mismatch in the new toast test I added below.

app/src/themes/tokens.css
lrnselfreliance: Right, and this one would have done real damage. The prose promised a ramp "in every theme" while the tests deliberately exempt light and dark — so the next person reconciling the two would most likely have "fixed" light, which means reordering light's hues to satisfy a test.

Narrowed to the split the tests actually enforce (distinct + 3:1 everywhere; ordering and danger-vs-text for monochrome only), with that reasoning stated so the exemption reads as deliberate rather than as an oversight.

app/src/components/ui/Feedback.tsx
lrnselfreliance: Correct — `.wrolpi-message-error` still forced `border: 1.5px dashed var(--red)` while the accent and icon on the same element had moved to the role. Now follows `var(--message-color)`, so the frame cannot disagree with the accent.

Your parenthetical is the important part: in night that is *almost the same pixel*, so no colour test could ever have caught it. I went looking for siblings and found six more library rules naming a hue for something that means danger — the night danger button, invalid form fields, invalid paths, path error text, danger panels, and failed table rows.

All converted, plus a source guard that fails any rule whose **selector** means severity and whose body names a severity hue. Focus rings and active-tab accents deliberately keep `--blue`: those are the primary accent, not a severity, which is why `--primary` is a separate token.

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Took this, and it caught me out — in the good way.

I added the four-toast and four-Message monochrome cases, planted your exact hypothetical (`toastRoles` back to `blue/green/yellow/red`), and **the distinctness assertion stayed green**. In night those four hues are four different values: 2.29, 3.14, 4.52, 5.64 against a panel.

Which means the headline claim in my PR description was wrong. The four toast kinds were never "the same pixel" in night — they were cramped and badly ranked, with info at 2.29:1, below the 3:1 floor. The byte-identical pairs (`--orange`/`--text`, `--yellow`/`--amber`) are real but bite a call site reaching for those names, not the toast mapping. I have corrected the PR body.

So distinctness is not the assertion with teeth; the **legibility floor** is. The test now measures every kind against the toast's own surface, and your hypothetical regression fails it naming the number: `expected 2.292758105196531 to be at least 3`.

app/src/components/ui/Feedback.tsx
lrnselfreliance: Agreed, and this is the follow-up already flagged in the PR description — roughly 24 app-level `color=` sites plus the inline `var(--green)`/`var(--red)` uses you name in `ControllerPage` and `Files.js`.

Deliberately out of scope here: each needs a judgement about which role it means, and at least two already disagree with each other (`Downloads.js` colours the same "Download error" label `orange` on line 241 and `red` on line 456). Worth its own diff where those decisions are reviewable rather than buried under a token change.

Also agreed on the separation — Button roles (`primary`/`save`/`retry`/`danger`) are product actions, not severities, and stay on their own track. That is the same reason `--primary` is not one of the five.

PR #611, opened 2026-08-02
ui: route genuine success and failure through the roles

lrnselfreliance:
Sixteen call sites, chosen by reading each one rather than grepping for a hue. I had four sub-agents judge them independently against one rule: **convert only what genuinely means success or failure**; identification, category and decoration stay hue-named.

## Converted

| Site | Was | Now |
|---|---|---|
| SMART drive health — all four branches | green/amber/red/muted | success/warning/danger/neutral |
| Config validity: valid / invalid / empty | violet/red/muted | success/danger/neutral |
| "Failed to load preview" | red | danger |
| Both "Download error" triggers | orange **and** red | danger |
| Expired inventory items (desktop + mobile) | red | danger |
| Refresh phases: complete / active / not reached | green/blue/muted | success/info/neutral |
| Refresh Progress in its error state | red | danger |
| Cookie status: Unlocked / Locked | green/amber | success/warning |
| Flasher chip-mismatch label | red | warning |
| CPU load thresholds (`LoadStatistic`) | orange/red | warning/danger |
| `useAPIButton` outcome | green/red | success/danger |
| Gallery "Danger zone" + load readings | red/orange | danger/warning |

## Left alone, deliberately

- **The file browser toolbar** — every action has its own colour so they can be told apart, and delete already uses `role='danger'`. Your ruling.
- **Diff add/remove** in the config and inventory import previews. Adding is not a success and removing is not a failure — in Overwrite mode the removals are precisely what the user asked for, and painting them `danger` would report their intended action as broken. One agent also measured that in night `--green`/`--red` are *further apart* than `--success`/`--danger`, so converting would make the add/remove distinction harder to see, not easier.
- Seasons in `DatesSelector`, the uploader star and favourite heart, `TagIcon`, downloader category buttons, "Recommended to Keep".

## Three sites were actually wrong

- Both **"Download error"** triggers guard on the *same condition* and open the *same modal*, but one was `orange` and the other `red`. And `orange` is the worst possible choice — byte-identical to `--text` in night, so that error button was indistinguishable from an ordinary control.
- **Config validity** drew its valid state in `--violet`. No user parses lavender as "valid".
- **`LoadStatistic`** — the thing users actually watch on the Status page — rendered its warning threshold in that same invisible orange.

## Two notes on the tests

**My first `LoadStatistic` test was vacuous.** It rendered the component and asserted the inline colour — but jsdom *rejects* `color: var(--danger)` as an invalid declaration and drops it, so every colour reads back as the empty string. Two of the four assertions passed for that reason alone and would have kept passing with the hues restored. It is a source guard now, with the painted result asserted in a browser.

**That browser test then found something real, which I have not fixed.** An uncoloured reading inherits `--text`, which in night is **4.98:1** against `--warning`'s **4.91:1** — so a machine under load is marginally *quieter* than one that is fine.

The fix is a palette change, not a call-site one. On a dark panel the only way up is lightness, and night's danger would desaturate from `#ff3e3e` to about `#ff5d5d` — trading the theme's "red only" character for loudness. That is your call, so the test documents it and asserts what does hold (three distinct readings; danger louder than both) rather than my quietly making it.

## Verified

- jest **1058 passing** / 59 suites
- `ui-layout.cy.js` **113 passing** (was 111)
- `npm run build` clean
- every conversion checked red against a planted reversion

### DIFF
--- a/app/src/components/Common.test.js
+++ b/app/src/components/Common.test.js
@@ -1,3 +1,5 @@
+import fs from 'fs';
+import path from 'path';
 import React from 'react';
 import {act, render, screen, waitFor} from '../test-utils';
 import userEvent from '@testing-library/user-event';
@@ -558,5 +560,48 @@ describe('contrastingColor with shorthand hex', () => {
                                      ['#f00', '#ff0000'], ['#1b2', '#11bb22']]) {
             expect(contrastingColor(short)).toBe(contrastingColor(long));
         }
+    });
+});
+
+describe('LoadStatistic ranks load by role, not by hue', () => {
+    /*
+     * A source guard, not a render test, and deliberately so: jsdom's CSS parser REJECTS
+     * `color: var(--danger)` as an invalid declaration and drops it, so every inline colour
+     * reads back as the empty string here.  A render test would pass whatever the component
+     * did -- the first version of this asserted `''` for the below-threshold cases and would
+     * have gone on passing with the hues restored.
+     *
+     * What is painted is asserted in ui-layout.cy.js, in a browser, at the thresholds.
+     */
+    it('names roles rather than hues at both thresholds', () => {
+        const source = fs.readFileSync(path.join(__dirname, 'Common.js'), 'utf8');
+        const helper = source.slice(
+            source.indexOf('export function LoadStatistic'),
+            source.indexOf('export function', source.indexOf('export function LoadStatistic') + 10));
+
+        expect(helper).toContain("color = 'danger'");
+        expect(helper).toContain("color = 'warning'");
+        expect(helper.match(/'(red|green|amber|yellow|orange)'/)).toBeNull();
+    });
+});
+
+describe('a drive\'s SMART assessment is a severity, not a colour', () => {
+    /*
+     * A source guard: `getHealthColor` is module-private, and exporting it only so a test
+     * can reach it would be shaping the code around the test.  What matters is that the
+     * helper stays on the ramp -- all four branches or none, since a helper where PASS and
+     * FAIL rank but WARN is an unranked hue is worse than either choice.
+     */
+    it('maps every branch to a role', () => {
+        const source = fs.readFileSync(
+            path.join(__dirname, 'admin', 'ControllerPage.js'), 'utf8');
+        const helper = source.slice(
+            source.indexOf('const getHealthColor'),
+            source.indexOf('};', source.indexOf('const getHealthColor')));
+
+        expect(helper).toContain("'var(--success)'");
+        expect(helper).toContain("'var(--warning)'");
+        expect(helper).toContain("'var(--danger)'");
+        expect(helper.match(/var\(--(red|green|amber|yellow|orange)\)/)).toBeNull();
     });
 });
--- a/app/src/components/Common.test.js
+++ b/app/src/components/Common.test.js
@@ -1,11 +1,11 @@
-import fs from 'fs';
-import path from 'path';
 import React from 'react';
 import {act, render, screen, waitFor} from '../test-utils';
 import userEvent from '@testing-library/user-event';
 import {
-    contrastingColor, contrastRatio, DirectorySearch, SearchResultsInput, TAG_TEXT_DARK, TAG_TEXT_LIGHT,
+    contrastingColor, contrastRatio, DirectorySearch, loadRole, SearchResultsInput,
+    TAG_TEXT_DARK, TAG_TEXT_LIGHT,
 } from './Common';
+import {healthRole} from './admin/ControllerPage';
 
 // Mock debounce to make tests run faster
 jest.mock('lodash/debounce', () => jest.fn(fn => {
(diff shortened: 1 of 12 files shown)


app/src/components/ThemeSamplePage.js
```diff
@@ -197,7 +197,7 @@ export function ThemeSamplePage() {
             <StatisticGroup>
                 <Statistic value='0.4' label='1 Min. Load'/>
                 <Statistic value='2.6' label='5 Min. Load' color='orange'/>
```
lrnselfreliance: **[bug]** The "Coloured readings, as Status draws them" sample updates 15 Min Load to `color='danger'` but leaves 5 Min Load on `color='orange'`. That is the half-migration the rest of the PR is fixing: in night/amber `orange` collapses toward ordinary text, so the sample page itself still cannot demonstrate that a mid-severity load ranks between fine and danger.

**Suggestion:** Change to `color='warning'` so the trio is uncoloured / warning / danger, matching `LoadStatistic` and the new Cypress fixture.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -333,6 +333,56 @@ describe('semantic roles read as severity in every theme', () => {
...
+     * load rendered its warning as an ordinary uncoloured number.
+     *
+     * Only a browser can see this.  jsdom rejects `color: var(--warning)` as invalid and
+     * drops it, so the inline colour reads back empty whatever the component did.
+     */
+    monochromeThemes.forEach((theme) => {
+        it(`separates fine, warning and danger loads in ${theme}`, () => {
+            cy.mountUI(
+                <Panel>
+                    <Statistic value='1.9' label='fine' cores={4} color={undefined}/>
```
lrnselfreliance: **[suggestion]** The new suite is titled and commented as covering `LoadStatistic` "at the thresholds", and `Common.test.js` explicitly defers paint checks to this file. The fixture never imports or mounts `LoadStatistic`; it hard-codes `<Statistic color='warning'/>` / `color='danger'`. That only re-proves role tokens resolve distinctly (already covered by the earlier "semantic roles read as severity" block) and would stay green if `LoadStatistic` were reverted to `orange`/`red` while the fixture stayed on roles. The unit "source guard" is the only thing holding the ranking strings.

**Suggestion:** Mount the real helper at the documented thresholds, e.g. `<LoadStatistic value={1.9} cores={4}/>`, `<LoadStatistic value={2.6} cores={4}/>`, `<LoadStatistic value={4.1} cores={4}/>`, and assert the three painted values. Keep the role-distinctness checks. Soften the `Common.test.js` comment so it does not claim thresholds are browser-tested until they are.

app/src/components/Common.test.js
```diff
@@ -560,3 +562,46 @@ describe('contrastingColor with shorthand hex', () => {
...
+    /*
+     * A source guard, not a render test, and deliberately so: jsdom's CSS parser REJECTS
+     * `color: var(--danger)` as an invalid declaration and drops it, so every inline colour
+     * reads back as the empty string here.  A render test would pass whatever the component
+     * did -- the first version of this asserted `''` for the below-threshold cases and would
+     * have gone on passing with the hues restored.
+     *
+     * What is painted is asserted in ui-layout.cy.js, in a browser, at the thresholds.
+     */
+    it('names roles rather than hues at both thresholds', () => {
```
lrnselfreliance: **[nit]** The source-guard tests are a reasonable defense against hue regressions given jsdom's `var()` blind spot, but they are brittle string slices (`indexOf('export function', …)`, `indexOf('};', …)`) and only cover two helpers. They will not notice a new severity branch added outside the slice, and they do not execute the ranking logic (cores thresholds, assessment strings).

**Suggestion:** Prefer exporting tiny pure mappers (`loadRole(value, cores)`, `healthRole(assessment)`) for direct unit tests of branch→role, and keep one thin source guard or Cypress paint test for integration. Not blocking if Issue 4 lands.

lrnselfreliance (commented):
## Summary

This PR correctly migrates several high-value severity call sites (`LoadStatistic`, `useAPIButton`, SMART assessment, download errors, inventory expiry, cookies status, refresh phase markers) onto the role tokens from PR #610, and the meaning→role choices it does make are sound (e.g. chip mismatch and shadowed-mount caution as `warning`, not `danger`). The gap is completeness: the Status page still ranks temperature, IO wait, CPU/memory/drive usage, and bandwidth with hard-coded hues sitting next to the newly-role-aware `LoadStatistic`, and the same half-migration shows up in ThemeSamplePage (`orange` left next to `danger`) and in SMART sector cells still on `var(--amber)`. Tests lean on source-string guards and a Cypress check that never mounts `LoadStatistic`, so the ranking contract is only weakly enforced.

## Issue counts by severity

- bugs: 3
- suggestions: 3
- nits: 1

## Issues outside the diff

- **[bug]** app/src/components/admin/Status.js:29 — The PR migrates `LoadStatistic` (the Status page CPU-load reading) to `warning`/`danger`, but every other severity-ranked reading on the same page still hard-codes palette hues: `DriveInfo` (`red`/`orange` at 90%/80%), `DiskBandwidthProgress` (`red`/`orange`/`yellow`), `CPUTemperatureStatistic` (`red`/`orange`), `IOWaitStatistic` (`red`/`orange`), `BandwidthProgress` (`yellow`/`red`), `CPUUsageProgress` (`red`/`brown`/`orange`), `MemoryUsageProgress` (`red`/`orange`). In night mode `--orange` is byte-identical to `--text`, so a hot CPU, high IO wait, or 85%-full drive still paints as an ordinary uncoloured bar/number — the exact failure mode the PR description and the new Cypress block call out for load. Leaving these next to a role-aware load reading is an incomplete product fix, not a later follow-up of equal priority.
  **Suggestion:** Map each threshold the same way `LoadStatistic` did: high → `danger`, mid → `warning` (and, where a third soft band is needed, prefer `warning` over `yellow`/`brown` rather than inventing a sixth role). Drop `yellow`/`brown` as intermediate steps unless they carry non-severity meaning. Example for temp/IO wait:

```js
if (/* critical */) color = 'danger';
else if (/* high */) color = 'warning';
```

Same for drive/CPU/memory progress. While touching `BandwidthProgress`, also fix the pre-existing unreachable branch (`if (percent > 70) … else if (percent > 90)` never paints red).

- **[bug]** app/src/components/admin/ControllerPage.js:681 — `getHealthColor` was correctly moved onto `var(--success|warning|danger|neutral)`, and a source guard freezes that helper — but the three SMART attribute cells in the same modal still colour non-zero reallocated / pending / uncorrectable sectors with `var(--amber)`. Those are severity signals (the same WARN band the helper comment says must not stay on an unranked hue). In night, `--amber` is not on the role ramp the helper just joined, so a drive can show PASS/WARN/FAIL correctly while its bad-sector counts still use a hue synonym.
  **Suggestion:** Use `var(--warning)` for `> 0` on all three cells (lines 681, 689, 697). Optionally extend the source guard to the sector-cell block, or share a tiny helper so the modal cannot drift again.

- **[suggestion]** app/src/components/admin/Configs.js:209 — This file was partially migrated (`WarningIcon` and failed-preview use roles), but the backup preview headers still encode add/remove meaning with `var(--green)` / `var(--red)`. The same add/remove pattern remains in `InventoryImportModal.js` (`var(--green)` / `var(--red)`). Those are not decorative accents; they label constructive vs destructive outcomes of an import — the role contract's `success` / `danger`.
  **Suggestion:** `var(--success)` for "to add", `var(--danger)` for "to remove" in both Configs and InventoryImportModal. Prefer `Header`'s `color` prop (`color='success'`) over inline styles where it already resolves `var(--${color})`.

- **[suggestion]** app/src/components/FileBrowser.js:888 — Upload progress status colouring in `FileBrowser.js` and the twin logic in `Upload.js` still map complete/failed/conflicting to `green`/`red`/`orange`. That is the same complete/failed/warning shape `Status` and the refresh steps already express with roles (`success` / `danger` / `warning`). Missed genuine severity call sites outside the Status page.
  **Suggestion:** `complete → success`, `failed → danger`, `conflicting → warning`, pending stay uncoloured or `neutral`/`grey` as today. Apply the same table in both FileBrowser and Upload to avoid drift.

app/src/components/ThemeSamplePage.js
lrnselfreliance: Caught me — and in the worst possible place. That sample's own caption tells the reader to go and check night and amber, which is exactly where `orange` is byte-identical to `--text`. So the one demo whose job is to show a mid-severity reading was demonstrating the bug instead of the fix.

Fixed: the trio is now uncoloured / `warning` / `danger`, matching what `LoadStatistic` actually produces.

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Correct, and this is the important one. The suite was titled "a load reading ranks itself", commented as covering the thresholds, and cross-referenced from `Common.test.js` as the place the painted result gets checked — while mounting a hand-built `<Statistic color='warning'/>`. It proved only that role tokens resolve distinctly, which the roles suite two blocks up already covers.

Same shape as the column-border test in #609: a claim that outran its measurement. I keep generating this failure mode when I write the assertion before deciding what the subject is.

It mounts `<LoadStatistic value={1.9|2.6|4.1} cores={4}/>` now. Verified with your exact scenario — reverting the helper to `orange`/`red` fails it, because in night `orange` resolves to `--text` and the warning reading becomes indistinguishable from the plain one, collapsing three distinct readings to two. It did not fail before.

The `Common.test.js` cross-reference is gone entirely, along with the guards it belonged to — see the next comment.

app/src/components/Common.test.js
lrnselfreliance: Took the suggestion as written. `loadRole(value, cores)` and `healthRole(assessment)` are exported pure functions now, and the source guards are deleted.

Your criticism was right on every count: they were `indexOf` slices, they covered two helpers, and they never executed a branch. Worth adding that they were also a bad answer to the problem that produced them — jsdom rejects `color: var(--danger)` and drops it, so the mapping cannot be read off a rendered component. Extracting the mapping is the honest fix; asserting on source text was me working around the harness instead of the design.

The direct tests reach things the guards could not: either side of both load thresholds, lower-case input (the value comes from `smartctl`), and the one that actually matters — an unreadable drive must report `neutral`, never `success`. That last case is a real hazard the guard could never have expressed, and it goes red when planted.

PR #612, opened 2026-08-02
ui: headings open their section, and keep their help icon on one line

lrnselfreliance:
Two Header issues. The second is a regression this migration caused.

## Headings sat flush against what came before

`.wrolpi-header` had `margin: 0 0 10px` — **no top margin at all**. On `/more/statistics`, "Tags" touched the panel above it with a measured **0px gap**, so it read as part of that panel rather than as the start of the next section.

Now 22px above, 10px below — a heading belongs to what follows it.

`* + .wrolpi-header` rather than a blanket margin, for the same reason `StatisticGroup` needed it: a heading that opens a Panel or a page is already where it should be, and pushing it down just dents the top of its own container. Both cases are asserted.

## Every help icon had dropped to its own line

Clearest on "Root CA Certificate" in Settings. `.inline-header` set `h1..h5 { display: inline-block }` so the heading and a sibling `<span>` shared a row — which worked while these were a bare `<h2>`, and stopped working the moment they gained a block wrapper: the `<label>`'s block child forced a break.

Fixed in the library rather than patched in `App.css`:

- `Header` takes an **`after`** slot, rendered *inside* the heading element where the existing flex row already handles alignment and spacing
- `HelpHeader` and `InfoHeader` pass their icon through it
- the `<label>` moves inside the heading, so `for_` keeps associating the text with its field — three call sites rely on that
- the `.inline-header` rules are deleted

**Fifteen call sites** are fixed by fixing those two wrappers.

## Tests

Containment in jest (`toContainElement`), and the parts jsdom cannot judge in a browser — that the two boxes actually *share a line*, and that the heading stays one row tall. Containment alone would not have caught this: the icon was always in the DOM, just below.

Planted every original shape:

| Plant | Result |
|---|---|
| top margin removed | RED — flush heading returns |
| blanket margin instead of `* +` | RED — dents its own container |
| control moved back outside the heading | RED in both suites |

## Verified

- jest **1070 passing** / 59 suites
- `ui-layout.cy.js` **120 passing** (was 113)
- `npm run build` clean
- checked on `/more/statistics`, `/admin/settings` and the gallery; swept eight routes for headings whose spacing the blanket-ish rule might have disturbed — none
- gallery gained a header with a help icon

Pre-existing cypress failures unchanged at 12.

### DIFF
--- a/app/src/components/Common.js
+++ b/app/src/components/Common.js
@@ -810,14 +810,22 @@ export function InfoHeader({
                                required = false,
                                ...props
                            }) {
-    return <div className='inline-header' {...props}>
-        <label htmlFor={for_}>
-            <Header as={headerSize}>{headerContent} {required && <RequiredAsterisk/>}</Header>
-        </label>
-        <span>
-            <InfoPopup content={popupContent} iconSize={iconSize} icon={icon} position={popupPosition} {...popupProps}/>
-        </span>
-    </div>
+    /*
+     * The icon goes in the heading's own `after` slot.  It used to be a sibling `<span>`
+     * beside a `<label>`, held on one line by `.inline-header h1..h5 {display: inline-block}`
+     * -- which worked while these headings were a bare `<h2>` and stopped working the moment
+     * they gained a block wrapper: the label's block child forced a break and every help icon
+     * in the app dropped to its own line.
+     *
+     * The label stays, now inside the heading, because `for_` associates the text with a
+     * field on three of these call sites.
+     */
+    return <Header as={headerSize} {...props} after={
+        <InfoPopup content={popupContent} iconSize={iconSize} icon={icon} position={popupPosition}
+                   {...popupProps}/>
+    }>
+        <label htmlFor={for_}>{headerContent} {required && <RequiredAsterisk/>}</label>
+    </Header>
 }
 
 export function HelpModal({
@@ -856,14 +864,13 @@ export function HelpHeader({
                                 required = false,
                                 ...props
                             }) {
-    return <div className='inline-header' {...props}>
-        <label htmlFor={for_}>
-            <Header as={headerSize}>{headerContent} {required && <RequiredAsterisk/>}</Header>
-        </label>
-        <span>
-            <HelpModal helpPath={helpPath} iconSize={iconSize} icon={icon} title={helpTitle} {...helpModalProps}/>
-        </span>
-    </div>
+    // See the note in InfoHeader: the icon belongs in the heading's row, not beside it.
+    return <Header as={headerSize} {...props} after={
+        <HelpModal helpPath={helpPath} iconSize={iconSize} icon={icon} title={helpTitle}
+                   {...helpModalProps}/>
+    }>
+        <label htmlFor={for_}>{headerContent} {required && <RequiredAsterisk/>}</label>
+    </Header>
 }
 
 // All the hardware toggles share this shape: a Toggle whose state comes from a
--- a/app/src/components/Common.js
+++ b/app/src/components/Common.js
@@ -820,11 +820,24 @@ export function InfoHeader({
      * The label stays, now inside the heading, because `for_` associates the text with a
      * field on three of these call sites.
      */
+    /*
+     * A `<label>` ONLY when `for_` names a field.  Several call sites pass an InfoHeader as
+     * the `label` prop of InputForm/ToggleForm, which wraps whatever it is given in a
+     * `<label>` of its own -- an unconditional label here nests one inside another, which is
+     * invalid HTML and confuses what clicking the text does.  That predates this change; it
+     * is fixed here rather than carried forward.
+     *
+     * `margin: 0` on the icon because the row's `gap` does the spacing now.  The 0.5em
+     * default was there to separate a free-floating sibling, and inside the row it adds to
+     * the gap on every side.
+     */
     return <Header as={headerSize} {...props} after={
         <InfoPopup content={popupContent} iconSize={iconSize} icon={icon} position={popupPosition}
-                   {...popupProps}/>
+                   iconStyle={{margin: 0}} {...popupProps}/>
     }>
-        <label htmlFor={for_}>{headerContent} {required && <RequiredAsterisk/>}</label>
+        {for_
+            ? <label htmlFor={for_}>{headerContent} {required && <RequiredAsterisk/>}</label>
+            : <>{headerContent} {required && <RequiredAsterisk/>}</>}
     </Header>
 }
 
@@ -864,12 +877,15 @@ export function HelpHeader({
                                 required = false,
                                 ...props
                             }) {
-    // See the note in InfoHeader: the icon belongs in the heading's row, not beside it.
+    // See the notes in InfoHeader: the control belongs in the heading's row, a label only
+    // when it names a field, and the row's gap is the spacing.
     return <Header as={headerSize} {...props} after={
         <HelpModal helpPath={helpPath} iconSize={iconSize} icon={icon} title={helpTitle}
-                   {...helpModalProps}/>
+                   iconStyle={{margin: 0}} {...helpModalProps}/>
     }>
-        <label htmlFor={for_}>{headerContent} {required && <RequiredAsterisk/>}</label>
+        {for_
+            ? <label htmlFor={for_}>{headerContent} {required && <RequiredAsterisk/>}</label>
+            : <>{headerContent} {required && <RequiredAsterisk/>}</>}
     </Header>
 }
 
--- a/app/src/components/Common.test.js
+++ b/app/src/components/Common.test.js
@@ -2,8 +2,8 @@ import React from 'react';
 import {act, render, screen, waitFor} from '../test-utils';
 import userEvent from '@testing-library/user-event';
 import {
-    contrastingColor, contrastRatio, DirectorySearch, loadRole, SearchResultsInput,
-    TAG_TEXT_DARK, TAG_TEXT_LIGHT,
+    contrastingColor, contrastRatio, DirectorySearch, HelpHeader, InfoHeader, loadRole,
+    SearchResultsInput, TAG_TEXT_DARK, TAG_TEXT_LIGHT,
 } from './Common';
 import {healthRole} from './admin/ControllerPage';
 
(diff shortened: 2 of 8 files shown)


app/src/components/Common.js
```diff
@@ -810,14 +810,22 @@ export function InfoHeader({
...
+     * The icon goes in the heading's own `after` slot.  It used to be a sibling `<span>`
+     * beside a `<label>`, held on one line by `.inline-header h1..h5 {display: inline-block}`
+     * -- which worked while these headings were a bare `<h2>` and stopped working the moment
+     * they gained a block wrapper: the label's block child forced a break and every help icon
+     * in the app dropped to its own line.
+     *
+     * The label stays, now inside the heading, because `for_` associates the text with a
+     * field on three of these call sites.
+     */
+    return <Header as={headerSize} {...props} after={
```
lrnselfreliance, line 834: **[suggestion]** `InfoHeader` / `HelpHeader` place `InfoPopup` / `HelpModal` in the new flex `after` row (which already has `gap: 8px` on `.wrolpi-header-text`), but those controls still default to `iconStyle = { margin: '0.5em' }` (`InfoPopup` / `HelpModal`). That margin was useful when the icon was a free sibling after the heading; inside the flex row it adds ~0.5em on all sides on top of the gap, so the control can sit farther from the title and slightly less vertically snug than the Cypress tests (which mount a plain `Button`, not these defaults) imply.

**Suggestion:** From `InfoHeader` / `HelpHeader`, pass `iconStyle={{ margin: 0 }}` (and `HelpModal` `iconStyle` the same way), or change the defaults when used inside a header row. Optionally add a Cypress case that mounts real `HelpHeader` / `InfoHeader` so the default icon chrome is covered.

app/src/components/ui/ui.css
```diff
@@ -743,6 +743,30 @@ html[data-theme="night"] .wrolpi-tag::before {
...
+ */
+* + .wrolpi-header {
+    margin-top: 22px;
+}
+
+/*
+ * Trailing content keeps the heading's baseline but not its size: a help button rendered
+ * at heading scale is a 26px glyph next to a 17px word.
+ */
+.wrolpi-header-after {
```
lrnselfreliance, line 777: **[suggestion]** `.wrolpi-header-after` is a flex item with default `flex-shrink: 1`. On a narrow width or with a long title, the trailing help control can compress while the text span competes for space. The Cypress layout suite only uses the short title “Root CA Certificate”, so this edge case is untested.

**Suggestion:** Add `flex-shrink: 0` (and optionally `min-width: 0` on the text `<span>` if you want long titles to wrap/ellipsis instead of overflowing). Worth a narrow-viewport Cypress assert if you care about mobile settings forms.

app/src/components/Common.js
lrnselfreliance, line 834: **[suggestion]** `...props` used to land on the outer `div.inline-header`. They now go through `Header`, which puts `className` on `.wrolpi-header` but `style` on the inner `<h*>` (`Surfaces.tsx`). The only in-tree `style` call site is `InfoHeader` with `style={{marginBottom: '1em'}}` on Status (“Drive Bandwidth”). That may still look acceptable via margin collapse with `.wrolpi-header { margin: 0 0 10px }`, but it is no longer a reliable way to space the whole header block (collapse breaks if the wrapper ever gains padding, border, overflow, or flex). Same class of issue for any future `style` / layout props on these helpers.

**Suggestion:** Either document that layout styles belong on a wrapper around `Header`, or have `InfoHeader`/`HelpHeader` (or `Header` itself) apply `style` to the outer `.wrolpi-header` when the intent is block spacing. A quick visual check of Status → Drive Bandwidth is enough to confirm this call site.

app/src/components/ui/ui.css
```diff
@@ -743,6 +743,30 @@ html[data-theme="night"] .wrolpi-tag::before {
...
+/*
+ * A heading opens a section, so it needs more air above it than below — it belongs to
+ * what follows, not to what came before.  With no top margin at all a heading sat flush
+ * against the panel above it, which read as part of that panel rather than as the start
+ * of the next thing (visible on /more/statistics, where "Tags" touched the panel above).
+ *
+ * `* +` rather than a blanket margin: a heading that opens a Panel or a page is already
+ * where it should be, and pushing it down would just dent the top of its own container.
+ */
+* + .wrolpi-header {
```
lrnselfreliance, line 755: **[suggestion]** `* + .wrolpi-header { margin-top: 22px }` is the right global default and is well tested for “first child of Panel stays flush”. Several existing call sites still pass `style={{ marginTop: '1em' }}` / `'1.5em'` on `Header` (e.g. ExtensionPage “After installing” / “Install steps”, calculators, ThemeSamplePage). Those styles hit the inner heading, not the wrapper; with adjoining-margin collapse the visual result may be close to `max(22px, N em)`, but it is redundant and fragile next to the new rule.

**Suggestion:** Follow-up sweep to drop now-redundant `marginTop` on `Header` call sites (or keep them only where you intentionally want more than 22px and apply them to a wrapper / the outer class). Not a blocker for this PR.

app/src/components/Common.js
```diff
@@ -810,14 +810,22 @@ export function InfoHeader({
...
+        <InfoPopup content={popupContent} iconSize={iconSize} icon={icon} position={popupPosition}
+                   {...popupProps}/>
+    }>
+        <label htmlFor={for_}>{headerContent} {required && <RequiredAsterisk/>}</label>
```
lrnselfreliance: **[suggestion]** The label is always rendered (`<label htmlFor={for_}>...</label>`), even when `for_` is null. That is fine for the three Download.js `for_` sites, but several call sites pass `InfoHeader` *as* the `label` prop of `InputForm` / `ToggleForm` (Videos settings), which already wrap content in their own `<label>`. Result: nested `<label>` (invalid HTML) and an extra label inside a heading inside a label. Nested labels were already a problem with the old `inline-header` structure; this PR preserves the always-on `<label>` while moving it inside the heading.

**Suggestion:** Render a `<label>` only when `for_` is set; otherwise render the text (and optional asterisk) as a plain node. That keeps the association where it is needed and avoids nested labels when `InfoHeader` is used as a form label.

app/src/components/Common.test.js
```diff
@@ -622,3 +622,49 @@ describe('healthRole', () => {
             expect(healthRole(value)).toBe('neutral'));
     });
 });
+
+describe('a heading with a help icon keeps the icon on its own line', () => {
```
lrnselfreliance: **[nit]** The describe name says “keeps the icon on its own line”, which usually means a *separate* line — the opposite of the fix. The Cypress suite (`ui-layout.cy.js`) uses the same “on its own line” wording for “beside the text”.

**Suggestion:** Rename to something like “keeps the help icon on one line with the heading” / “keeps its trailing control on the heading’s row” so failures and history read correctly.

app/src/components/ui/Surfaces.tsx
```diff
@@ -51,6 +59,7 @@ export function Header({
                 ? <Icon name={icon} size='medium'/>
                 : <Icon component={icon} size='medium'/>)}
             <span>{children}</span>
+            {after && <span className='wrolpi-header-after'>{after}</span>}
```
lrnselfreliance: **[nit]** Putting the help `IconButton` inside the heading means the heading’s accessible name often becomes “Root CA Certificate Help” (button name included). That is a common tradeoff when trailing controls share the heading row, and focus order remains usable, but screen-reader heading navigation is slightly noisier than text-only headings.

**Suggestion:** Acceptable as-is for this UX. If it bothers you later, keep the control in the flex row visually but move it outside the `<h*>` in the DOM (e.g. sibling inside `.wrolpi-header` with the flex row on the wrapper) so the heading name stays clean without reintroducing the wrap bug.

lrnselfreliance (commented):
## Summary

Solid, well-motivated UI fix: putting trailing help/info controls in a Header `after` slot correctly addresses the block-wrapper wrap that dropped every help icon onto its own line, and the sibling-only top margin (`* + .wrolpi-header`) matches the “heading opens a section” intent without denting panel tops. Tests are above average for a layout change — jest asserts real DOM containment and `for_` association, and Cypress uses getBoundingClientRect for the layout claim jsdom cannot judge. Residual risks are mostly spacing/back-compat details (default icon margins stacking on the new flex row, `style` no longer targeting an outer wrapper, long-title flex shrink) rather than a broken approach.

## Issue counts by severity

- bugs: 0
- suggestions: 5
- nits: 2

app/src/components/Common.js
lrnselfreliance: Both parts taken. `iconStyle={{margin: 0}}` from both wrappers — the row's `gap` is the spacing now.

The second half of your suggestion was the more important one. My cypress test mounted a plain `Button` in an `after` slot, so the default icon chrome was precisely what it could not see. It mounts a real `HelpHeader` now.

Worth recording that my first assertion for it was still wrong: I measured the gap to `.wrolpi-header-after`, which is a flex container — a margin on the button sits *inside* it and the wrapper's own rect never moves, so the check could not detect the margin at all. Measuring the button's rect makes it fail with the default restored: `expected 16 to be at most 12`, which is exactly the 8px gap plus 8px margin.

app/src/components/ui/ui.css
lrnselfreliance: `flex-shrink: 0` added — but I owe you a retraction on the test.

I could not demonstrate that it matters. Three attempts: ordinary words just wrap and never put the row under pressure; one unbreakable word overflows but the heading's automatic minimum size absorbs the shrinkage; and content given an explicit width still kept it. The help control is also a Mantine ActionIcon with a min-width it will not go below, so the real case was never at risk.

Rather than keep engineering a scenario or leave a green test that does not discriminate, the declaration stays as a statement of intent with a comment saying plainly that it is unproven and why. If you can construct the case I could not, I will happily add the test.

I did not add `min-width: 0` to the text span — with no test showing the overflow is reachable I would be guessing at the remedy too.

app/src/components/Common.js
lrnselfreliance: Real, and fixed by moving `style` to the wrapper rather than by documenting the trap.

Verified the live call site: Status → "Drive Bandwidth" now reports `marginBottom: 16px` on `.wrolpi-header` and `0px` on the `<h2>`. Before, the `1em` sat on the inner heading and only had any effect through margin collapse — which, as you say, breaks the instant the wrapper gains padding, a border or overflow.

Checked all 14 `style` call sites before moving it: eight are layout (margin/marginTop/marginBottom) and are fixed by this; the rest are `textAlign` and `color`, which inherit through the wrapper and are unaffected. The `color` *prop* stays on the heading, since that is about the text and not the box. Two of those `color` styles in Configs.js should really be using the prop — noted for the follow-up sweep.

app/src/components/ui/ui.css
lrnselfreliance: Agreed, and left as a follow-up as you suggested. With `style` now on the wrapper, an inline `marginTop` simply overrides the 22px rather than collapsing against it, so the values are honest — `1.5em` ≈ 24px, `24` exactly — just redundant. Nothing produces double spacing, which is what I checked for before deciding not to touch them here.

They want one decision each (is this heading meant to be further from what precedes it than every other heading?), and that is a different review from a Header fix.

app/src/components/Common.js
lrnselfreliance: Correct, and a real HTML validity bug that I was about to carry forward. `<label>` is rendered only when `for_` names a field now; otherwise the text and asterisk are a plain fragment.

Verified on `/videos/settings`, which is where the collision lives: **zero** `label label` on the page. "Video Resolutions" keeps its label because it has `for_`; "Video File Format", which is passed as `InputForm`'s `label` prop, no longer emits one of its own.

Thank you for chasing it past the three `for_` sites — the ones passing an InfoHeader *as* a label prop are not visible from the `for_` grep I had done.

app/src/components/Common.test.js
lrnselfreliance: You are right, and it read as the opposite of the fix in both files. Renamed to "keeps its help icon on the same row as the text" in `Common.test.js` and "keeps its trailing control on its own row" in the cypress suite. A failure message that describes the bug as the desired behaviour is worse than no message.

app/src/components/ui/Surfaces.tsx
lrnselfreliance: Took the alternative rather than accepting the tradeoff — you had already described the better structure, and it costs nothing.

The control is now a sibling of the heading inside `.wrolpi-header-row`, with the flex row on that wrapper instead of on the `<h*>`. The heading's accessible name is back to `Root CA Certificate`, asserted with `toHaveAccessibleName`, and the wrap bug stays fixed because the row is what holds them together.

The row is only rendered when `after` is present, so the markup for the other 46 call sites is byte-identical to before. Planted its removal: the control wraps again and both suites go red.

PR #613, opened 2026-08-02
fix: a concurrent writer no longer aborts a file move with "database is locked"

lrnselfreliance:
Tagging a Channel with **move files** checked failed the whole move: the Collection's directory was updated, but its files stayed put.

## Root cause

`build_move_plan_bulk` reads FileGroups and then INSERTs the ones missing from the DB, inside a transaction that begins *deferred*. SQLite refuses a mid-transaction lock upgrade **immediately** — `busy_timeout` is deliberately skipped for upgrades to avoid deadlock — so any concurrent writer (a refresh, a download, a config save) aborted planning outright:

```
File "/opt/wrolpi/wrolpi/files/worker.py", line 689, in build_move_plan_bulk
    session.flush()
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) database is locked
[SQL: INSERT INTO file_group (...)]
```

`_execute_move_chunks` has the same shape and fails the same way one step later, on its `INSERT INTO directory`.

## Fix

Both transactions now begin as writers (`get_immediate_db_session`), so `busy_timeout` (30s) absorbs the contention instead of erroring instantly.

That requires the writes nested inside the move to share its transaction rather than open a second connection, which would deadlock against the write lock the move now holds up front:

- `_bulk_update_file_groups_db` runs on the caller's session instead of `get_db_curs`.
- The old-directory cleanup deletes its `Directory` record through that session instead of calling `delete_directory` (same non-recursive semantics: `rmdir`, then drop the record).

Side effect: a failed move no longer leaves earlier chunks' path updates committed after their files have been moved back — the rollback now undoes disk and DB together.

## Tests

Two regression tests in `wrolpi/files/test/test_worker_move.py`, each written first and each reproducing the production failure before the fix:

- `test_build_move_plan_waits_for_concurrent_writer` — an unindexed file, so planning must INSERT. Failed with `database is locked` at `worker.py:689`, matching the reported traceback.
- `test_file_worker_move_survives_concurrent_writer` — all files indexed, so the first write is in the execute phase. Failed with `INSERT INTO directory ... database is locked`.

Both hold the SQLite write lock from a second connection for one second, then release it; a correctly-written move waits and succeeds.

Full backend suite: 1711 passed, 5 skipped. No frontend changes.

## Known, not addressed here

In `Collection.move` the new directory is committed *before* the files move, and `handle_move` swallows its exception without marking the job, so `wait_for_job` spins the full 300s timeout before the revert runs. That window is why the channel appeared moved with its files left behind. Worth a separate change.

### DIFF
--- a/wrolpi/files/lib.py
+++ b/wrolpi/files/lib.py
@@ -18,7 +18,7 @@ from pathlib import Path
 from typing import Callable, List, Tuple, Union, Dict, Generator, Iterable, Set
 
 import cachetools.func
-from sqlalchemy import asc, or_
+from sqlalchemy import asc, or_, text as sa_text
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session
 
@@ -1594,23 +1594,25 @@ def _bulk_update_file_groups_db(session: Session, chunk_plan: Dict[pathlib.Path,
     if not chunk_plan:
         return
 
-    with get_db_curs(commit=True) as curs:
-        # Build the update rows: [old_path, new_dir, new_path]
-        updates = json.dumps([
-            [str(old_path), str(new_path.parent), str(new_path)]
-            for old_path, new_path in chunk_plan.items()
-        ])
+    # Build the update rows: [old_path, new_dir, new_path]
+    updates = json.dumps([
+        [str(old_path), str(new_path.parent), str(new_path)]
+        for old_path, new_path in chunk_plan.items()
+    ])
 
-        # Use UPDATE FROM json_each pattern for efficient batch update
-        sql = '''
-            UPDATE file_group AS fg SET
-                directory = json_extract(v.value, '$[1]'),
-                primary_path = json_extract(v.value, '$[2]'),
-                indexed = CASE WHEN fg.data IS NULL THEN FALSE ELSE fg.indexed END
-            FROM json_each(?) AS v
-            WHERE fg.primary_path = json_extract(v.value, '$[0]')
-        '''
-        curs.execute(sql, (updates,))
+    # Use UPDATE FROM json_each pattern for efficient batch update.  This runs on the caller's
+    # session (rather than a second connection) so it shares the move's transaction: a second
+    # connection would deadlock against the write lock the caller already holds, and its separate
+    # commit would survive a rollback of the rest of the move.
+    sql = '''
+        UPDATE file_group AS fg SET
+            directory = json_extract(v.value, '$[1]'),
+            primary_path = json_extract(v.value, '$[2]'),
+            indexed = CASE WHEN fg.data IS NULL THEN FALSE ELSE fg.indexed END
+        FROM json_each(:updates) AS v
+        WHERE fg.primary_path = json_extract(v.value, '$[0]')
+    '''
+    session.execute(sa_text(sql), dict(updates=updates))
 
 
 def _json_serial(obj):
--- a/wrolpi/files/test/test_worker_move.py
+++ b/wrolpi/files/test/test_worker_move.py
@@ -7,6 +7,7 @@ import time
 import pytest
 
 from wrolpi.conftest import production_like_sessions
+from wrolpi.files import worker as worker_module
 from wrolpi.files.lib import _move_file_group_files
 from wrolpi.files.models import FileGroup, Directory
 from wrolpi.files.worker import file_worker, FileTask, FileTaskType, build_move_plan_bulk
@@ -592,13 +593,14 @@ async def test_build_move_plan_waits_for_concurrent_writer(
 
 @pytest.mark.asyncio
 async def test_file_worker_move_survives_concurrent_writer(
-        async_client, test_session, test_directory, make_files_structure
+        async_client, test_session, test_directory, make_files_structure, monkeypatch
 ):
-    """A move of already-indexed files must survive a concurrent writer holding the write lock.
+    """Executing a move must survive a concurrent writer holding the write lock.
 
-    Planning writes nothing here (every file is already indexed), so the first write is the
-    `FileGroup` UPDATE inside `_execute_move_chunks`.  That transaction must begin as a writer,
-    or SQLite refuses the mid-transaction lock upgrade instantly and the move aborts.
+    `_execute_move_chunks` reads FileGroups/Directories and then writes them, so its transaction
+    must begin as a writer too - SQLite refuses the mid-transaction lock upgrade instantly and the
+    move aborts.  The lock is taken only *after* planning finishes, so this fails if the execute
+    site alone regresses; planning cannot absorb the contention on the execute site's behalf.
     """
     file1, = make_files_structure(['source/file1.txt'])
     FileGroup.from_paths(test_session, file1)
@@ -612,9 +614,20 @@ async def test_file_worker_move_survives_concurrent_writer(
     task = FileTask(FileTaskType.move, [file1], destination=dest)
     file_worker.private_queue.put_nowait(task)
 
-    with production_like_sessions(test_session) as maker:
-        with _write_lock_held_briefly(db_file):
-            await file_worker.process_queue()
+    with contextlib.ExitStack() as stack:
+        maker = stack.enter_context(production_like_sessions(test_session))
+
+        real_build_move_plan_bulk = worker_module.build_move_plan_bulk
+
+        async def build_plan_then_hold_write_lock(*args, **kwargs):
+            plan = await real_build_move_plan_bulk(*args, **kwargs)
+            # Planning is done; the contention now falls entirely on the execute phase.
+            stack.enter_context(_write_lock_held_briefly(db_file))
+            return plan
+
+        monkeypatch.setattr(worker_module, 'build_move_plan_bulk', build_plan_then_hold_write_lock)
+
+        await file_worker.process_queue()
 
         assert (dest / 'file1.txt').is_file(), 'file was not moved'
         assert not file1.exists(), 'file was left at its old path'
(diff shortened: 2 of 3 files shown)


wrolpi/files/worker.py
```diff
@@ -556,7 +556,11 @@ async def build_move_plan_bulk(
         if progress_callback and i % 100 == 0:
             progress_callback(i, total_sources * 2)  # *2 because file collection is second half
 
-    with get_db_session() as session:
+    # Planning reads FileGroups and then INSERTs the ones missing from the DB, so the transaction
+    # must begin as a writer.  A deferred transaction upgrading to a writer mid-way gets "database
+    # is locked" *immediately* (busy_timeout is skipped for lock upgrades), so any concurrent
+    # writer - a refresh, a download, a config save - would abort the whole move.
+    with get_immediate_db_session() as session:
```
lrnselfreliance, line 603: **[suggestion]** `build_move_plan_bulk` now opens `get_immediate_db_session()` before filesystem collection (`glob_shared_stem` / `walk`) that still runs inside the same `with` block. `BEGIN IMMEDIATE` therefore holds the main-DB write lock for the entire planning window, including pure FS work and TEMP-table setup. Under the old deferred session, already-indexed plans typically never needed a main write lock (TEMP + reads only); unindexed plans only upgraded when inserting FileGroups. Holding the write lock for large tree walks will make concurrent writers (refresh, download, config save) block and then fail after `busy_timeout` (30s), trading an instant move abort for longer global write starvation.

**Suggestion:** Collect `all_file_paths` / `dir_file_mapping` / deleted sources **outside** the DB session, then open a short `get_immediate_db_session()` only for TEMP load, FileGroup reads/inserts, and commit. Keep execute IMMEDIATE, but consider committing per move chunk (or at least not holding one IMMEDIATE across all chunks + `shutil.move` + `asyncio.sleep(0)`) if long multi-file moves are common on Pi hardware.

wrolpi/files/test/test_worker_move.py
```diff
@@ -530,3 +536,89 @@ async def test_file_worker_move_cleans_directory_records(
...
+
+    with production_like_sessions(test_session):
+        with _write_lock_held_briefly(db_file):
+            plan, _ = await build_move_plan_bulk([file1], dest)
+
+    assert plan == {file1: dest / 'file1.txt'}
+
+
+@pytest.mark.asyncio
+async def test_file_worker_move_survives_concurrent_writer(
```
lrnselfreliance, line 595: **[suggestion]** `test_file_worker_move_survives_concurrent_writer` intends to prove execute-time IMMEDIATE is required for already-indexed moves, but with this PR both planning and execute use IMMEDIATE under the same 1s lock window. Planning runs first, waits out `_write_lock_held_briefly`, and only then does execute start — usually with no remaining contention. The test still fails if **neither** site is fixed, but it would stay green if someone reverted only the execute `get_immediate_db_session()` change while leaving planning IMMEDIATE in place. That weakens repro fidelity for the execute path called out in the docstring.

**Suggestion:** Hold the write lock only around execute (e.g. acquire after a patched/no-op plan, or start the holder from a monkeypatch on `_execute_move_chunks` / `get_immediate_db_session`), or split into two tests: one that only calls `build_move_plan_bulk` (already present) and one that builds the plan first, then holds the lock solely during `process_queue` execute / `_execute_move_chunks`.

wrolpi/files/worker.py
```diff
@@ -2037,9 +2043,11 @@ def on_planning_progress(processed: int, total: int):
                 operation_total=len(plan),
             )
 
-            # Execute the plan in chunks
+            # Execute the plan in chunks.  This reads FileGroups/Directories then updates them, so
+            # it must begin as a writer; a deferred transaction's lock upgrade fails instantly when
+            # anything else is writing (see `build_move_plan_bulk`).
             with flags.file_worker_discovery:
-                with get_db_session(commit=True) as session:
+                with get_immediate_db_session() as session:
```
lrnselfreliance, line 2048: **[suggestion]** On failure, `handle_move` still reverts files and emits an event but never marks the job failed/complete (`_set_job_status(..., 'failed')` / `_complete_job`). Callers such as `Collection.move` that `await wait_for_job(job_id)` can still sit until the 300s timeout. This is pre-existing and outside the core lock fix, but it is adjacent to the changed session line and remains the user-visible outcome when contention lasts longer than `busy_timeout` (move waits up to 30s, then fails, then the waiter can hang another 300s).

**Suggestion:** In the `except Exception` path (and the invalid-task early return), set the job to `failed` (or complete with error) so `wait_for_job` unblocks; add a small test that a failed move under `production_like_sessions` leaves a terminal job status.

wrolpi/files/lib.py
```diff
@@ -1594,23 +1594,25 @@ def _bulk_update_file_groups_db(session: Session, chunk_plan: Dict[pathlib.Path,
...
+    # commit would survive a rollback of the rest of the move.
+    sql = '''
+        UPDATE file_group AS fg SET
+            directory = json_extract(v.value, '$[1]'),
+            primary_path = json_extract(v.value, '$[2]'),
+            indexed = CASE WHEN fg.data IS NULL THEN FALSE ELSE fg.indexed END
+        FROM json_each(:updates) AS v
+        WHERE fg.primary_path = json_extract(v.value, '$[0]')
+    '''
+    session.execute(sa_text(sql), dict(updates=updates))
```
lrnselfreliance, line 1615: **[nit]** Switching from sqlite3 `?` + `get_db_curs` to SQLAlchemy `text()` + `:updates` is correct for SA 1.3 and for sharing the caller's transaction. No functional issue found with named binds vs `json_each` (JSON is a bound value, not interpolated). Minor consistency note: `_bulk_update_file_groups_reorganize` still uses positional `?` via `get_db_curs`, and this helper no longer expires the identity map after raw UPDATEs (pre-existing for the move path; reorganize explicitly `expire_all`s). Same-session raw UPDATEs make stale `FileGroup.primary_path` on identity-map instances slightly more relevant if a later chunk ever re-loads a moved group by new path.

**Suggestion:** Optionally `session.expire_all()` (or expire just the updated FileGroups) after the bulk UPDATE for symmetry with reorganize; not required for the current one-pass plan keys.

lrnselfreliance (commented):
## Summary

This PR correctly fixes the SQLite deferred-upgrade `database is locked` failure on moves by using `get_immediate_db_session()` for plan and execute, and by running bulk FileGroup updates and directory-record cleanup on the caller's session instead of nested connections. Nested-write sharing is the right call: a second connection under an open `BEGIN IMMEDIATE` would deadlock/wait out `busy_timeout`, and the old separate `get_db_curs(commit=True)` commit could survive an outer rollback. Regression tests target the real lock-upgrade failure mode with `production_like_sessions`; remaining concerns are lock-hold duration during FS work and incomplete isolation of the execute-path regression test.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 1

wrolpi/files/worker.py
lrnselfreliance: Correct, and fixed in cfc5464e. `BEGIN IMMEDIATE` fires on the session's first statement (the `CREATE TEMP TABLE`), so the write lock really was held across every `glob_shared_stem` and `walk` — on a large channel that is minutes of global write starvation, which is worse than what this PR set out to fix.

The collection now completes before the session opens; only the temp-table load, the FileGroup reads/inserts, and the commit run under the lock. `file_source_set` is built during that same pass rather than re-globbing every file source a second time further down.

On per-chunk commits in execute: leaving it as one transaction. Sources and destination are both validated to be under the media directory, so `shutil.move` is a same-filesystem rename — the execute phase is dominated by cheap renames, not copies, so the lock window is short even for a few thousand files. Committing per chunk would also reinstate the split this PR removed, where a mid-move failure leaves earlier chunks' path updates committed after `_revert_move` has already moved those files back.

wrolpi/files/test/test_worker_move.py
lrnselfreliance: Right — the test proved "at least one of the two sites is fixed", not the execute site specifically. Fixed in cfc5464e: `build_move_plan_bulk` is wrapped so the lock holder starts only *after* planning returns, putting the contention entirely on the execute phase.

Verified by mutation, each site now failing its own test and only its own:

- revert the execute `get_immediate_db_session()` → `test_file_worker_move_survives_concurrent_writer` fails on the shared-session `UPDATE file_group`; the other 18 pass.
- revert the planning one → `test_build_move_plan_waits_for_concurrent_writer` fails; the other 18 pass.

wrolpi/files/worker.py
lrnselfreliance: Agreed on the diagnosis, deliberately out of scope here. This is the user-visible half of the bug report (the Channel's directory updated, files left behind) but it is a separate defect with separate semantics to settle: `wait_for_job` currently only recognizes `complete`, so adding a `failed` terminal state means deciding whether it raises, and `Collection.move`'s revert path keys off that. It deserves its own tests rather than riding along on the lock fix. Filing it as a follow-up.

wrolpi/files/lib.py
lrnselfreliance: Skipping the `expire_all()`. Within this session nothing re-reads the updated rows: plan keys are unique per FileGroup, so each group appears in exactly one chunk and is never re-queried by its new path; post-processing runs after the session closes, on fresh sessions. An `expire_all()` would only force needless reloads of the Directory rows the loop is still using.

The `?` vs `:name` inconsistency with `_bulk_update_file_groups_reorganize` is real but it is on the other side of the fence — reorganize keeps its own connection and its own commit, which is what forced the named binds here. Worth unifying when reorganize gets the same treatment.

PR #614, opened 2026-08-02
ui: progress bars get room for their label, and keep their band inside

lrnselfreliance:
## Too thin

16px tall with an 11px bold label left about half a pixel of air above and below — and the Status page draws **44 of them**. Now 22px with a 12px label, since these carry device names and units rather than just a number.

## The band escaped the bar

`indeterminate` slides on `margin-left` from `-35%` to `100%` and nothing clipped it, so the animation ran from outside the left edge to outside the right one — exactly as you saw. `overflow: hidden` on the bar.

## Why CPU/RAM touched but bandwidth didn't

It isn't the bars. **CPU and RAM are bare siblings inside a `Panel`** (`{cpuProgress}{memoryUsageProgress}`), and `.wrolpi-progress` had no margin. **Each bandwidth bar sits alone in a `<Grid.Col>`**, so the Mantine `Grid` supplies the gutter.

`* + .wrolpi-progress { margin-top: 8px }` separates those two cases on its own — it only fires when a bar actually follows something, so an only-child bar in a column is untouched. Verified on the live page: 8px between CPU and RAM, **0px on every bandwidth bar**. A blanket margin would have doubled the Grid's gutter, and that's asserted too.

## Gallery

New Progress section: six percentages (including the mid-range ones where the label crosses the fill boundary), custom labels, no label, indeterminate, and the four roles. I removed the demo's own gaps so what it shows is the spacing the component really has.

## Measured, and NOT fixed — your call

Building the gallery turned up something worth your eye. The label spans the whole bar, so it sits on the fill **and** the track:

| theme | over fill | over track |
|---|---|---|
| light | **5.49** ✓ | 14.95 |
| dark | **2.35** ✗ | 13.45 |
| night | **2.48** ✗ | 5.18 |
| amber | **2.74** ✗ | 9.16 |

Light clears it because of the `brightness(1.75)` filter; the other three don't. Reaching AA over the fill needs the label drawn twice and clipped at the fill boundary so each half takes its own colour — and **in night not even that works**: `--text` is only ~4:1 against pure black, and a fill is lighter than black by definition.

I've recorded it in `ui.css` and in the test, which asserts the **track half only** rather than picking a threshold low enough to pass. Say the word if you want the two-copy clip for dark and amber.

## One test was unfalsifiable

"Keeps a long label from widening the bar" **cannot fail** — the label is `position: absolute; inset: 0` and takes no part in its parent's width. What a long label really does is wrap to a second line inside a 22px bar and get cut in half by the new clip. It measures line boxes now.

## Verified

- jest **1074 passing** / 59 suites
- `ui-layout.cy.js` **133 passing** (was 124)
- `npm run build` clean
- five plants, all red: 16px height, clip removed, blanket margin, no `nowrap`, no adjacent margin
- checked all four themes in the gallery and on `/admin/status`

### DIFF
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -521,17 +521,14 @@ describe('a progress bar has room for the label inside it', () => {
 
         cy.get('.wrolpi-progress').should(($bar) => {
             const box = $bar[0].getBoundingClientRect();
-            const text = $bar[0].querySelector('.wrolpi-progress-text')
-                .firstChild.getBoundingClientRect
-                ? $bar[0].querySelector('.wrolpi-progress-text').getBoundingClientRect()
-                : null;
             expect(box.height, 'bar height').to.be.at.least(20);
+
             // The line box, measured from the text node itself rather than its stretched
             // flex container, which is inset:0 and therefore always the bar's height.
             const range = document.createRange();
             range.selectNodeContents($bar[0].querySelector('.wrolpi-progress-text'));
-            const line = range.getBoundingClientRect();
-            expect(box.height - line.height, 'air around the label').to.be.at.least(4);
+            expect(box.height - range.getBoundingClientRect().height, 'air around the label')
+                .to.be.at.least(4);
         });
     });
 
@@ -551,10 +548,21 @@ describe('a progress bar has room for the label inside it', () => {
         );
 
         cy.get('.wrolpi-progress-text').should(($text) => {
+            const style = getComputedStyle($text[0]);
+            expect(style.whiteSpace, 'label does not wrap').to.equal('nowrap');
+            expect(style.textOverflow, 'and is cut with an ellipsis, not mid-glyph')
+                .to.equal('ellipsis');
+
             const range = document.createRange();
             range.selectNodeContents($text[0]);
-            const lines = range.getClientRects().length;
-            expect(lines, 'the label occupies one line box').to.equal(1);
+            expect(range.getClientRects().length, 'one line box').to.equal(1);
+            /*
+             * And that one line is genuinely wider than the space it has -- otherwise the
+             * test proves nothing, since a label that fits occupies one line either way.
+             * This is the case that used to wrap and get bisected by the clip.
+             */
+            expect(range.getBoundingClientRect().width, 'the label really is too long')
+                .to.be.greaterThan($text[0].getBoundingClientRect().width);
         });
     });
 
@@ -619,11 +627,15 @@ describe('a progress bar has room for the label inside it', () => {
 
     it('keeps the indeterminate band inside the bar', () => {
         /*
-         * The band slides on `margin-left`, from -35% to 100%.  Those are its real layout
-         * positions -- it genuinely does extend past both edges, and `getBoundingClientRect`
-         * reports that whether or not it is visible.  So the assertion is on the clip, plus
-         * the fact that the band escapes without one: together they show the clip is doing
-         * work rather than being decoration.
+         * The band slides on `margin-left` from -35% to 100%, so it genuinely does extend
+         * past both edges and `getBoundingClientRect` reports that whether or not it is
+         * painted.  The clip is what keeps it out of sight.
+         *
+         * The band is PINNED here rather than sampled mid-animation.  Reading a running
+         * animation made this both flaky and, worse, permanently false for anyone with
+         * Reduce Motion turned on: `prefers-reduced-motion` replaces the slide with
+         * `width: 100%` and no negative margin, so the band never leaves the box and an
+         * assertion that it does could not pass on that machine at all.
          */
         cy.mountUI(bar({indeterminate: true, label: 'Uploading…'}), {theme: 'light'});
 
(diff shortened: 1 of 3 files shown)


app/src/components/ui/ui-layout.cy.js
```diff
@@ -509,6 +509,136 @@ describe('a heading opens a section rather than closing the last one', () => {
...
+         */
+        cy.mountUI(bar({indeterminate: true, label: 'Uploading…'}), {theme: 'light'});
+
+        cy.get('.wrolpi-progress-indeterminate').should(($bar) => {
+            expect(getComputedStyle($bar[0]).overflow, 'the bar clips its contents')
+                .to.equal('hidden');
+
+            const box = $bar[0].getBoundingClientRect();
+            const band = $bar[0].querySelector('.wrolpi-progress-fill').getBoundingClientRect();
+            expect(band.left < box.left - 1 || band.right > box.right + 1,
```
lrnselfreliance: **[bug]** `keeps the indeterminate band inside the bar` asserts that the fill's layout box currently extends past the track (`band.left < box.left - 1 || band.right > box.right + 1`). That is only true while `wrolpi-progress-slide` is running. Under `prefers-reduced-motion: reduce`, `ui.css` disables the animation and forces the fill to `width: 100%` with no negative margin, so the band never leaves the box and the expectation is permanently false. Anyone (or any CI host) with Reduce Motion enabled will see a hard fail; the assertion is also time-dependent for everyone else (Cypress retries usually save it, but it is still sampling a moving target).

**Suggestion:** Stop relying on a live animation frame. Pin the fill to a known out-of-band position, then assert clip + escape, e.g. set `animation: none; margin-left: -35%; width: 35%` on `.wrolpi-progress-fill` inside the `should` callback (or freeze via the Web Animations API), then check `overflow === 'hidden'` and that the band's rect crosses an edge. That passes with reduced motion and does not depend on the clock.

app/src/components/ui/ui.css
```diff
@@ -1029,15 +1052,29 @@ html .mantine-Notification-closeButton:hover {
...
     justify-content: center;
-    font-size: 11px;
+    /* 12px rather than 11: these labels carry units and device names, not just a number. */
+    font-size: 12px;
     font-weight: 700;
     color: var(--text);
     font-variant-numeric: tabular-nums;
+    /* A long label must not push the bar wider than the column it was given. */
+    padding: 0 6px;
+    overflow: hidden;
```
lrnselfreliance, line 1068: **[suggestion]** Long labels get `overflow: hidden` + `white-space: nowrap` with no `text-overflow: ellipsis`. That fixes the wrap-and-bisect problem the test describes, but the result is a hard mid-glyph cut. Real call sites feed long strings into this surface (upload filenames in `Upload.js` / `FileBrowser.js`, drive mount paths in `Status.js` DriveInfo, refresh phase labels in `Files.js`). A clipped `Complete: very-long-archive-name.tar.gz` reads as broken rather than truncated.

**Suggestion:** Add `text-overflow: ellipsis` next to the existing overflow/nowrap rules on `.wrolpi-progress-text`. Optionally assert `getComputedStyle(text).textOverflow === 'ellipsis'` (or that the range width exceeds the bar and the used value is ellipsis) in the long-label spec.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -509,6 +509,136 @@ describe('a heading opens a section rather than closing the last one', () => {
...
+    it('leaves air above and below the text', () => {
+        /*
+         * The Status page draws 44 of these.  At 16px tall the 11px bold label left about
+         * half a pixel of air, which is legible only if you already know what it says.
+         */
+        cy.mountUI(bar({percent: 62, label: 'CPU Usage (16 cores)'}), {theme: 'light'});
+
+        cy.get('.wrolpi-progress').should(($bar) => {
+            const box = $bar[0].getBoundingClientRect();
+            const text = $bar[0].querySelector('.wrolpi-progress-text')
```
lrnselfreliance: **[nit]** The air-around-label test builds a `text` rect from a ternary on `firstChild.getBoundingClientRect` and never uses it. The real measurement is the Range API block below. The dead branch reads like a leftover from an earlier draft and confuses what is actually being asserted.

**Suggestion:** Delete the unused `text` rect lines and keep only the Range-based line-box measurement.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -509,6 +509,136 @@ describe('a heading opens a section rather than closing the last one', () => {
...
+         * the clip, so that is what is measured.
+         */
+        cy.mountUI(
+            <div style={{width: 200}}>
+                <Progress percent={40} label='nbd0 read 128.4 MBps sustained over 5 minutes'/>
+            </div>,
+            {theme: 'light'},
+        );
+
+        cy.get('.wrolpi-progress-text').should(($text) => {
```
lrnselfreliance, line 550: **[suggestion]** `keeps a long label on one line` only checks `range.getClientRects().length === 1`. With `white-space: nowrap` present that is nearly tautological if the rule applied, and it does not lock in the clip that the comment claims is the real failure mode (wrap inside a 22px bar, second line bisected).

**Suggestion:** Also assert `getComputedStyle($text[0]).whiteSpace === 'nowrap'` and `overflow === 'hidden'`, and/or that the text range width is greater than the bar's content width (proving the label would wrap without nowrap and is being clipped rather than expanding the bar).

lrnselfreliance (commented):
## Summary

Solid, well-motivated layout PR: raising the bar to 22px, clipping the indeterminate band with `overflow: hidden`, and spacing stacked bars with the house-style `* +` selector all match the stated Status-page problems. The Cypress specs measure real geometry (Range line boxes, getBoundingClientRect gaps, contrast against the track) rather than class-name theater, and ThemeSample now exercises the hard mid-fill cases. A few issues remain around test robustness under reduced motion, long-label truncation polish, and a bit of dead code in the new specs.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 1

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Correct on both counts, and I reproduced it rather than taking it on faith — forcing the reduced-motion branch for the whole run fails the old form and passes the new one:

```
OLD sampling form + reduced motion: 1 failing
    1) keeps the indeterminate band inside the bar
NEW pinned form + reduced motion:   134 passing
```

So this was not a latent risk, it was a test that could not pass on a machine with Reduce Motion set — and would have looked like a real regression to whoever hit it.

Pinned to where the animation's first keyframe puts it (`animation: none; width: 35%; margin-left: -35%`), then asserting the clip plus that escape. I also added the branch you implied: with motion suppressed the bar fills instead of sliding, so a band that never moves must never leave the box. That half is deterministic too, and together they cover both code paths in the CSS.

app/src/components/ui/ui.css
lrnselfreliance: Agreed, and the call sites are real — I checked before adding it. `Upload.js:59` and `FileBrowser.js:903` both label with `${statusString} ${name}`, i.e. a filename, and `Files.js:1038` labels with the raw error string. A hard cut through `very-long-archive-name.tar.gz` reads as a rendering bug rather than truncation.

`text-overflow: ellipsis` added, and the long-label spec asserts the computed value. Removing it now goes red.

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Right — leftover from the draft where I measured the flex container before realising it is `inset: 0` and therefore always the bar's height. Deleted; only the Range measurement remains, which is the one that was doing the work.

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Fair, and it was weaker than that even: with `nowrap` applied, "one line box" is close to a restatement of the rule, and a label that simply *fits* would satisfy it too.

It now asserts the computed `whiteSpace` and `textOverflow`, and — the part that gives it teeth — that the line is genuinely **wider** than the space it has, so the fixture is provably the wrapping case and not a short label that would pass either way. Both `nowrap` and `ellipsis` go red when removed.

PR #616, opened 2026-08-02
ui: cards get legible titles, contained posters, and meta beside the title

lrnselfreliance:
Four defects reported on `/videos`, `/docs` and `/archives`.

### Titles were invisible outside light

`.card-link` carried a literal `color: black` from before the themes existed.

| theme | title | panel | contrast |
|---|---|---|---|
| light | `#000000` | `#ffffff` | 21:1 |
| dark | `#000000` | `#26292b` | **1.36:1** |
| night | `#000000` | `#160404` | **1.06:1** |
| amber | `#000000` | `#1a0f00` | **1.09:1** |

It covers every card title and the channel/author/domain line beside it. The date was fine, because it takes `--muted` from the card body — which is exactly what made this read as a card-title bug rather than a link bug. Titles now take `--text`: they are links, but they are the card's *text*, not a wall of blue.

### Posters were cropped, not mis-proportioned

The image carried a fixed `290x163` box. A card in the default grid is ~233px wide and Mantine's `Card` sets `overflow: hidden`, so the browser painted the middle 233px of a 290px image — the ratio was right in the layout tree and wrong on screen.

Replaced with caps on **both** axes (the card's width, and `--card-poster-max-height: 200px`) with `width`/`height` left `auto`, so the pair behaves as "contain". Bounding the width alone was the first attempt and is not enough — a square thumbnail then stands as tall as the card is wide, and a portrait cover taller still:

| source | before | width cap only | both caps |
|---|---|---|---|
| 1280×720 | 290×163, cropped to 233 | 231×130 | 231×130 |
| 600×600 | 163×163, cropped | 231×231 | 200×200 |
| 306×396 | 126×163 | 231×299 | 155×200 |
| 120×90 | 120×90 | 120×90 | 120×90 |

### The meta line was foot-anchored

`margin-top: auto` on the meta meant that on `/docs` a PDF with no author had its date shoved up to **55px** below the title, at the bottom of a card that was tall only because of its neighbour in the row. It now sits directly under the title, 4px down.

### Archive cards read buttons-then-domain

Because the buttons were `children` and the meta was what got pushed to the foot. `Card` gains an `actions` slot, rendered last and anchored to the foot, so a row lines its buttons up however many lines each title took.

### Gallery

A poster row covering all five ratios, and a meta/actions row with deliberately unequal titles.

### Tests

1076 jest / 59 suites, 145 in `ui-layout.cy.js`, clean `npm run build`. The 12 failures elsewhere in the cypress component run are the documented pre-existing Semantic-era ones.

Every new guard was run against a planted regression, in **every** suite that claims the behaviour, with the plant kept as narrow as the claim:

| plant | red | stays green (inverse) |
|---|---|---|
| drop `max-height` | square, portrait, squash | 16:9 and small poster — not height-bound |
| `width: 100%` | containment + no-upscale | 16:9 — still 233 wide either way |
| restore `color: black` | dark, night, amber | light — black on white *is* legible |
| re-anchor the meta | cypress gap + jest order | — |
| swap slot order | cypress gap + foot, jest order | — |

The poster cap is read from the stylesheet rather than repeated in the spec, so the two cannot drift; deleting the declaration makes it `NaN` and every comparison fails. A `within(80, 240)` bound keeps "read whatever the CSS says" from being satisfied by any value at all.

### DIFF
--- a/app/src/components/Archive.js
+++ b/app/src/components/Archive.js
@@ -426,22 +426,22 @@ export function ArchiveCard({file}) {
         </div>
     </>;
 
-    return <Card media={media} title={titleElm} meta={meta}
-                 color={mimetypeColor(file.mimetype, file.primary_path)}>
-        <div style={{display: 'flex', gap: '0.5em', marginTop: '0.5em'}}>
-            <Link to={`/archives/${file.id}`}>
-                <Button icon='file alternate'>Details</Button>
-            </Link>
-            <IconButton
-                icon='external'
-                label='Open original URL'
-                component='a'
-                href={file.url}
-                target='_blank'
-                rel='noopener noreferrer'
-            />
-        </div>
-    </Card>
+    const actions = <div style={{display: 'flex', gap: '0.5em'}}>
+        <Link to={`/archives/${file.id}`}>
+            <Button icon='file alternate'>Details</Button>
+        </Link>
+        <IconButton
+            icon='external'
+            label='Open original URL'
+            component='a'
+            href={file.url}
+            target='_blank'
+            rel='noopener noreferrer'
+        />
+    </div>;
+
+    return <Card media={media} title={titleElm} meta={meta} actions={actions}
+                 color={mimetypeColor(file.mimetype, file.primary_path)}/>
 }
 
 // Domain table column configuration
--- a/app/src/components/Common.js
+++ b/app/src/components/Common.js
@@ -174,19 +174,30 @@ export function isoDatetimeToString(dt, time = false) {
     return d;
 }
 
-export function CardLink({to, newTab = false, ...props}) {
+/*
+ * Both card links COMPOSE the caller's className rather than being overwritten by it.
+ * Written as `className='...' {...props}`, the spread put the caller's class last and it
+ * replaced the pair outright -- so every archive title, which passes
+ * `card-title-ellipsis`, lost `card-link` and rendered in the link colour instead of the
+ * card's text colour.  The one card type whose classes are set this way, and the one whose
+ * title looked wrong.
+ */
+const cardLinkClass = (className) =>
+    ['no-link-underscore', 'card-link', className].filter(Boolean).join(' ');
+
+export function CardLink({to, newTab = false, className, ...props}) {
     props = newTab === true ? {...props, target: '_blank', rel: 'noopener noreferrer'} : props;
-    return <Link to={to} className="no-link-underscore card-link" {...props}>
+    return <Link to={to} className={cardLinkClass(className)} {...props}>
         {props.children}
     </Link>
 }
 
-export function ExternalCardLink({to, children, ...props}) {
+export function ExternalCardLink({to, children, className, ...props}) {
     return <a
         href={to}
         target='_blank'
         rel='noopener noreferrer'
-        className='no-link-underscore card-link'
+        className={cardLinkClass(className)}
         {...props}
     >
         {children}
@@ -726,7 +737,7 @@ export function findPosterPath(file) {
     }
 }
 
-export function CardPoster({to, file}) {
+export function CardPoster({to, file, overlay}) {
     const navigate = useNavigate();
 
     // Marks a tagged file, pinned to the poster's corner.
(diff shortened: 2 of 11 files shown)


app/src/components/Common.js
```diff
@@ -741,25 +739,17 @@ export function CardPoster({to, file}) {
...
-            </Link>
-        } else {
-            // Preview the file.
-            return <div style={style}>
-                <PreviewLink file={file}>
-                    {image}
-                </PreviewLink>
-            </div>
-        }
+        return <div className='wrolpi-card-poster'>
```
lrnselfreliance, line 770: **[suggestion]** The tag mark is absolutely positioned on `.wrolpi-card-poster` (full card width), while the image is centered with `margin: 0 auto` and may be narrower than the band when height-capped (portrait covers, small posters, squares). The badge then sits in empty letterbox space rather than on the poster’s corner — the “pinned to the poster’s corner” convention the comments describe. Video duration overlays wrapped around `CardPoster` have the same geometry when the image does not fill the band.

**Suggestion:** Shrink-wrap a relatively positioned frame around the image (and tag), then center that frame in `.wrolpi-card-poster`, so absolute corner chrome tracks the painted poster rather than the full media band.

app/src/components/Archive.js
```diff
@@ -426,22 +426,22 @@ export function ArchiveCard({file}) {
         </div>
```
lrnselfreliance, line 426: **[suggestion]** The PR makes `.card-link` use `var(--text)` so title/meta links read as body text, not a wall of blue. Archive titles go through `ExternalCardLink` with `className='card-title-ellipsis'`, and that component assigns `className` before spreading props, so the caller’s className replaces `no-link-underscore card-link`. Archive titles therefore stay on the default link colour and never hit the new rule (or the underscore suppression). Contrast is still fine via `--blue`, but this is the one card type this PR migrates for layout and it misses the title treatment.

**Suggestion:** Pass a merged class list at the call site (`no-link-underscore card-link card-title-ellipsis`), or fix `ExternalCardLink` / `CardLink` to compose `className` instead of overwriting.

app/src/components/ui/ui.css
```diff
@@ -102,6 +102,28 @@ html[data-theme="night"] .wrolpi-button-danger:hover:not(:disabled) {
...
+ * other toolbar that mixes the two.  It already ships the aligned scale as
+ * `--ai-size-input-*`, for an ActionIcon standing beside an Input; those five values are
+ * identical to `--button-height-*`.  Pointing one at the other fixes every size at once,
+ * including ones no call site uses yet.
+ *
+ * The `html` prefix is load-bearing.  Mantine declares this scale on ActionIcon's own
+ * hashed class, which has the same specificity as `.mantine-ActionIcon-root` and would win
+ * on source order.
+ */
+html .mantine-ActionIcon-root {
```
lrnselfreliance, line 119: **[suggestion]** Remapping every `.mantine-ActionIcon-root` size token to `--ai-size-input-*`, plus defaulting `IconButton` to `sm`, correctly pairs Details/Open on cards and is well tested. It is also global: every `IconButton size='xs'` (dense tables, flasher rows, inventory schema editors, etc.) and default-sized controls (e.g. SearchBox clear) grow to button/input heights. That may be desired, but it is a wider visual change than the card toolbar that motivated it.

**Suggestion:** Spot-check dense `xs` toolbars and the search clear control after the remapping; if some need to stay compact, give those call sites an explicit compact size or scope the CSS more tightly than all ActionIcons.

app/src/components/Common.js
```diff
@@ -741,25 +739,17 @@ export function CardPoster({to, file}) {
...
+            {imageLabel}
+            {to
```
lrnselfreliance, line 771: **[nit]** The tag badge used to render inside the `Link` / `PreviewLink` alongside the image, so the tagged corner was part of the poster hit target. It is now a sibling before the link, so clicks on the badge no longer navigate/preview.

**Suggestion:** Keep the badge inside the same interactive wrapper as the image (or put `pointer-events: none` on the badge if it must sit above the link).

lrnselfreliance (commented):
## Summary

Solid, well-motivated card UI work: title colour tracks `--text`, posters are contained with ratio-preserving caps, meta moves under the title, and actions get a dedicated foot slot. The `Card` API change is clean, Archive is migrated to `actions`, and the Cypress/jest coverage is real layout and contrast work rather than mock theatre. Main gaps: Archive’s own poster path never joins the new containment, and the tag badge can detach from letterboxed posters; IconButton’s global size remap is intentional but worth a visual pass on dense `xs` call sites.

## Issue counts by severity

- bugs: 1
- suggestions: 3
- nits: 1

## Issues outside the diff

- **[bug]** app/src/components/Archive.js:390 — **[bug]** `ArchiveCard` is updated to use the new `actions` prop, but its media path still renders a full-width image with no height cap and without `.wrolpi-card-poster`. Archive screenshots are often tall page captures; at grid card width they can still produce the “media band dwarfs the title” problem this PR documents and tests for `CardPoster`. Theme samples and layout tests only cover the shared poster class, so this call site is unguarded.
  **Suggestion:** Route archive screenshots through the same containment as `CardPoster` — e.g. wrap the image in `div.wrolpi-card-poster` and drop the inline `width: '100%'` so `max-width` / `max-height` / ratio rules apply. Prefer reusing `CardPoster` or a shared helper if the archive screenshot path can be expressed that way.

PR #618, opened 2026-08-02
ui: centre the pagination strip

lrnselfreliance:
The pager sat hard against the left edge of its container, despite three call sites wrapping it in `<center>` or `text-align: center`.

### Why those wrappers never worked

Mantine builds a Pagination as a **block** root — full width, so the wrappers had nothing to centre — holding a flex `Group` that carries the controls and lays them out at `flex-start`:

```
.wrolpi-pagination        display: block   width: 600px   <- the wrappers centred this, to no effect
  └─ .mantine-Group-root  display: flex    justify-content: flex-start
```

So `justify-content` on the root does nothing at all; it has to reach the Group.

Setting `--group-justify` on the root doesn't work either, which cost me a detour: `Group` **defaults** that prop and writes the variable *inline* on its own element, where an inherited value can't reach it. The property itself is fair game — Mantine only ever reads the variable from a hashed class — so `.wrolpi-pagination .mantine-Group-root { justify-content: center }` at (0,2,0) is what lands.

The three wrappers are now plain divs keeping their margin. `<center>` is a deprecated element that was doing nothing here.

### Tests

**The gap either side is equal** — measured, not read off `justify-content`, so the claim is about where the control actually lands. With an assertion that the container really is wider than the strip, since otherwise the equality holds trivially at 0 and 0.

**And an inverse: at 240px nothing hangs off the left edge.** That's the risk centring introduces — overflow in a `flex-start` row spills right, where it can at least be scrolled to; centred, it spills *both* ways and the first page number becomes unreachable. It passed before this change and must keep passing after.

Verified red first (**0px left against 230px right**), and the wrong-element fix — centring the root rather than the Group — reddens it again.

**1080 jest / 60 suites, 160 in `ui-layout.cy.js`, clean build.**

### DIFF
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import {
     ActionInput, Button, Card, Header, Icon, IconButton, IconStack, Loading, MultiSelect, Panel, PathInput,
-    Message, Placeholder, Progress, SearchBox, Statistic, StatisticGroup, Status, TabBar, Table,
+    Message, Pagination, Placeholder, Progress, SearchBox, Statistic, StatisticGroup, Status, TabBar, Table,
     tabClassName, TextInput,
 } from './index';
 import {MemoryRouter} from 'react-router';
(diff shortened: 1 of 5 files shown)


app/src/components/ui/ui-layout.cy.js
```diff
@@ -2020,3 +2020,54 @@ describe('the search clear button matches the field it clears', () => {
         });
     });
 });
+
+describe('pagination sits in the middle of its container', () => {
+    /*
+     * Three call sites wrap the pager in `<center>` or `text-align: center` and it still sat
+     * hard against the left edge.  Both of those centre INLINE content, and Mantine's
+     * Pagination root is a flex container that fills its parent's width -- so the buttons
```
lrnselfreliance, line 2028: **[nit]** The new describe-block comment says Mantine’s “Pagination root is a flex container” with buttons laid out at `flex-start` inside it. That conflates layers: the Pagination root (`.wrolpi-pagination` / `PaginationRoot` `Box`) is a normal block; the flex container with `flex-start` is the child `.mantine-Group-root`. The matching comment in `ui.css` already describes this accurately; the test comment does not.

**Suggestion:** Align the test comment with the CSS comment (root is block/full width; Group is the flex row that needs `justify-content: center`).

lrnselfreliance (commented):
## Summary

Correct, well-scoped fix for pagination that never actually centred under Mantine. The root cause analysis is right: `<center>` / `text-align: center` only centre inline content, while Mantine’s `Pagination` root is a full-width block holding a `Group` that lays controls out at `flex-start` via an inline `--group-justify`. Overriding `justify-content` on `.wrolpi-pagination .mantine-Group-root` is the right lever. All production call sites that tried to centre drop the ineffective wrappers while keeping `marginTop`; bare uses pick up centering for free. Cypress geometry tests are real and complementary.


## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 1

PR #619, opened 2026-08-02
dev: stop the app container linting on every rebuild

lrnselfreliance:
`react-scripts` lints in-process on every incremental rebuild. Each save pays for it, and the result is buried under a few hundred warnings the repo has carried for years — `no-useless-computed-key` across the wire-gauge tables, `react-hooks/exhaustive-deps` through most of `Common.js`. Finding `Compiled successfully` in that output takes a scroll.

The `app` container only ever serves; it never builds. Linting still happens where it is acted on — `npm run build` and CI both run it, and both fail on a real error.

### It is not a memory fix, and I said it was

I reached for this after the dev server was OOM-killed three times in an afternoon, and told Roland it would be "the biggest single saving." That was a guess presented as a fact. Measuring it, with the same procedure each time (recreate the container, load a page, wait for the typecheck to finish, read `docker stats`):

| configuration | settled |
|---|---|
| linting on | 3.53 GB |
| linting off | 3.03 GB |
| linting off, again | 3.75 GB |

**The spread between two runs of the same configuration is wider than the spread between configurations.** Whatever the effect is, it is below what this measurement can resolve, and I am not going to claim it.

What actually stopped the kills was raising Docker Desktop's VM limit from 8 GB (its default, unrelated to the 48 GB host) to 16 GB.

Keeping the change anyway, on the rebuild time and the log noise — neither of which is in doubt.

### Verification

Zero lint lines in the container log after the change, and the compile line went from `webpack compiled with 1 warning` to `webpack compiled successfully`.

### DIFF
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import {
     ActionInput, Button, Card, Header, Icon, IconButton, IconStack, Loading, MultiSelect, Panel, PathInput,
-    Message, Placeholder, Progress, SearchBox, Statistic, StatisticGroup, Status, TabBar, Table,
+    Message, Pagination, Placeholder, Progress, SearchBox, Statistic, StatisticGroup, Status, TabBar, Table,
     tabClassName, TextInput,
 } from './index';
 import {MemoryRouter} from 'react-router';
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -90,6 +90,17 @@ services:
...
+      # save and buries the compile result under a few hundred warnings the repo has carried
+      # for years.  This container only serves; linting still happens where it is acted on --
+      # `npm run build` and CI both run it, and both fail on a real error.
+      #
+      # NOT a memory fix, though it was tried as one after the dev server was OOM-killed three
+      # times in an afternoon.  Measured either way, the container settles anywhere between
+      # 3.0GB and 3.7GB and the spread between two runs of the SAME configuration is wider
+      # than the spread between configurations.  Raising Docker Desktop's VM limit is what
+      # actually stopped the kills.
+      - DISABLE_ESLINT_PLUGIN=true
```
lrnselfreliance, line 103: **[suggestion]** This PR mixes two independent concerns: (1) disabling the eslint-webpack-plugin in the dev app container, and (2) fixing pagination centering across Files/Map/Zim/ui.css with Cypress coverage. The pagination work is coherent and well-tested on its own, but bundling it here makes review, bisect, and revert harder, and the PR title/commit framing only describe the eslint change.

**Suggestion:** Land `DISABLE_ESLINT_PLUGIN` alone (or re-title/re-summarize the PR if the pagination work is intentionally included). Prefer a dedicated UI PR for the pager centering so reviewers and history stay focused.

docker-compose.yml
```diff
@@ -90,6 +90,17 @@ services:
     environment:
       # Force React to use Caddy port to connect to it's websocket.
       - WDS_SOCKET_PORT=0
+      # react-scripts lints in-process on every incremental rebuild, which costs time on each
+      # save and buries the compile result under a few hundred warnings the repo has carried
+      # for years.  This container only serves; linting still happens where it is acted on --
+      # `npm run build` and CI both run it, and both fail on a real error.
```
lrnselfreliance, line 96: **[nit]** The comment claims “`npm run build` and CI both run it, and both fail on a real error.” That is roughly true for ESLint *errors*: CI’s “Build app” step still runs `npm run build` without `DISABLE_ESLINT_PLUGIN`, so eslint-webpack-plugin remains active there and in production `scripts/build_app.sh`. However, `.github/workflows/jest-tests.yml` deliberately sets `CI: ''` on the build step so pre-existing *warnings* are non-blocking. There is also no standalone `npm run lint` script or dedicated lint job — only the webpack plugin during build. The comment slightly oversells “CI enforces lint” for the day-to-day warning noise this PR is trying to escape.

**Suggestion:** Tighten the comment to something like: lint still runs in `npm run build` (CI + production packaging); only the docker `app` dev server skips it. Optionally note that CI already treats warnings as non-fatal via `CI: ''`.

app/src/components/ui/ui.css
```diff
@@ -190,6 +190,28 @@ html[data-theme="amber"] .wrolpi-theme-option-active {
...
+ * Setting `--group-justify` on the root does not work either: Group defaults that prop and
+ * writes the variable INLINE on its own element, where an inherited value cannot reach it.
+ * The property itself is fair game, though -- Mantine only ever reads the variable from a
+ * hashed class -- so this rule at (0,2,0) is what actually lands.
+ *
+ * The Group already wraps by default (`--group-wrap` falls back to `wrap`), which matters
+ * once the row is centred: centred overflow spills off BOTH edges, and the first page would
+ * be unreachable rather than merely off-screen to the right.
+ */
+.wrolpi-pagination .mantine-Group-root {
```
lrnselfreliance, line 211: **[nit]** Centering depends on Mantine’s static class `mantine-Group-root` (emitted only while `withStaticClasses` remains true, the MantineProvider default). That works today and is better-explained than most overrides, but a Mantine upgrade or `withStaticClasses: false` would silently stop centering without a compile error.

**Suggestion:** Prefer a structural selector under the stable house class, e.g. `.wrolpi-pagination > * { justify-content: center; }`, or document the `withStaticClasses` dependency next to the rule. The existing Cypress gap test would catch a silent break if it is run.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -2020,3 +2020,54 @@ describe('the search clear button matches the field it clears', () => {
         });
     });
 });
+
+describe('pagination sits in the middle of its container', () => {
```
lrnselfreliance, line 2024: **[nit]** The new pagination geometry tests are high quality (measure real gaps, guard left-edge clipping, assert slack exists so equality is not vacuously 0/0). They are not exercised by GitHub Actions (jest-tests only runs Jest + `npm run build`; Cypress is local/AGENTS.md). That is a pre-existing CI gap, not introduced here, but it means the main regression net for this UI fix is manual.

**Suggestion:** No change required for this PR unless you already planned Cypress in CI; worth a follow-up if layout regressions keep shipping unnoticed.

lrnselfreliance (commented):
## Summary

The core change — `DISABLE_ESLINT_PLUGIN=true` on the docker-compose `app` service only — is correct for CRA/react-scripts 5, scoped to the development container (`npm run start`), and does not disable lint for production builds or CI’s `npm run build` step. Pagination centering is a well-diagnosed Mantine flex issue with solid Cypress geometry tests, but it is unrelated scope in a PR titled around eslint/dev rebuild cost. No correctness bugs found; approve with small process/docs nits.


## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 3

PR #620, opened 2026-08-02
fix: the map viewer was blank in night and amber

lrnselfreliance:
`MapViewer` handed protomaps-themes-base the **WROLPi theme name** as its basemap "flavor". That worked while the only themes were `light` and `dark`, and broke the day night and amber arrived — protomaps looks the flavor up in a record and dereferences the result without checking, so an unknown name is not a fallback but:

```
TypeError: Cannot read properties of undefined (reading 'background')
    at nolabels_layers (base_layers.ts:18)
```

thrown before the first tile. MapLibre never constructed, so `/map` rendered nothing in **two of the four themes**.

### The mapping

`mapFlavor()` in `themes/names.ts`. Only `light` and `dark` are available, and the reason is not obvious: the flavor is also interpolated into `/map-assets/sprites/<flavor>`, and those are the two sprite sets we ship. protomaps' `black` and `grayscale` are real flavors that would load a style and then find no icons.

Night and amber take `dark` and let the media filter tint the canvas from there. Handing them `light` would not work even with the filter on — a filter shifts hue, so a white basemap becomes a red one, which is the opposite of what a night-vision theme is for.

### Two layers of test, covering different halves

**`map-flavor.test.js`** checks every theme against the *real* protomaps — calling `layers()` and requiring it not to throw — and against the sprites directory *on disk*. Deliberately not against a list of valid names copied into the test, since a copied list is exactly what would have passed here in the first place. It fails the day a fifth theme is added without a mapping.

**`map-themes.cy.js`** drives the actual page in all four themes and waits for the MapLibre canvas. The unit test can only check that the mapping returns something valid; it cannot see whether `MapViewer` *uses* it, which is the half that was broken.

Verified against the planted regression:

| theme | with `theme` passed directly (the bug) | fixed |
|---|---|---|
| light | ✓ | ✓ |
| dark | ✓ | ✓ |
| night | **no canvas after 40s** | ✓ |
| amber | **no canvas after 40s** | ✓ |

### Not a bug

The `404` on `/blobs/map-overview.pmtiles` in the same trace. That is a `HEAD` probe for an optional planet overview, already gated on `resp.ok`, and the blob is absent in a dev checkout. The console line is unavoidable for a probe.

**1093 jest / 61 suites, clean build.**

### DIFF
--- a/app/cypress/e2e/map-themes.cy.js
+++ b/app/cypress/e2e/map-themes.cy.js
@@ -1,0 +1,44 @@
+/*
+ * The map basemap has a "flavor" of its own, and MapViewer used to hand protomaps the WROLPi
+ * theme name.  That worked while the only themes were `light` and `dark`, and broke when
+ * night and amber arrived: protomaps looks the flavor up in a record and dereferences the
+ * result without checking, so an unknown name throws
+ * `Cannot read properties of undefined (reading 'background')` before the first tile and the
+ * viewer renders nothing at all.
+ *
+ * Driven through the real page in every theme, because the unit test can only check that the
+ * mapping returns something valid -- it cannot see whether MapViewer USES it, which is the
+ * half that was broken.
+ */
+const THEMES = ['light', 'dark', 'night', 'amber'];
+
+describe('the map viewer builds a basemap in every theme', () => {
+    THEMES.forEach((theme) => {
+        it(`initialises MapLibre in ${theme}`, () => {
+            const errors = [];
+            cy.visit('/map', {
+                onBeforeLoad(win) {
+                    // Set the theme before the app boots, so the viewer builds its style once
+                    // with the theme under test rather than rebuilding after a switch.
+                    win.localStorage.setItem('color-scheme', theme);
+                    // MapViewer reports a failed init through console.error and then renders
+                    // nothing, so the console is where the defect is visible.
+                    const original = win.console.error;
+                    win.console.error = (...args) => {
+                        errors.push(args.map(String).join(' '));
+                        original(...args);
+                    };
+                },
+            });
+
+            // The canvas only exists once MapLibre constructed a style successfully.
+            cy.get('canvas.maplibregl-canvas', {timeout: 40000}).should('exist');
+
+            cy.then(() => {
+                const failures = errors.filter(message =>
+                    /MapLibre failed to initialize|reading 'background'/.test(message));
+                expect(failures.join('\n')).to.equal('');
+            });
+        });
+    });
+});
--- a/app/cypress/e2e/map-themes.cy.js
+++ b/app/cypress/e2e/map-themes.cy.js
@@ -1,3 +1,5 @@
+import {mapSprite, themeNames, themeSessionKey} from '../../src/themes/names';
+
 /*
  * The map basemap has a "flavor" of its own, and MapViewer used to hand protomaps the WROLPi
  * theme name.  That worked while the only themes were `light` and `dark`, and broke when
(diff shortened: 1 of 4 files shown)


lrnselfreliance:
Pushed a second commit answering the review question: **the monochrome themes now draw protomaps' `black` basemap, not `dark`.** Measured rather than guessed:

| flavor | distinct colours | hued | earth luminance |
|---|---|---|---|
| `dark` | 19 | **4** (max chroma 25) | 0.014 |
| `black` | 13 | **0** (max chroma 0) | **0.007** |

Both media filters are the same pure luminance projection — `0.2126 0.7152 0.0722` — so they keep brightness and discard hue. `black` is *already* achromatic, so filtering it is a hue rotation and nothing else: all thirteen greys were authored as a monochrome hierarchy and every one arrives intact. `dark` has more colours, but four are hued (water, parks) and those distinctions do not survive at all — those features land at whatever brightness their hue happened to carry rather than one anybody chose. Black also emits half the light on the earth fill, which is the point of a night-vision theme.

`grayscale` is not an option despite the name: it is a **light** theme (earth L0.604), and a luminance filter turns a light map into a bright red one.

### The sprite constraint was my mistake

The first commit claimed `black` was impossible because we only ship `light`/`dark` sprites. `sprite` and the layer flavor are **separate style properties** that merely shared one variable. A `black` basemap with `dark` icons is an ordinary style. They are separate arguments now, with `mapSprite()` beside `mapFlavor()`.

### Fixed in passing

Contour labels branched on `flavor === "dark"` and would have taken the **light** branch for `black` — dark brown text on a black map. Now branches on `light`, so every dark flavor gets the pale label.

### The test worth having

The monochrome themes' flavor must paint with greys only — paired with an assertion that `dark` is **not** achromatic. Without that pair the first passes on any flavor at all and proves nothing about the choice.

**1096 jest / 61 suites, 4 e2e themes, clean build.**

app/cypress/e2e/map-themes.cy.js
```diff
@@ -0,0 +1,44 @@
...
+ * night and amber arrived: protomaps looks the flavor up in a record and dereferences the
+ * result without checking, so an unknown name throws
+ * `Cannot read properties of undefined (reading 'background')` before the first tile and the
+ * viewer renders nothing at all.
+ *
+ * Driven through the real page in every theme, because the unit test can only check that the
+ * mapping returns something valid -- it cannot see whether MapViewer USES it, which is the
+ * half that was broken.
+ */
+const THEMES = ['light', 'dark', 'night', 'amber'];
```
lrnselfreliance: **[nit]** `THEMES` is a hardcoded `['light', 'dark', 'night', 'amber']` list, while the unit suite iterates `themeNames` from `names.ts`. A fifth theme would get mapping coverage from Jest automatically but would not be cold-started by Cypress until this list is updated by hand.

**Suggestion:** Import `themeNames` (or share the same source of truth) so e2e and unit stay aligned when themes are added.

app/cypress/e2e/map-themes.cy.js
```diff
@@ -0,0 +1,44 @@
...
+
+describe('the map viewer builds a basemap in every theme', () => {
+    THEMES.forEach((theme) => {
+        it(`initialises MapLibre in ${theme}`, () => {
+            const errors = [];
+            cy.visit('/map', {
+                onBeforeLoad(win) {
+                    // Set the theme before the app boots, so the viewer builds its style once
```
lrnselfreliance, line 34: **[suggestion]** E2e only boots `/map` with `color-scheme` already set; it never switches theme while the MapLibre instance is open. The rebuild path (`map.setStyle(...)` in the `[theme, mapReady, scaleUnit]` effect) does call `mapFlavor`/`mapSprite`, so the original TypeError should not recur there, but a future regression that only breaks the live-switch branch (e.g. capturing flavor once at init and ignoring later theme changes) would still pass this suite. Live switch is also not covered for sprite URL correctness (a wrong sprite tends to 404 icons rather than throw, so the canvas can still exist).

**Suggestion:** Optionally add one Cypress case that loads in `light`, switches to `night`/`amber` via the theme control, and re-asserts the canvas still exists with no `MapLibre failed` / `background` console errors. A shallow style-URL assertion for `/map-assets/sprites/dark` under monochrome would close the sprite-wiring gap the unit tests cannot see.

lrnselfreliance (commented):
## Summary

This is a solid, well-scoped fix for a real blank-map failure: WROLPi theme names (`night`, `amber`) were passed straight into protomaps-themes-base, which only accepts flavors such as `light`/`dark`/`black` and throws on unknown keys. `mapFlavor` / `mapSprite` correctly separate basemap flavor from shipped sprite sets (monochrome → `black` + `dark` sprites), both MapViewer call sites (init and theme rebuild) use them, and the contour label light/dark branch was flipped to treat non-`light` flavors (including `black`) as dark. Tests are unusually good: unit tests call real `layers()` and read sprite files on disk, and the Cypress suite cold-starts `/map` in every theme so a regression that drops the mapping again fails on night/amber.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

lrnselfreliance:
Both comments were right, and the second named a real blind spot.

### Hardcoded theme list

Fixed — the spec imports `themeNames` and `themeSessionKey` from `names.ts`, so a fifth theme is cold-started the day it is added rather than the day someone remembers to edit the list.

### The sprite was invisible to the suite

This was the good catch. A wrong sprite name 404s the icons and MapLibre carries on, so the canvas exists and the suite passes — exactly as described. The spec now intercepts `/map-assets/sprites/**` and asserts the requested set equals `mapSprite(theme)`.

Planted it by collapsing the sprite back into the flavor:

| theme | canvas | sprite assertion |
|---|---|---|
| light | ✓ | ✓ |
| dark | ✓ | ✓ |
| night | **✓ (still builds)** | **fails** |
| amber | **✓ (still builds)** | **fails** |

Night and amber get *past* the canvas assertion and fail on the sprite, which is the point.

### The rebuild branch cannot be reached from a browser

I went to write the live-switch case and there is nothing to click: the theme picker lives on Settings and in the gallery, **never in the nav**, so there is no way to change the theme with `/map` open — navigating to Settings unmounts the viewer. The branch is real code that a future compact nav picker would light up, but no Cypress case can drive it today.

Guarded in the source instead: both `buildStyle` calls must pass `mapFlavor(theme)` and `mapSprite(theme)`, and none may pass a bare `theme`. Crude, and the honest alternative to a browser test that cannot reach the branch it claims to cover.

Planted by breaking **only** the rebuild call — the source guards go red and the e2e stays green. That split is the whole argument for keeping it.

One note on that guard: the call-site scan counts parens rather than matching a regex. The arguments are themselves calls, so anything non-greedy stops at `mapFlavor(theme)` and reports half an argument list as the whole of it — which is how the first version failed while the code was correct.

**1098 jest / 61 suites, 4 e2e themes, clean build.**

PR #621, opened 2026-08-02
ui: the navbar measures its foreground instead of assuming one

lrnselfreliance:
The navbar's icons were hard to read. They were hard to read in a specific, measurable way.

## What was wrong

The bar's background is the one surface in the app a **user** picks — twelve colour names, which each theme resolves to twelve of its own colours. That is forty-eight backgrounds, and the links and status icons were drawn on all of them in a single token, `--btn-text`, which is near-black in three of the four themes.

Night's `--brown` is `#451212`. A near-black glyph on it is **1.33:1**.

| | below 4.5:1 | below 3:1 |
|---|---|---|
| light | 1 of 12 | 0 |
| dark | 0 | 0 |
| night | 9 of 12 | 7 |
| amber | 7 of 12 | 5 |

`--btn-text` is the right token for a button, whose fill the theme chooses. It was never the right one here; it was applied without looking at what it was being drawn on.

## The fix

The foreground is measured against the background the theme actually resolved, choosing between that theme's own `--btn-text`, `--black` and `--white`.

Its *own*, not literal ones — a white pixel in night mode undoes the dark adaptation the theme exists to protect, so night's "white" is `#ff8a8a` and its "black" is `#2a0808`. Measuring at runtime rather than shipping an `--on-red` token per colour also means the custom YAML themes still ahead of us get this for free: a theme supplies a palette, and the correct foreground for each entry falls out of it.

**Nothing is below 3:1 any more. The worst case is 3.27:1, up from 1.33:1, and no combination is worse than it was.**

## What measurement cannot fix — for you to decide

Ten of the forty-eight are still under 4.5:1, and no choice of foreground will lift them, because the shortfall is in the **background**. A monochrome theme resolves all twelve colour names onto one red (or amber) ramp, and a mid-tone of a hue cannot carry text of that same hue. Night's `--green` is `#b03030`: 3.27:1 from the theme's darkest red and 2.52:1 from its brightest. The one this picks is the best available, and it is still short.

`amber/blue` `amber/grey` `amber/teal` `light/yellow` `night/blue` `night/green` `night/grey` `night/pink` `night/teal` `night/violet`

So **/theme-sample now draws all twelve bars with the measured ratio printed beside each one**, and calls out the ten. That is the thing you asked to see, and it is where this gets decided — the options are retuning the monochrome palettes, giving the bar its own colour scale, or accepting a 3.3:1 floor on ten combinations nobody may ever pick.

## Two things I deliberately left alone

- **The warning-icon colours.** `conflictingColors` in Nav.js is the same defect solved by hand — a hardcoded list standing in for a measurement. It is worse than I expected: `--danger`/`--warning` measure **1.0–1.8:1** against the bar in *light and dark too*, because a red icon on a red bar is a red icon on a red bar. Fixing it properly means encoding severity as something other than hue (a filled badge, say), which is a redesign rather than a patch. Flagging it rather than bundling it.
- **`--btn-text` elsewhere.** Only the bar changed.

## Testing

- `themes/navColors.test.js` — 17 cases. Measures all forty-eight against the palette **parsed out of tokens.css**, not a list restated in the spec. Asserts the pick is the best available, that no combination got worse, that all clear 3:1, and pins the ten that don't clear 4.5:1 as a list so it cannot quietly grow. Paired with the inverse (`--btn-text` does *not* clear 3:1) so the suite cannot pass on the mapping it replaced.
- Neither bar in Nav.js is reachable from a test that could measure it — `<NavBar/>` needs the settings, status and worker contexts and a dozen polling hooks, and a media query picks which of the two renders. So a **source guard** holds both to `navBarStyle(navColors)`, the same technique used for MapViewer's unreachable branch.
- `ui-layout.cy.js` — 17 new cases (160 → 177). Measures the **painted** result in a real browser in all four themes, checks the icons inherit what the links get, and holds the painted ratio against the number the component reports.
- **Planted the revert.** A fixed foreground reddens 4 jest cases and 3 cypress ones.

1115 jest / 62 suites, 177 in `ui-layout.cy.js`, clean `npm run build`.

## Incidental

The WCAG primitives move to `themes/contrast.ts` so theme code can measure a colour without importing `Common.js`, which pulls in half the component library — `themes/` importing it is how the last import cycle started.

`Common.js` forwards `contrastRatio` with a **delegating function**, not `export {contrastRatio}`, and that is load-bearing. A bare re-export compiles to a getter; six specs mock the module with `{...jest.requireActual('./Common')}`; a spread invokes every getter; and those spreads run re-entrantly while the module is still evaluating. The getter fired before its import binding existed and took six suites down. Found it by running them, not by reasoning about it.

### DIFF
--- a/app/src/components/Nav.js
+++ b/app/src/components/Nav.js
@@ -111,8 +111,18 @@ function DropdownMenuItem({link}) {
  * button that eventually rendered -- the row overflowed by that much and the bar wrapped.
  * Measuring one thing and drawing another is only ever right by coincidence.
  */
-const NavDropdownTrigger = React.forwardRef(({text, ...props}, ref) =>
-    <button type='button' ref={ref} className='wrolpi-navbar-link wrolpi-navbar-link-button'
+const NavDropdownTrigger = React.forwardRef(({text, className, ...props}, ref) =>
+    /*
+     * `className` is pulled out and MERGED, never spread over.  Menu.Target clones its child
+     * and hands it a className of its own, and a `{...props}` spread placed after the
+     * attribute replaces ours with it -- which left the real More button with no class at
+     * all, so it rendered as a bare grey UA button instead of taking the bar's colour, its
+     * padding and its hover.  The hidden placeholder is not cloned by Menu.Target, so it
+     * kept its class and measured wider than the button that actually drew.
+     */
+    <button type='button' ref={ref}
+            className={['wrolpi-navbar-link', 'wrolpi-navbar-link-button', className]
+                .filter(Boolean).join(' ')}
             {...props}>
         {text}
         <Icon name='dropdown' size='small' style={{marginLeft: '0.35em'}}/>
--- a/app/src/components/Nav.js
+++ b/app/src/components/Nav.js
@@ -208,7 +208,7 @@ export function NavBar() {
      * user's colour, not taken from a single token.  See themes/navColors.
      */
     const navColors = useNavColors(navColor);
-    const wrolpiIcon =<img src='/icon.svg' height='32px' width='32px' alt='WROLPi Home Icon'/>;
+    const wrolpiIcon = <img src='/icon.svg' height='32px' width='32px' alt='WROLPi Home Icon'/>;
     const name = <i>{NAME || wrolpiIcon}</i>;
     const topNavText = wrolModeEnabled ? <>{name}&nbsp; <Icon name='lock'/></> : name;
 
--- a/app/src/components/Nav.js
+++ b/app/src/components/Nav.js
@@ -102,15 +102,30 @@ function DropdownMenuItem({link}) {
     </Menu.Item>
 }
 
+/*
+ * The trigger for a navbar dropdown.
+ *
+ * Shared with the hidden placeholder DesktopNav measures on its first render, and that is
+ * the point of extracting it.  The placeholder used to be a plain `<button>More</button>`
+ * with no caret, so the space reserved for the overflow menu was about 21px short of the
+ * button that eventually rendered -- the row overflowed by that much and the bar wrapped.
+ * Measuring one thing and drawing another is only ever right by coincidence.
+ */
+const NavDropdownTrigger = React.forwardRef(({text, ...props}, ref) =>
+    <button type='button' ref={ref} className='wrolpi-navbar-link wrolpi-navbar-link-button'
+            {...props}>
+        {text}
+        <Icon name='dropdown' size='small' style={{marginLeft: '0.35em'}}/>
+    </button>
+);
+NavDropdownTrigger.displayName = 'NavDropdownTrigger';
+
 function DropdownLinks({link}) {
     // A labelled dropdown trigger: the desktop "More" overflow, or any nested
     // link group handed to MenuLink.
     return <Menu position='bottom-end' withinPortal>
         <Menu.Target>
-            <button type='button' className='wrolpi-navbar-link wrolpi-navbar-link-button'>
-                {link.text}
-                <Icon name='dropdown' size='small' style={{marginLeft: '0.35em'}}/>
-            </button>
+            <NavDropdownTrigger text={link.text}/>
         </Menu.Target>
         <Menu.Dropdown>
             {link.links.map(l => <DropdownMenuItem key={l.key} link={l}/>)}
@@ -130,9 +145,23 @@ function MobileMenu({links}) {
     </Menu>
 }
 
-function NavIconWrapper({children}) {
+/*
+ * One status indicator's slot in the right-hand corner.
+ *
+ * This used to carry `marginTop: 0.8em`, a nudge left over from Semantic's menu.  The corner
+ * centres its children, and a top margin is part of the margin box being centred, so the
+ * nudge pushed every indicator down by half of it -- about 6px.
+ *
+ * That read as an error on some and not others, which is why it survived: the bare-icon
+ * anchors (share, search) sit high inside their line box because it reserves descender space
+ * for text that is not there, so the nudge happened to cancel out.  An IconButton is a square
+ * ActionIcon that centres its glyph exactly, so for the hotspot and the theme picker the
+ * nudge was 6px of pure error.  Laying the slot out as a flex box centres both shapes on
+ * their real glyph box and needs no nudge at all.
+ */
+export function NavIconWrapper({children}) {
     if (children) {
-        return <div style={{marginTop: '0.8em', marginLeft: '1.5em'}}>{children}</div>
+        return <div className='wrolpi-navbar-icon'>{children}</div>
     } else {
         // Do not use navbar space if children is empty.
         return <React.Fragment/>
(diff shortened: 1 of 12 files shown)


app/src/components/Nav.js
```diff
@@ -163,7 +203,12 @@ export function NavBar() {
...
     const navColor = useNavColorSetting();
-    const wrolpiIcon = <img src='/icon.svg' height='32px' width='32px' alt='WROLPi Home Icon'/>;
+    /*
+     * The bar's foreground is measured against the background the theme resolved for the
+     * user's colour, not taken from a single token.  See themes/navColors.
+     */
+    const navColors = useNavColors(navColor);
+    const wrolpiIcon =<img src='/icon.svg' height='32px' width='32px' alt='WROLPi Home Icon'/>;
     const name = <i>{NAME || wrolpiIcon}</i>;
     const topNavText = wrolModeEnabled ? <>{name}&nbsp; <Icon name='lock'/></> : name;
```
lrnselfreliance, line 214: **[suggestion]** Status/warning icons still pick colour via the old `conflictingColors` name list (`null`/`black` vs `yellow`/`red`) rather than measuring against the resolved bar background the way links and neutral icons now do. On monochrome mid-tone bars this can leave `var(--red)` / `var(--yellow)` / `var(--black)` weak or muddy while the rest of the bar is intentionally measured. The gallery sample does not exercise this path either: `ThemeSamplePage.js` draws bare warning Icons that inherit the measured bar colour, so it cannot show the real warning-icon failure modes.

**Suggestion:** Either measure warning candidates against `navColors.background` (same `bestForeground` / palette tokens), or document the leftover explicitly next to the 4.5:1 residue (and make the sample use `colorVar` the way `NavBar` does so the gap is visible).

app/src/themes/contrast.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * These live in `themes/` rather than in Common.js because theme code needs them and
+ * Common.js is a large module that imports half the component library -- `themes/`
+ * importing it is how an import cycle starts, and this app has already paid for one of
+ * those (see import-cycles.test.js).  Common.js re-exports `contrastRatio` from here so
+ * its existing callers are unaffected.
+ */
+
+/** A hex colour as [r, g, b], accepting both `#rrggbb` and the shorthand `#rgb`. */
+export function hexToRGBArray(color: string): number[] | undefined {
```
lrnselfreliance, line 12: **[suggestion]** Measurement only accepts `#rgb` / `#rrggbb`. `relativeLuminance` returns 0 for anything else (`rgb()`, `hsl()`, `var(--x)`, 8-digit hex). `navColors.ts` claims custom YAML themes will get correct foregrounds "for free," but a non-hex token still passes the truthiness check in `navColorsFrom`, gets a wrong best-candidate against luminance 0, and can freeze a bad hex (or a non-hex background string) into the inline style.

**Suggestion:** Parse `rgb()`/`rgba()` (and ideally resolve via a temporary element / `CSS.supports` path), or reject non-hex backgrounds and fall back to `var(--name)` / `var(--btn-text)` instead of measuring garbage. Add a unit case for at least `rgb(r, g, b)`.

app/src/themes/navColors.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    React.useLayoutEffect(() => {
+        const update = () => setColors(resolveNavColors(navColor));
+        update();
+
+        if (typeof MutationObserver !== 'function') {
+            return;
+        }
+        const observer = new MutationObserver(update);
+        observer.observe(document.documentElement, {
```
lrnselfreliance, line 183: **[suggestion]** `useNavColors` only re-resolves when `data-theme` mutates. That correctly tracks ThemeProvider and the index.html bootstrap script. It will not re-measure if token values change without a `data-theme` write (future custom-theme editor patching CSS variables in place, late stylesheet application, etc.). After a successful measure, the bar holds frozen hex snapshots, so stale foreground/background can stick until something else changes `navColor` or `data-theme`.

**Suggestion:** For custom-theme work, also observe relevant custom properties (or re-resolve on a theme-tokens-changed event). Until then, a short comment that measurement assumes token values only change with `data-theme` would match the "free custom themes" claim more carefully.

app/src/themes/navColors.test.js
```diff
@@ -0,0 +1,244 @@
...
+        const failing = everyCombination()
+            .filter(({background, palette}) => contrastRatio(palette['--btn-text'], background) < 3);
+
+        expect(failing.length).toBeGreaterThan(5);
+    });
+});
+
+describe('what measurement cannot fix', () => {
+    /*
+     * Eleven of the forty-eight combinations still sit under 4.5:1, and no choice of
```
lrnselfreliance: **[nit]** Comment says "Eleven of the forty-eight combinations still sit under 4.5:1," but the locked expectation immediately below is ten combinations (`leaves ten combinations short of 4.5:1`). ThemeSamplePage correctly says ten.

**Suggestion:** Change the comment to "Ten" so the documented residue matches the assertion and gallery copy.

app/src/components/Nav.js
```diff
@@ -163,7 +203,12 @@ export function NavBar() {
     const wrolModeEnabled = useWROLMode();
     const {status} = React.useContext(StatusContext);
     const navColor = useNavColorSetting();
-    const wrolpiIcon = <img src='/icon.svg' height='32px' width='32px' alt='WROLPi Home Icon'/>;
+    /*
+     * The bar's foreground is measured against the background the theme resolved for the
+     * user's colour, not taken from a single token.  See themes/navColors.
+     */
+    const navColors = useNavColors(navColor);
+    const wrolpiIcon =<img src='/icon.svg' height='32px' width='32px' alt='WROLPi Home Icon'/>;
```
lrnselfreliance: **[nit]** Minor formatting slip: `const wrolpiIcon =<img` is missing a space after `=`. Harmless, but inconsistent with surrounding style.

**Suggestion:** `const wrolpiIcon = <img ...`.

lrnselfreliance (commented):
## Summary

Solid, well-reasoned PR: extracting WCAG primitives into `themes/contrast.ts`, measuring navbar foreground against the resolved bar background, and forcing anchors/ActionIcons to inherit fixes the real unreadable cases (e.g. night brown at ~1.33:1) without import cycles. Both desktop and mobile bars share `navBarStyle(useNavColors(...))`, inheritance is CSS-backed for the three icon routes, and tests are unusually strong (real `tokens.css` palette, painted Cypress checks, source guards, inverse cases). No ship-blocking correctness bugs found; remaining items are known leftover gaps, forward-compat for non-hex/custom themes, and small documentation nits.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 2

## Issues outside the diff

- **[suggestion]** app/src/components/ui/ui-layout.cy.js:674 — **[suggestion]** The regression guard `to.not.equal(toRgb('#0a0000'))` hardcodes night's current `--btn-text`. Unit tests carefully parse `tokens.css` so palette edits stay honest; this Cypress inverse test would either go stale or lose its teeth if that token moves.
  **Suggestion:** Read `--btn-text` from `getComputedStyle(document.documentElement)` under the night mount (same idea as the unit reader) instead of a literal hex.

PR #622, opened 2026-08-02
ui: only the top modal answers Escape, and the audited widths

lrnselfreliance:
Two things, both reported from the running app.

## Nested modals could not exist

Mantine binds a **window-level keydown listener per open modal**, gated only on whether that modal is open — nothing checks which one the user is looking at:

```js
// @mantine/core ModalBase/use-modal.mjs
useWindowEvent('keydown', (event) => {
  if (event.key === 'Escape' && closeOnEscape && !event.isComposing && opened) { ... onClose(); }
}, {capture: true});
```

Two open modals meant two listeners, so **one Escape closed both**. Reproduced on the dashboard: download modal, search modal over it, one press, both gone. They also both sat at `z-index: 200`, so which one drew on top was DOM order rather than which the user opened.

Mantine's own answer is `Modal.Stack` with a `stackId`, which is the only thing that gates `closeOnEscape`:

```js
const stackProps = ctx && stackId ? {closeOnEscape: ctx.currentId === stackId, ...} : {}
```

**Not used**, for two reasons: every call site would have to be rewritten to hoist its modal into a stack, and it sets `__hidden` on everything below the top — the parent *disappears* while the child is open. Roland asked for the parent dimmed and still visible, which is what it did before the migration.

So the wrapper keeps the register itself: a module-level list of open modals, most recent on top, from which each modal takes `closeOnEscape`, `trapFocus` and a z-index two above its parent. **Module-level rather than React context**, because the two modals in the bug aren't nested in the JSX at all — the search modal is rendered by `KeyboardShortcutsProvider` near the app root, the download modal by the dashboard. They're siblings that happen to be open together, and no context relates them. The increment is deliberately small: Mantine puts popovers at 300, and a stack climbing past that would paint over the select dropdowns inside its own forms.

## Widths

Roland audited the full modal list and chose sizes against Mantine's scale.

**Wider (16):** Download from the Internet, Upload from your device, Edit Tags, Unknown File, Add to Playlist, Add Samba Share, Edit Download, Edit Video Download, Edit Archive Download, Directory Conflict, Bulk Tag Files, Search Filters → `large` (620). Download Error ×2, Download Progress, Region Preview → `fullscreen`.

**Unchanged in pixels (11)**, but the size is now written down so the decision lives in the source rather than in a default: Tags picker, Channel tag picker, Sort By, Donate QR, Search, Share this page, Zim delete, Hotspot QR, New Inventory, Restore Backup (all `small`/440) and Unlock Cookies (`tiny`/380).

Flasher was passing Mantine's raw `sm` — worked, but bypassed the naming the audit used. It says `tiny` now.

## The size table is now guarded

It's the single thing that halved every modal in the migration, and a quiet edit moves seventy at once. Semantic's widths, **measured on the pre-migration build still running on the QA Pi**, against Mantine's, read out of its stylesheet:

| name | Semantic | Mantine | |
|---|---|---|---|
| mini | 360 | `xs` 320 | |
| tiny | 540 | `sm` 380 | −30% |
| small | 720 | `md` 440 | **−39%** |
| large | 1080 | `lg` 620 | −43% |
| *(none)* | 900 | `md` 440 | **−51%** |

Three cases hold the scale, the mapping, and the rule that no call site may name a size the table doesn't know — an unknown name is passed to Mantine, ignored, and silently rendered at the default, so a typo is a modal that quietly reverts to 440px.

## Testing

A real browser throughout: none of this is visible to jsdom — the listener is on `window` with capture, the stacking is z-index, and "still visible behind" is geometry.

One thing worth flagging: **`should('be.visible')` is the wrong assertion for the parent.** It's an occlusion check, so a child drawn correctly over its parent trips it — my first version failed on correct behaviour. The test now uses a wide viewport and asserts the parent is rendered (not `display:none`, not `visibility:hidden`, non-zero box) and shows to *both sides* of the child, which is the actual inverse of Mantine's hide-the-parent.

The register also has its own test: a modal that unmounts without deregistering would leave a ghost on top forever and then **nothing** would close on Escape — silent, permanent, and invisible to any test that only opens one modal.

**Planting the pre-fix behaviour reddens three.**

1124 jest / 62 suites, 8 in `modal-stack.cy.js`, clean `npm run build`. The 12 failures in `Tags/Channels/Download/Inventory.cy.js` are the known pre-existing Semantic-era selector failures — same four files, same count, unchanged.

### DIFF
--- a/app/src/components/ui/Overlays.tsx
+++ b/app/src/components/ui/Overlays.tsx
@@ -45,6 +45,78 @@ export interface ModalProps extends Omit<React.ComponentProps<typeof MModal>, 'o
     closeIcon?: boolean;
 }
 
+/*
+ * Nested modals: which one is on top.
+ *
+ * Mantine binds a window-level `keydown` listener per open modal, gated only on whether that
+ * modal is open -- there is no check for whether it is the one the user is looking at.  Two
+ * modals open meant two listeners, so a single Escape closed BOTH: opening the search modal
+ * over the dashboard's download modal and pressing Escape once left you on the dashboard.
+ * Semantic handled this; the migration lost it.
+ *
+ * Mantine's own answer is `Modal.Stack` with a `stackId` per modal, which is the only thing
+ * that gates `closeOnEscape`.  It is not used here for two reasons: it requires every call
+ * site to be rewritten to hoist its modal into a stack, and it sets `__hidden` on everything
+ * below the top -- the parent disappears while the child is open.  We want the parent dimmed
+ * and still visible behind the child, which is what it did before.
+ *
+ * So the wrapper keeps the register itself.  A module-level list rather than React context,
+ * because the two modals in the bug are not nested in the JSX at all: the search modal is
+ * rendered by KeyboardShortcutsProvider near the app root while the download modal is
+ * rendered by the dashboard.  They are siblings that happen to be open at once, and no
+ * context relates them.  Most recently opened is on top.
+ */
+let openModalIds: string[] = [];
+const modalStackListeners = new Set<() => void>();
+const notifyModalStack = () => modalStackListeners.forEach(listener => listener());
+
+/** Mantine's default modal z-index; popovers and tooltips sit at 300. */
+const MODAL_Z_INDEX = 200;
+
+/**
+ * Where this modal sits in the stack of open ones.
+ *
+ * Exported for the tests, which need to assert the register empties -- a modal that failed to
+ * deregister would leave every later modal believing it was not on top, and nothing would
+ * close on Escape at all.
+ */
+export const openModalCount = () => openModalIds.length;
+
+function useModalStackPosition(opened: boolean): {isTop: boolean; zIndex: number} {
+    const id = React.useId();
+    const [, rerender] = React.useReducer((count: number) => count + 1, 0);
+
+    React.useEffect(() => {
+        modalStackListeners.add(rerender);
+        return () => {
+            modalStackListeners.delete(rerender);
+        };
+    }, []);
+
+    React.useEffect(() => {
+        if (!opened) return;
+        // Re-opening an already-registered modal moves it to the top, which is what a user
+        // who clicked its trigger again means.
+        openModalIds = [...openModalIds.filter(other => other !== id), id];
+        notifyModalStack();
+        return () => {
+            openModalIds = openModalIds.filter(other => other !== id);
+            notifyModalStack();
+        };
+    }, [opened, id]);
+
+    const depth = openModalIds.indexOf(id);
+    return {
+        isTop: depth === -1 || depth === openModalIds.length - 1,
+        /*
+         * Two per level, so a child's overlay clears its parent's content.  Kept small on
+         * purpose: Mantine puts popovers and select dropdowns at 300, and a modal stack that
+         * climbed past that would paint over the dropdowns inside itself.
+         */
+        zIndex: MODAL_Z_INDEX + Math.max(depth, 0) * 2,
+    };
+}
+
 function ModalBase({open, opened, size, closeIcon, title, children, ...props}: ModalProps) {
     const slots: Record<string, React.ReactNode[]> = {header: [], actions: [], body: []};
     React.Children.forEach(children, child => {
@@ -54,13 +126,25 @@ function ModalBase({open, opened, size, closeIcon, title, children, ...props}: M
         else slots.body.push(child);
     });
 
+    const isOpen = opened ?? open ?? false;
+    const {isTop, zIndex} = useModalStackPosition(isOpen);
+
     return <MModal
-        opened={opened ?? open ?? false}
+        opened={isOpen}
         centered
         size={typeof size === 'string' ? (modalSizes[size] ?? size) : size}
         title={slots.header.length ? slots.header : title}
         // Flat surfaces: the overlay separates the modal from the page, not a shadow.
         overlayProps={{backgroundOpacity: 0.55, blur: 0}}
+        /*
+         * Only the modal on top answers the keyboard.  Without this every open modal has its
+         * own window-level Escape listener and one press closes the lot.  Focus goes with it:
+         * two traps fighting over the same Tab is its own bug, and the user is working in the
+         * top one.
+         */
+        closeOnEscape={isTop}
+        trapFocus={isTop}
+        zIndex={zIndex}
         {...props}
     >
         {slots.body}
--- a/app/src/components/ui/Overlays.tsx
+++ b/app/src/components/ui/Overlays.tsx
@@ -107,7 +107,13 @@ function useModalStackPosition(opened: boolean): {isTop: boolean; zIndex: number
 
     const depth = openModalIds.indexOf(id);
     return {
-        isTop: depth === -1 || depth === openModalIds.length - 1,
+        /*
+         * Not yet registered -- the first paint of a modal that just opened, before the
+         * effect runs.  It is only on top if nothing else is open; claiming top while a
+         * parent is registered gives both modals the same z-index for a frame, which is a
+         * visible flash of the wrong one in front when a child opens.
+         */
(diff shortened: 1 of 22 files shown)


app/src/components/ui/Overlays.tsx
```diff
@@ -54,13 +126,25 @@ function ModalBase({open, opened, size, closeIcon, title, children, ...props}: M
...
         title={slots.header.length ? slots.header : title}
         // Flat surfaces: the overlay separates the modal from the page, not a shadow.
         overlayProps={{backgroundOpacity: 0.55, blur: 0}}
+        /*
+         * Only the modal on top answers the keyboard.  Without this every open modal has its
+         * own window-level Escape listener and one press closes the lot.  Focus goes with it:
+         * two traps fighting over the same Tab is its own bug, and the user is working in the
+         * top one.
+         */
+        closeOnEscape={isTop}
```
lrnselfreliance: **[bug]** Stack-gated `closeOnEscape`, `trapFocus`, and `zIndex` are set and then immediately overridden by `{...props}` on the next line. Call sites that pass `closeOnEscape` (and could pass the others) win, so a non-top modal can still register Mantine’s window-level Escape listener. Mantine’s own `Modal.Stack` does the opposite — it spreads stack props *after* caller props so the stack always wins. Concrete breakage: `CollectionReorganizeModal` keeps `closeOnEscape={!isInProgress}` while `ConflictResolutionModal` opens as a sibling on top; when not in progress, Escape closes *both*. The same pattern exists in `BatchReorganizeModal` and `Flasher`’s flash modal, and the original search-over-download bug returns for any of those parents when search (or another modal) opens above them.

**Suggestion:** Destructure `closeOnEscape`, `trapFocus`, and `zIndex` out of props (defaults `true`/`true`/undefined), then apply:
`closeOnEscape={isTop && closeOnEscape}`,
`trapFocus={isTop && trapFocus}`,
`zIndex={zIndexProp ?? zIndex}`,
after `{...props}` (or omit them from the spread). Compose “only top” with the caller’s intentional disable during long-running work. Add a Cypress case: parent with `closeOnEscape={!busy}` plus a child open; one Escape must close only the child.

app/src/components/ui/Overlays.tsx
```diff
@@ -45,6 +45,78 @@ export interface ModalProps extends Omit<React.ComponentProps<typeof MModal>, 'o
...
+        openModalIds = [...openModalIds.filter(other => other !== id), id];
+        notifyModalStack();
+        return () => {
+            openModalIds = openModalIds.filter(other => other !== id);
+            notifyModalStack();
+        };
+    }, [opened, id]);
+
+    const depth = openModalIds.indexOf(id);
+    return {
```
lrnselfreliance, line 109: **[suggestion]** Until the open-effect registers the id, `depth === -1` and the hook reports `isTop: true` with `zIndex: 200`. On the first paint of a newly opened child, parent and child can both briefly claim top / the same z-index before `notifyModalStack` re-renders. Unlikely to matter for Escape after paint, but stacking can flash incorrectly for one frame when opening a second modal.

**Suggestion:** Optional: treat unregistered-but-open as top only when the stack is empty (`depth === -1 ? openModalIds.length === 0 : depth === length - 1`), or register synchronously during render (with care for StrictMode / concurrent rules). Low priority if no visible flash in practice.

app/src/components/ui/modal-stack.cy.js
```diff
@@ -0,0 +1,199 @@
...
+        // worthless if the top does not respond either.
+        cy.mountUI(<Nested/>);
+        cy.get('[data-testid="open-parent"]').click();
+        cy.contains('.mantine-Modal-title', 'Parent').should('exist');
+
+        pressEscape();
+
+        cy.contains('.mantine-Modal-title', 'Parent').should('not.exist');
+        expect(titles()).to.not.include('Parent');
+    });
```
lrnselfreliance, line 254: **[suggestion]** After the Cypress assertions, bare `expect(titles()).to.not.include('Parent')` runs in the synchronous phase of the test function (before any `cy.*` commands execute), not after Escape. At that moment no modal is open, so the assertion always passes and does not guard the behavior under test. Elsewhere the file correctly uses `cy.then(() => expect(...))` for register counts.

**Suggestion:** Drop the dead `expect`, or wrap it in `cy.then(() => …)` after the close assertion if you still want a titles() check.

app/src/components/ui/ui.test.js
```diff
@@ -1487,3 +1487,75 @@ describe('Confirm', () => {
...
+            .map(m => [m[1], m[2]]));
+
+        expect(pairs).toEqual({
+            mini: 'xs', tiny: 'sm', small: 'md', large: 'lg', fullscreen: '100%',
+        });
+    });
+
+    it('never lets a call site name a size the table does not know', () => {
+        /*
+         * A name the table has never heard of is passed straight to Mantine, which ignores
```
lrnselfreliance: **[nit]** The comment says unknown size names are “passed straight to Mantine, which ignores anything it does not recognise.” Mantine *does* recognize its own tokens (`xs`/`sm`/`md`/…). The real hazard for unknown *Semantic* typos is silent fallback to default `md`; for raw Mantine names the hazard is bypassing the audited vocabulary (as with Confirm/`sm`), not ignore-and-default.

**Suggestion:** Tighten the comment so future readers do not assume Mantine drops `sm`/`md`/etc.

app/src/components/ui/ui.test.js
```diff
@@ -1487,3 +1487,75 @@ describe('Confirm', () => {
...
+                const full = path.join(directory, entry.name);
+                if (entry.isDirectory()) return entry.name === 'node_modules' ? [] : walk(full);
+                if (!/\.(js|jsx|tsx)$/.test(entry.name)) return [];
+                if (/\.(test|cy)\./.test(entry.name)) return [];
+                return [[full, fs.readFileSync(full, 'utf8')]];
+            });
+
+        for (const [file, source] of walk(path.join(__dirname, '..', '..'))) {
+            // Only `<Modal ... size='x'>`; a computed `size={...}` is the caller's business.
+            for (const match of source.matchAll(/<Modal\b[^>]*?\bsize='([^']+)'/g)) {
```
lrnselfreliance: **[nit]** The static scan only matches single-quoted `size='…'`. Double-quoted `size="…"` or template literals would slip through. Current call sites appear consistent with single quotes, so this is defensive only.

**Suggestion:** Extend the regex to allow `["']` if you want the guardrail to be quote-agnostic.

lrnselfreliance (commented):
## Summary

The modal stack approach (module-level register, `closeOnEscape`/`trapFocus`/`zIndex` only for the top modal) is the right shape for sibling modals and is covered by solid Cypress tests for the happy path. However, `{...props}` is spread *after* the stack-gated props, so existing call sites that pass `closeOnEscape` silently re-enable Escape on non-top modals — including the real nested case of reorganize + conflict resolution. Separately, `Confirm` still uses Mantine’s `size='sm'`, which the new audit test is designed to fail on. Fix those before merge; widths and register lifecycle otherwise look sound.

## Issue counts by severity

- bugs: 2
- suggestions: 2
- nits: 2

## Issues outside the diff

- **[bug]** app/src/components/ui/Overlays.tsx:188 — **[bug]** `Confirm` renders `<Modal ... size='sm'>`. That is a raw Mantine size, not a Semantic name from `modalSizes`. The new audit test in `ui.test.js` walks every `app/src` `js|jsx|tsx` file and fails on any `<Modal ... size='…'>` not in `{mini,tiny,small,large,fullscreen}` — `Overlays.tsx` itself is included, so `size='sm'` makes the suite fail. Functionally `sm` still maps through as Mantine `sm` (380px, same as Semantic `tiny`), but the PR specifically fixed Flasher’s `sm` → `tiny` for this rule and left Confirm as the remaining offender.
  **Suggestion:** Change Confirm to `size='tiny'` (or another audited Semantic name if the confirm dialog should be wider). Re-run the `modal sizes` tests.

PR #624, opened 2026-08-02
ui: a disabled button keeps its own colour

lrnselfreliance:
The file browser's buttons hadn't lost their colours — the **disabled** ones had.

Its footer is eight buttons, six of them disabled until something is selected, so the toolbar you meet on arriving at `/files` was an undifferentiated grey row: Delete, Rename, Move, Ignore and Tag rendered as five identical pixels. Only New Folder and Upload, the two that work with nothing selected, kept any colour.

## What changed and why

Mantine repaints a disabled button flat grey — background, text and border all replaced with `--mantine-color-disabled`. Semantic kept the colour and dropped the whole control to `opacity: 0.45`. Measured on the QA Pi, which still runs Semantic:

| | Semantic (disabled) | Mantine (disabled) |
|---|---|---|
| Delete | red, `opacity: 0.45` | `rgb(228, 230, 229)` |
| Rename | yellow, `opacity: 0.45` | `rgb(228, 230, 229)` |
| Move | teal, `opacity: 0.45` | `rgb(228, 230, 229)` |
| Tag | violet, `opacity: 0.45` | `rgb(228, 230, 229)` |

Opacity is the better signal: it says "not available" without also saying "no longer the delete button".

Mantine leaves the *variables* alone and overrides only the painted properties, so the colour the call site asked for is still sitting in `--button-bg` and can simply be reinstated. `--button-bd` is a whole border shorthand rather than a colour, so it goes back via `border`.

This applies to **every** disabled Button and ActionIcon, not just the file browser's. It's one convention, and having two would be worse than either.

## A pre-existing test earned its keep

The `IconStack` corner disc was pinned to `--mantine-color-disabled` so it would follow a disabled button's grey — otherwise a disc left on `--blue` was a blue chip on a grey surface. It reddened the moment the button stopped being grey. That override is gone: `--button-bg` is correct in both states now, so the general rule covers them, and the test's comment records both shapes the bug has taken.

## Testing

Ten new cases:

- **Fill equality** against each button's own enabled twin, in all four themes.
- **Separately, across the five colours at once** — comparing each against its twin would still pass if every pair were grey, which is precisely the bug. That assertion is what names the symptom: `five colours, distinct fills: rgb(228,230,229) rgb(228,230,229) …`.
- **Opacity from both sides** — the enabled button must be solid and the disabled one faded but not invisible, because keeping the colour must not cost the affordance.
- Icon-only buttons too, forced to `variant='filled'` so there's a colour to lose; `IconButton` defaults to a white face and comparing white with white would prove nothing.

Removing the rule reddens eleven.

1128 jest / 63 suites, 209 in `ui-layout.cy.js`, clean `npm run build`. Confirmed on the running app — the footer now reads as faded red / yellow / teal / grey / violet against solid blue and green.

### DIFF
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -2892,35 +2892,133 @@ describe('a disabled button keeps its own colour', () => {
         });
 
         it(`still marks them as unavailable in ${theme}`, () => {
-            // Keeping the colour must not cost the affordance: a disabled button that looks
-            // identical to an enabled one is a worse bug than a grey one.
+            /*
+             * Keeping the colour must not cost the affordance: a disabled button that looks
+             * identical to an enabled one is a worse bug than a grey one.
+             *
+             * Against `--disabled-opacity` rather than a band.  A band tolerates an
+             * accidentally different fade, or an opacity leaking in from somewhere else, and
+             * still reports success -- and the fade is an authored value, so it can be read.
+             */
             mountPairs(theme);
 
             cy.get('[data-testid="off-red"]').should(($off) => {
+                const authored = parseFloat(getComputedStyle(document.documentElement)
+                    .getPropertyValue('--disabled-opacity'));
                 const disabled = parseFloat(getComputedStyle($off[0]).opacity);
                 const enabled = parseFloat(
                     getComputedStyle(Cypress.$('[data-testid="on-red"]')[0]).opacity);
 
+                expect(authored, 'the token resolved').to.be.within(0.1, 0.9);
                 expect(enabled, `${theme} enabled is solid`).to.equal(1);
-                expect(disabled, `${theme} disabled is faded`).to.be.lessThan(0.7);
-                expect(disabled, `${theme} disabled is still legible`).to.be.greaterThan(0.25);
+                expect(disabled, `${theme} disabled matches the token`).to.equal(authored);
             });
         });
     });
 
-    it('tells two disabled buttons of different colours apart', () => {
+    themeNames.forEach((theme) => {
+        it(`tells two disabled buttons of different colours apart in ${theme}`, () => {
+            /*
+             * The claim stated as the user meets it, and the inverse of the bug: five grey
+             * blobs were five EQUAL greys.  Comparing each against its enabled twin above
+             * would still pass if every pair were grey, so this compares them to each other.
+             *
+             * Every theme, not just light.  Night and amber are where it is both most
+             * valuable and most fragile: one hue means the five differ only in brightness,
+             * so if any theme is going to collapse them back together it is those two.
+             */
+            mountPairs(theme);
+
+            cy.get('[data-testid^="off-"]').should(($disabled) => {
+                const fills = [...$disabled]
+                    .map(button => getComputedStyle(button).backgroundColor);
+
+                expect(new Set(fills).size,
+                    `${theme}: five colours, distinct fills: ${fills.join(' ')}`)
+                    .to.equal(pairs.length);
+            });
+        });
+    });
+
+    it('does not fade a button that is merely busy', () => {
         /*
-         * The claim stated as the user meets it, and the inverse of the bug: five grey blobs
-         * were five EQUAL greys.  Comparing each against its enabled twin above would still
-         * pass if every pair were grey, so this compares them against each other.
+         * Mantine renders a loading button as `disabled={disabled || loading}`, so every busy
+         * control in the app is `:disabled` -- an APIButton mid-request, refresh, the
+         * flasher's connect.  Mantine's own grey repaint excludes `[data-loading]` for
+         * exactly this reason, and the first version of this rule did not, so "working..."
+         * faded to 0.45 and read as unavailable rather than active.
          */
-        mountPairs('light');
+        cy.mountUI(<div>
+            <Button color='blue' loading data-testid='busy'>Saving</Button>
+            <Button color='blue' disabled data-testid='off'>Save</Button>
+            <IconButton icon='trash' label='Deleting' color='red' variant='filled' loading
+                        data-testid='busy-icon'/>
+        </div>);
 
-        cy.get('[data-testid^="off-"]').should(($disabled) => {
-            const fills = [...$disabled].map(button => getComputedStyle(button).backgroundColor);
+        cy.get('[data-testid="busy"]').should(($busy) => {
+            expect(parseFloat(getComputedStyle($busy[0]).opacity), 'a busy button is solid')
+                .to.equal(1);
+            // Paired with the disabled one, so this cannot pass by nothing fading at all.
+            expect(parseFloat(getComputedStyle(Cypress.$('[data-testid="off"]')[0]).opacity),
+                'while a disabled one beside it still fades').to.be.lessThan(1);
+            expect(parseFloat(getComputedStyle(Cypress.$('[data-testid="busy-icon"]')[0]).opacity),
+                'and the same for an icon button').to.equal(1);
+        });
+    });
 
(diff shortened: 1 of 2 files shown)


app/src/components/ui/ui.css
```diff
@@ -1455,3 +1458,39 @@ html[data-theme="amber"] .wrolpi-th-sort:focus-visible {
...
+ * Colour is how those are told apart at a glance.
+ *
+ * Mantine leaves the variables alone and overrides only the painted properties, so the colour
+ * the call site asked for is still in `--button-bg` and can simply be reinstated.  `--button-bd`
+ * is a whole border shorthand, not a colour, so it is set with `border`.
+ *
+ * The `html` prefix beats Mantine's own `.m_hash:disabled` at (0,2,0) without needing
+ * `!important`.
+ */
+html .mantine-Button-root:disabled,
```
lrnselfreliance: **[bug]** The disabled rules match every `:disabled` / `[data-disabled]` Button and ActionIcon, including loading ones. Mantine sets `disabled={disabled || loading}` on the root so a loading control is `:disabled`, but its own grey repaint carefully uses `:disabled:not([data-loading])` / `[data-disabled]:not([data-loading])` so a busy button keeps full-strength colour and the loader chrome. This PR reintroduces `opacity: 0.45` (and variable reinstatement) on loading buttons used across the app (confirm modals, refresh, flasher connect, mount, etc.), so “working…” reads as faded/disabled rather than active-but-busy.

**Suggestion:** Mirror Mantine’s exclusion on both roots, e.g. `html .mantine-Button-root:disabled:not([data-loading])`, `html .mantine-Button-root[data-disabled]:not([data-loading])`, and the same for ActionIcon. Add a Cypress case that mounts `<Button loading color="blue">` (and an IconButton) and asserts opacity stays `1` while a sibling `disabled` button is faded.

lrnselfreliance: **[bug]** `html .mantine-Button-root:disabled` / `[data-disabled]` is specificity (0,2,1), equal to `html[data-theme="night"] .wrolpi-button-danger` and later in the file, so it wins. On disable it repaints from `--button-bg` / `--button-color` / `--button-bd` — the filled red Mantine left in the variables — undoing night’s intentional transparent fill + `1.5px dashed var(--danger)` treatment. Before this PR Mantine’s disabled rule used `:where(...)`, so it lost to the night danger rule and disabled Delete stayed dashed (though without an opacity fade). After this PR, night disabled danger becomes a filled red chip at 0.45. The same override hits ActionIcon for `IconButton role="danger"`. Night encodes danger by dashed borders; the file-browser Delete control is `role="danger"` and is disabled until a selection exists.

**Suggestion:** After the general disabled rule, restate night danger for the disabled state (background/color/border from the night rule, plus `opacity: 0.45`), or structure the general rule so it only sets `opacity` when a more specific theme override already owns the paint. Cover with a test: mount `<Button role="danger" disabled/>` under `theme: 'night'`, assert transparent (or non-filled) background, dashed border, danger-coloured text, and faded opacity.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -2837,3 +2840,109 @@ describe('a toast does not cover the navigation bar', () => {
...
+
+    const mountPairs = (theme) => cy.mountUI(<div>
+        {pairs.map(({color, label}) => <span key={color}>
+            <Button color={color} data-testid={`on-${color}`}>{label}</Button>
+            <Button color={color} disabled data-testid={`off-${color}`}>{label}</Button>
+        </span>)}
+    </div>, {theme});
+
+    themeNames.forEach((theme) => {
+        it(`keeps every colour recognisable while disabled in ${theme}`, () => {
```
lrnselfreliance, line 2878: **[suggestion]** Coverage is good for the happy path (filled colour pairs across `themeNames`, opacity band, cross-colour distinctness, filled IconButton) but leaves the regressions above uncaught. Distinctness runs only in `light`; monochrome night/amber is where hue collapse makes “still recognisable” both most valuable and most fragile. Assertions only compare `backgroundColor` — outline/subtle/default variants signal colour mainly via text/border, which are restored in CSS but never checked. There is no `data-disabled` case and no loading case.

**Suggestion:** Extend the suite with: (1) loading vs disabled opacity, (2) night `role="danger"` paint + opacity, (3) distinctness under night/amber or all themes, (4) an outline/default pair asserting `color` / border stay equal to the enabled twin, (5) a `disabled` link-button using `data-disabled` if that path is used in the app.

app/src/components/ui/ui.css
```diff
@@ -1455,3 +1458,39 @@ html[data-theme="amber"] .wrolpi-th-sort:focus-visible {
...
+ * The file browser is where the difference shows.  Its footer is eight buttons, six of them
+ * disabled until something is selected, so the toolbar a user meets on arriving at /files was
+ * an undifferentiated grey row -- Delete, Rename, Move, Ignore and Tag all the same pixel.
+ * Colour is how those are told apart at a glance.
+ *
+ * Mantine leaves the variables alone and overrides only the painted properties, so the colour
+ * the call site asked for is still in `--button-bg` and can simply be reinstated.  `--button-bd`
+ * is a whole border shorthand, not a colour, so it is set with `border`.
+ *
+ * The `html` prefix beats Mantine's own `.m_hash:disabled` at (0,2,0) without needing
```
lrnselfreliance: **[nit]** Comment claims the `html` prefix beats Mantine’s `.m_hash:disabled` at (0,2,0). Mantine actually wraps the disabled selector in `:where(...)`, so its disabled paint is only (0,1,0) — the class alone. The `html` prefix is still correct and load-bearing against other (0,2,1) rules, but the stated specificity math is wrong and could mislead a future specificity fight.

**Suggestion:** Note that Mantine’s disabled block is `:where(:disabled…)` at (0,1,0), and that `html` is needed to stay even with other `html …` overrides (night danger, etc.).

app/src/components/ui/ui-layout.cy.js
```diff
@@ -2837,3 +2840,109 @@ describe('a toast does not cover the navigation bar', () => {
...
+            // identical to an enabled one is a worse bug than a grey one.
+            mountPairs(theme);
+
+            cy.get('[data-testid="off-red"]').should(($off) => {
+                const disabled = parseFloat(getComputedStyle($off[0]).opacity);
+                const enabled = parseFloat(
+                    getComputedStyle(Cypress.$('[data-testid="on-red"]')[0]).opacity);
+
+                expect(enabled, `${theme} enabled is solid`).to.equal(1);
+                expect(disabled, `${theme} disabled is faded`).to.be.lessThan(0.7);
```
lrnselfreliance: **[nit]** Opacity checks use a wide band (`< 0.7` and `> 0.25`) rather than expecting the authored `0.45`. That tolerates an accidental different fade or a theme/global opacity leak without failing, while still documenting intent loosely.

**Suggestion:** Prefer `expect(disabled).to.equal(0.45)` (or `closeTo`) now that the value is an explicit design token in ui.css; keep a lower bound only if subpixel/compositing noise appears in CI.

lrnselfreliance (commented):
## Summary

The approach is sound for the main case: reinstate Mantine’s still-live `--button-*` / `--ai-*` variables on disabled Button/ActionIcon roots and signal unavailability with `opacity: 0.45`, matching Semantic’s behaviour and fixing the grey file-browser footer. The IconStack `:disabled` grey override is correctly removed once the parent keeps its fill. Two real CSS bugs remain: loading buttons are also `:disabled` and will be faded (Mantine deliberately excludes `[data-loading]`), and the new rule outranks night’s dashed danger treatment so disabled `role="danger"` in night becomes filled red. Tests are strong on filled multi-colour / multi-theme background equality and opacity range, but miss those edge cases and only prove distinctness in light.

## Issue counts by severity

- bugs: 2
- suggestions: 1
- nits: 2

PR #625, opened 2026-08-02
ui: three button-layout defects, each with a different cause

lrnselfreliance:
Reported together as "button groups", but they're three unrelated causes. Each one answers a "why".

## Why the footer touches and Channel Edit doesn't

Channel Edit lays its buttons out with Mantine's `Group` — a flex row with a gap. The file browser's footer is hand-rolled and was a plain **block**, so its inline-block buttons sat flush against each other. There wasn't even a space character between them: JSX drops whitespace-only lines, so `<Button/>\n<Button/>` emits no text node at all.

`.sticky-footer` is now the flex row every other button group already is, wrapping because it carries eight controls and a phone can't hold them.

## Why Import/Save are off-centre and Restore isn't

**Restore is an `IconButton`** — an ActionIcon, which centres a lone glyph by construction. **Import and Save are icon-only `Button`s**, and an icon on a Button goes into Mantine's `leftSection`, which carries a `margin-inline-end` to hold it off the label. With no label that margin is pure offset. Measured on the running app: **8px left of centre**.

An icon with no label is now rendered as the button's *content* rather than as a leading section, so it centres. `React.Children.toArray` decides that, because it drops nulls — and the file browser passes `{label('Delete')}`, which is null whenever the bar is too narrow for words. Those buttons are icon-only at exactly the widths where the offset is most visible, so a plain truthiness check on `children` would have missed them.

## Why the clear button looks like it's inside the field

The control is a bordered box with `padding: 0 8px`, which left the button **9px short of the right edge and 2px short of the full height**. Semantic welded it to the field — measured on the QA Pi: zero gap from the input, zero to the outer edge, full 44px height.

It now stretches, and negative margins cancel the padding and the flex gap, with a divider on its left so "attached" doesn't become "indistinguishable". Negative margins rather than dropping the control's padding, because that padding is what holds the search icon and the text off the border — only this one child should escape it.

The `html` prefix on that rule is load-bearing: Mantine sets ActionIcon's height from `--ai-size` on its own single class, so at equal specificity the winner is whichever stylesheet loaded last, which is not something to rely on.

## Testing

Fourteen new cases, each fix planted separately — the icon change reddens three, the footer one, the clear button one.

Every claim is paired with an inverse, because all three fixes could over-apply:
- a **labelled** button must keep its icon on the left (centring must not apply when there is a label);
- the field must still be **padded** around its other children (the negative margins are aimed at one child);
- the footer's buttons must stay on **one row** and share a top edge, so "it's a flex row" is checked as behaviour rather than as a declaration.

One correction worth recording: the height assertion first compared against the control's **border** box and failed on correct behaviour. The button fills the *content* box; the 1px above and below is the field's own border, which it should sit inside rather than paint over. The test says which box it means and why.

1128 jest / 63 suites, 226 in `ui-layout.cy.js`, clean `npm run build`.

Confirmed live: footer gaps `8px` on one row, config glyphs at **0.0px** from centre, clear button 1px from the field edge at the field's full inner height.

### DIFF
--- a/app/src/components/ui/Button.tsx
+++ b/app/src/components/ui/Button.tsx
@@ -83,7 +83,20 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>((
      * too narrow for words.  Those buttons are icon-only at exactly the widths where the
      * offset is most visible.
      */
-    const labelled = React.Children.toArray(children).length > 0;
+    /*
+     * Blank text is not a label.  `toArray` drops null, undefined and booleans, but keeps an
+     * empty or whitespace-only string, which would leave the icon in `leftSection` with its
+     * margin and the glyph off-centre again.  No call site does that today; the predicate
+     * should not depend on that staying true.
+     *
+     * Two limits, both deliberate.  "Icon-only" means exactly one of `icon`/`iconAfter`: a
+     * label-less button with both would keep both section margins and stay mis-centred, and
+     * there are no such call sites to design for.  And any ELEMENT child counts as a label,
+     * including an empty fragment -- `toArray` does not look inside one, so `<></>` is
+     * indistinguishable from real content without recursing into fragment props.
+     */
+    const labelled = React.Children.toArray(children)
+        .some(child => typeof child !== 'string' || child.trim().length > 0);
     const soleIcon = !labelled && (icon || iconAfter) && !(icon && iconAfter)
         ? renderIcon(icon || iconAfter)
         : undefined;
--- a/app/src/components/ui/Button.tsx
+++ b/app/src/components/ui/Button.tsx
@@ -65,9 +65,29 @@ export const resolveSize = <T, >(size: T): T =>
     typeof size === 'string' ? ((semanticSizes[size] ?? size) as T) : size;
 
 export const Button = forwardRef<HTMLButtonElement, ButtonProps>((
-    {role, icon, iconAfter, className, ...props}, ref
+    {role, icon, iconAfter, className, children, ...props}, ref
 ) => {
     const fromRole = role ? roleProps[role] : undefined;
+
+    /*
+     * An icon with no label is the button's CONTENT, not a leading icon.
+     *
+     * Mantine gives `leftSection` a `margin-inline-end` to hold it off the label.  With no
+     * label that margin is pure offset: the glyph sits about 8px left of the button's centre,
+     * which is what the import and save buttons on the Settings config table looked like next
+     * to a correctly centred Restore -- Restore is an IconButton, which centres a lone glyph
+     * by construction.
+     *
+     * `React.Children.toArray` drops null, undefined and booleans, which matters because the
+     * file browser's footer passes `{label('Delete')}` and that is null whenever the bar is
+     * too narrow for words.  Those buttons are icon-only at exactly the widths where the
+     * offset is most visible.
+     */
+    const labelled = React.Children.toArray(children).length > 0;
+    const soleIcon = !labelled && (icon || iconAfter) && !(icon && iconAfter)
+        ? renderIcon(icon || iconAfter)
+        : undefined;
+
     return <MButton
         ref={ref}
         /*
@@ -81,11 +101,13 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>((
         color={props.color ?? fromRole?.color}
         variant={props.variant ?? fromRole?.variant}
         className={[fromRole?.className, className].filter(Boolean).join(' ') || undefined}
-        leftSection={renderIcon(icon)}
-        rightSection={renderIcon(iconAfter)}
+        leftSection={soleIcon ? undefined : renderIcon(icon)}
+        rightSection={soleIcon ? undefined : renderIcon(iconAfter)}
         {...props}
         size={resolveSize(props.size)}
-    />
+    >
+        {soleIcon ?? children}
+    </MButton>
 });
 Button.displayName = 'Button';
 
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -3105,6 +3105,56 @@ describe('a button with an icon and no label centres the icon', () => {
         });
     });
 
+    it('treats blank and empty children as no label', () => {
+        /*
+         * `Children.toArray` drops null, undefined and booleans but KEEPS an empty or
+         * whitespace-only string, so a length check alone would call these labelled and leave
+         * the glyph off-centre.
+         */
+        cy.mountUI(<div>
+            <Button icon='upload' aria-label='Empty' data-testid='empty'>{''}</Button>
+            <Button icon='upload' aria-label='Blank' data-testid='blank'>{'   '}</Button>
+        </div>);
+
+        cy.get('[data-testid="blank"]').should(() => {
+            ['empty', 'blank'].forEach((id) => {
+                const button = Cypress.$(`[data-testid="${id}"]`)[0];
+                const box = button.getBoundingClientRect();
+                const glyph = button.querySelector('svg').getBoundingClientRect();
+
+                expect((glyph.left + glyph.width / 2) - (box.left + box.width / 2),
+                    `${id} glyph centre`).to.be.closeTo(0, 1);
+            });
+        });
+    });
+
+    it('treats any element child as a label, including an empty fragment', () => {
+        /*
+         * Recording what actually happens rather than what would be tidy.  `toArray` does not
+         * look inside a fragment -- it returns the fragment itself as one child -- so `<></>`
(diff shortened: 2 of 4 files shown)


app/src/components/ui/ui.css
```diff
@@ -348,6 +348,32 @@ html[data-theme="amber"] .wrolpi-searchbox-control:focus-within {
...
+ * left margin cancels the flex gap as well, so the divider lands against the text the way
+ * the attached button did.
+ *
+ * `html` because Mantine sets ActionIcon's height from `--ai-size` on its own single class;
+ * at equal specificity the winner is whichever stylesheet loaded last, which is not a thing
+ * to rely on.
+ */
+html .wrolpi-searchbox-clear {
+    align-self: stretch;
+    height: auto;
```
lrnselfreliance, line 384: **[suggestion]** The clear button uses hardcoded `margin-left: -8px` and `margin-right: -8px` to cancel `.wrolpi-searchbox-control`'s `gap: 8px` and `padding: 0 8px` respectively. Those three values must stay in lockstep; changing only gap or only padding will either re-inset the clear control or push it past the field edge. The right-edge test tolerates 1.5px so small drift might pass while looking wrong.

**Suggestion:** Introduce a single token (e.g. `--wrolpi-searchbox-inset: 8px`) and use it for control padding, gap, and the clear button's negative margins so the weld cannot drift.

app/src/components/ui/ui.css
```diff
@@ -348,6 +348,32 @@ html[data-theme="amber"] .wrolpi-searchbox-control:focus-within {
...
+ * search icon and the text off the border, and only this one child should escape it.  The
+ * left margin cancels the flex gap as well, so the divider lands against the text the way
+ * the attached button did.
+ *
+ * `html` because Mantine sets ActionIcon's height from `--ai-size` on its own single class;
+ * at equal specificity the winner is whichever stylesheet loaded last, which is not a thing
+ * to rely on.
+ */
+html .wrolpi-searchbox-clear {
+    align-self: stretch;
```
lrnselfreliance, line 383: **[suggestion]** `html .wrolpi-searchbox-clear` correctly beats Mantine's single-class `height: var(--ai-size)` (specificity `(0,1,1)` vs `(0,1,0)`), matching the established `html .mantine-ActionIcon-root` pattern. But ActionIcon also sets `min-height`, `width`, and `min-width` to `--ai-size`, and those are not overridden. Stretch can grow the clear control above `--ai-size`; it cannot shrink below it, and width stays square-capped. Today that matches the input-height remap, so tests should pass, but the rule does not fully own "fill the field" sizing the way the comment claims.

**Suggestion:** If stretch-to-field is the source of truth, also set `min-height: unset` (keep width/min-width if a fixed hit target is desired). Optionally align or retire the older "clear vs input height" test so one geometry claim owns this pair.

app/src/components/ui/ui.css
```diff
@@ -348,6 +348,32 @@ html[data-theme="amber"] .wrolpi-searchbox-control:focus-within {
...
+    height: auto;
+    margin-left: -8px;
```
lrnselfreliance: **[suggestion]** `margin-left: -8px` collapses the flex gap before the clear button regardless of which sibling precedes it. When `loading` is true, the DOM order is input → loading → clear (`SearchBox.tsx`), and `SearchResultsInput` can pass both `loading` and `clearable`. The spinner then sits flush against the clear divider instead of keeping the 8px gap used elsewhere in the control.

**Suggestion:** Prefer a structural layout (e.g. group the clear outside the gapped flex row, or cancel only `padding-right` and position the divider without always eating the preceding gap) so loading and clear can coexist without welding together.

app/src/components/ui/Button.tsx
```diff
@@ -65,9 +65,29 @@ export const resolveSize = <T, >(size: T): T =>
...
+     * which is what the import and save buttons on the Settings config table looked like next
+     * to a correctly centred Restore -- Restore is an IconButton, which centres a lone glyph
+     * by construction.
+     *
+     * `React.Children.toArray` drops null, undefined and booleans, which matters because the
+     * file browser's footer passes `{label('Delete')}` and that is null whenever the bar is
+     * too narrow for words.  Those buttons are icon-only at exactly the widths where the
+     * offset is most visible.
+     */
+    const labelled = React.Children.toArray(children).length > 0;
```
lrnselfreliance: **[nit]** `React.Children.toArray(children).length > 0` drops `null` / `undefined` / booleans (correct for FileBrowser's `{label(...)}` → `null`, and covered by the null-label test), but keeps empty strings and whitespace-only text nodes. `children=""` or an accidental whitespace-only child is treated as labelled, so the icon stays in `leftSection` with Mantine's `margin-inline-end` and the original off-centre bug remains. `APIButton` masks this via `props.children || null`, but direct `Button` call sites do not.

**Suggestion:** Treat blank strings as unlabelled (filter `toArray` results, or normalize children before the length check), and add a Cypress case for `children={''}` if that path matters.

app/src/components/ui/ui.css
```diff
@@ -348,6 +348,32 @@ html[data-theme="amber"] .wrolpi-searchbox-control:focus-within {
...
+ * Negative margins rather than removing the control's padding: the padding is what holds the
+ * search icon and the text off the border, and only this one child should escape it.  The
+ * left margin cancels the flex gap as well, so the divider lands against the text the way
+ * the attached button did.
+ *
+ * `html` because Mantine sets ActionIcon's height from `--ai-size` on its own single class;
+ * at equal specificity the winner is whichever stylesheet loaded last, which is not a thing
+ * to rely on.
+ */
+html .wrolpi-searchbox-clear {
```
lrnselfreliance, line 382: **[nit]** The clear `ActionIcon` keeps default `border-radius` while being stretched and pulled flush to a square-cornered `.wrolpi-searchbox-control` (no radius on the control). Bounding-box tests can pass while the filled background leaves small panel-colored wedges at the top-right and bottom-right of the field, slightly undercutting the "welded control" look.

**Suggestion:** Set `border-radius: 0` on `.wrolpi-searchbox-clear` (or match a shared control radius if one is introduced later).

app/src/components/ui/ui-layout.cy.js
```diff
@@ -3044,3 +3044,200 @@ describe('a disabled button keeps its own colour', () => {
...
+            const glyph = $button[0].querySelector('svg').getBoundingClientRect();
+            const offset = (glyph.left + glyph.width / 2) - (box.left + box.width / 2);
+
+            expect(box.width, 'the button has width to be off-centre in')
+                .to.be.greaterThan(glyph.width + 8);
+            expect(offset, 'glyph centre vs button centre').to.be.closeTo(0, 1);
+        });
+    });
+
+    it('treats a null label as no label, which is how the file browser renders', () => {
```
lrnselfreliance, line 3073: **[suggestion]** Icon-only tests cover missing children, `{null}`, labelled inverse, and `iconAfter`-only — real geometry with centre offsets, and an inverse that protects labelled buttons. They do not cover fragments or empty string, which behave differently under `toArray` (fragments flatten; `''` remains). Multiple children and both-`icon`+`iconAfter` are also untested.

**Suggestion:** Add cases for `<></>` (expect centred), `<>Delete</>` (expect leading icon), and optionally `children={''}` and both-section icons, to lock the edge matrix called out in the change description.

app/src/components/ui/Button.tsx
```diff
@@ -65,9 +65,29 @@ export const resolveSize = <T, >(size: T): T =>
...
+    const soleIcon = !labelled && (icon || iconAfter) && !(icon && iconAfter)
```
lrnselfreliance, line 100: **[nit]** `!(icon && iconAfter)` deliberately skips the sole-icon path when both ends are set without a label. That avoids inventing content for a dual-section control, but leaves both section margins in place so a label-less two-icon button stays mis-centred. Unlikely in current call sites (none obvious), but still a sharp edge of the heuristic.

**Suggestion:** Leave as-is unless such call sites exist; if they might, render both icons as children without sections, or document that icon-only means exactly one of `icon` / `iconAfter`.

lrnselfreliance (commented):
## Summary

Three independent layout fixes, each aimed at a real measured defect, with geometry-based Cypress coverage and inverse constraints. The sticky-footer flex row, icon-as-content path for label-less `Button`s, and search clear negative-margin weld are all directionally correct; `React.Children.toArray` correctly handles the file-browser `null` label case that a truthiness check would miss. No blocking correctness bugs found. Remaining risks are coupling/fragility (hardcoded `-8px`), incomplete ActionIcon size overrides (`min-height` still from `--ai-size`), a loading+clearable gap side effect, and a few untested icon-only edge cases (empty string, fragments).

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 3

PR #626, opened 2026-08-02
ui: the map pin buttons were three sizes for three reasons

lrnselfreliance:
The pin row's three actions measured **46×30, 30×30 and 54×36**. Three separate causes.

## 1. APIButton was dropping `size` entirely

`useAPIButton` declares `size = 'medium'` as a parameter and then never puts it into the props it hands to `Button`. So **every** APIButton in the app rendered at Mantine's default regardless of its call site — **17 call sites pass a size**, from `xs` on the map pins to `big` and `huge` on Settings. The delete pin was 36px beside two 30px siblings because of it.

Passing the parameter default through would have fixed the bug by breaking everything else: every APIButton naming no size would have grown from `sm` to `md`. So the default is `undefined` now and Button's own default stands.

**Eight buttons on three other pages do get bigger**, which is their call sites finally being honoured — VideoPlayer's `large`, one of Channels' `big`, and Settings' two `big` and two `huge`. Say the word if any of those were written in the Semantic era and now read as too large; they're one-line changes at the call site.

## 2. An icon-only Button wasn't square

A Button is as wide as its content plus horizontal padding; an ActionIcon is square by construction. The pin row mixes the two, because `AddToPlaylistButton` renders an `IconButton` when it has no label. Their *heights* already agreed — ActionIcon's scale was remapped onto Button's some commits back — and this is the other axis of the same alignment, keyed off Mantine's own `--button-height` so the two stay in step through every size.

## 3. They were misaligned and touching

Bare inline-block siblings align on text **baselines**, and a Button's baseline is not an ActionIcon's, so the middle control sat 4.3px above the other two. A `Group`, as every other row of buttons in the app uses.

## Testing — written test-first

Seven cypress cases went red on the sizes: widths 46 against 30, 54 against 36, 62 against 42, 70 against 50, and `xs is shorter than lg: expected 36 to be below 32` for the dropped size. Both inverses stayed green throughout — a labelled button remains as wide as its label, and an APIButton naming no size keeps the height it had.

The alignment claim is measured in `ui-layout.cy.js`, **including the premise that bare siblings really are misaligned** — three controls of identical height failing to line up is not believable without it. The call sites are held by a source guard, since `MapPins` fetches from the API and builds rows through `SortableTable`.

**That guard earned itself twice.** Written for the pin rows, it found the map-files subscribe cell putting a preview button beside a subscribe button with the same defect. Then my first attempt at a general "any cell with two controls" rule had to be abandoned: a cell whose buttons are the two arms of a ternary renders only one and needs no Group, and excluding ternaries to allow for that *also* excluded the subscribe cell, whose `role={isSubscribed ? … : …}` says nothing about how many controls it holds. Named cells with a stated scope beat a general rule that quietly stops covering things.

Two slips of mine in that file, both caught by running it: `expect(value, message)` is chai and throws in jest, and `onSave(pin.id, …)` wasn't unique to the cell it named — the edit row's TextInput calls it from `onKeyDown` in a different cell, so the check passed against the wrong one.

1132 jest / 64 suites, 243 in `ui-layout.cy.js`, clean `npm run build`. Confirmed live: three **30×30** controls, vertical spread **0**, **10px** gaps.

### DIFF
--- a/app/src/components/Common.js
+++ b/app/src/components/Common.js
@@ -1709,7 +1709,19 @@ export function normalizeEstimate(estimate) {
 
 export function useAPIButton(
     color = 'violet',
-    size = 'medium',
+    /*
+     * Undefined, not 'medium'.  `size` was never put into `buttonArgs` below, so every
+     * APIButton in the app rendered at Mantine's default no matter what its call site asked
+     * for -- seventeen of them pass a size, from `xs` on the map's pin actions to `huge` on
+     * Settings, and all seventeen were ignored.  The map's pin row is where it showed: three
+     * buttons meant to be `xs`, one of them 36px tall next to two 30px ones.
+     *
+     * Defaulting to 'medium' and passing that through would have fixed the bug by breaking
+     * everything else -- every APIButton that names no size would have grown from `sm` to
+     * `md`.  Undefined lets Button's own default stand, so only the call sites that asked for
+     * something see a change.
+     */
+    size,
     floated,
     onClick,
     disabled,
@@ -1795,7 +1807,7 @@ export function useAPIButton(
     // `color` was a Semantic color name; a role carries the meaning instead, and a
     // caller that really wants a color can still pass one through `props`.
     const buttonArgs = {
-        onClick: localOnClick, disabled, loading, type,
+        onClick: localOnClick, disabled, loading, type, size,
         ...props,
     };
     if (id) {
--- a/app/src/components/Map.js
+++ b/app/src/components/Map.js
@@ -15,7 +15,7 @@ import {
     deleteMapFile, deleteMapPin, fetchMapSubscriptions, getMapFiles, getMapPins,
     mapSubscribe, mapUnsubscribe, rebuildMapSearchIndex, updateMapPin
 } from "../api";
-import {Button, Divider, Header, Icon, Modal, Table, TextInput} from "./ui";
+import {Button, Divider, Group, Header, Icon, Modal, Table, TextInput} from "./ui";
 import {SortableTable} from "./SortableTable";
 import {Media, StatusContext} from "../contexts/contexts";
 import {MAP_VIEWER_URI} from "./Vars";
@@ -229,19 +229,23 @@ function MapCatalogRow({item, subscribedRegions, fetchData}) {
         <Table.Cell>{name}</Table.Cell>
         <Table.Cell>{humanFileSize(size_estimate)}</Table.Cell>
         <Table.Cell>
-            {bbox && <Button
-                size='xs'
-                icon='eye'
-                onClick={() => setPreviewOpen(true)}
-            />}
-            <Button
-                role={isSubscribed ? 'danger' : undefined}
-                color={isSubscribed ? undefined : 'violet'}
-                size='xs'
-                onClick={handleButton}
-            >
-                {isSubscribed ? 'Unsubscribe' : 'Subscribe'}
-            </Button>
+            {/* See PinEditRow: a Group so the preview and subscribe controls line up and are
+                spaced, rather than sitting on baselines they do not share. */}
+            <Group gap='xs' wrap='nowrap'>
+                {bbox && <Button
+                    size='xs'
+                    icon='eye'
+                    onClick={() => setPreviewOpen(true)}
+                />}
+                <Button
+                    role={isSubscribed ? 'danger' : undefined}
+                    color={isSubscribed ? undefined : 'violet'}
+                    size='xs'
+                    onClick={handleButton}
+                >
+                    {isSubscribed ? 'Unsubscribe' : 'Subscribe'}
+                </Button>
+            </Group>
         </Table.Cell>
         {previewOpen && <RegionPreviewModal
             bbox={bbox}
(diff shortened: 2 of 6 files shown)


app/src/components/Map.js
```diff
@@ -524,21 +533,24 @@ function MapPins() {
...
-                    size='xs'
-                    role='danger'
-                    icon='trash'
-                    confirmContent='Delete this pin?'
-                    confirmButton='Delete'
-                    onClick={() => handleDelete(pin.id)}
-                />
+                {/* See PinEditRow: a Group so the three controls line up and are spaced. */}
+                <Group gap='xs' wrap='nowrap'>
+                    <Button size='xs' icon='edit' onClick={() => setEditingId(pin.id)}/>
```
lrnselfreliance: **[nit]** The pin action row’s icon-only `Button` (edit) and `APIButton` (delete) still have no accessible name (`aria-label` / equivalent). `AddToPlaylistButton` correctly uses `IconButton`’s required `label`, and the new Cypress fixtures carefully pass `aria-label` on the same pattern, but the production pin controls remain unnamed for assistive tech. Pre-existing, but this PR is exactly about those icon-only controls and rewrites the row.

**Suggestion:** Add `aria-label='Edit'` on the edit button and `aria-label='Delete'` (or rely on a shared convention) on the trash `APIButton`; same for the catalog preview eye button if it stays icon-only.

app/src/components/Common.js
```diff
@@ -1795,7 +1807,7 @@ export function useAPIButton(
     // `color` was a Semantic color name; a role carries the meaning instead, and a
     // caller that really wants a color can still pass one through `props`.
     const buttonArgs = {
-        onClick: localOnClick, disabled, loading, type,
+        onClick: localOnClick, disabled, loading, type, size,
```
lrnselfreliance, line 1810: **[suggestion]** Passing `size` is correct, and defaulting to `undefined` (not `'medium'`) is the right non-breaking choice. Call sites that already passed Semantic sizes will now finally render them: Settings `huge`→`xl` and `big`→`lg`, Channels form `big`→`lg`, VideoPlayer `large`→`lg`. Call sites with `size='small'` resolve to `sm` and match the old accidental default, so they should not move. No code bug, but this is the highest-risk visual change outside the map.

**Suggestion:** Manually eyeball Settings (Shutdown/Restart/Upgrade/Save), Channel create/edit Save, and VideoPlayer “Refresh” comments empty-state once; if any of those sizes were cargo-culted, adjust the call site rather than re-breaking passthrough.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -3358,3 +3359,178 @@ describe('the search field\'s clear button is attached to it', () => {
...
+    });
+});
+
+describe('APIButton honours the size its call site asks for', () => {
+    /*
+     * `useAPIButton` assembled the props it hands to Button and left `size` out, so it was
+     * dropped at every call site -- seventeen of them, from the map pins' `xs` to Settings'
+     * `huge`.  The map's delete pin came out 36px tall beside two 30px siblings.
+     */
+    it('renders a small APIButton smaller than a large one', () => {
```
lrnselfreliance, line 3429: **[nit]** APIButton size tests cover `xs` vs `lg`, equality with Button at `xs`, and the no-size default. They do not exercise the Semantic aliases that the comments cite (`huge`/`big`/`large`) through `APIButton`, nor icon-only geometry while `loading` is true (loader is `height/1.8`, so it should fit the square, but that is unasserted). `iconAfter`-only as sole content is covered for centering elsewhere, not for the new square rule.

**Suggestion:** Optional one-liner mounts: `APIButton size='huge'` taller than default; icon-only Button/APIButton with `loading` still square within 1px. Not blocking.

app/src/components/Map.test.js
```diff
@@ -0,0 +1,78 @@
...
+    rows.forEach(({name, marker}) => {
+        it(`groups the ${name} cell`, () => {
+            const cell = cells().find(candidate => candidate.includes(marker));
+
+            /*
+             * The premise and the claim in one object, because jest's `expect` takes no
+             * message argument -- that is chai, and passing one throws rather than annotating.
+             * A rename of the marker would otherwise turn this into a test of nothing.
+             */
+            expect({marker, found: cell !== undefined, grouped: !!cell?.includes('<Group')})
```
lrnselfreliance: **[nit]** Source guard correctly names cells by unique markers and asserts `cell.includes('<Group')`. That would still pass if a cell later gained an empty/self-closing Group while leaving bare button siblings (or a comment containing `<Group`). Acceptable given SortableTable/internal-row constraints and the Cypress geometry tests owning the real claim.

**Suggestion:** If the guard is ever extended, assert a single `<Group` that also contains the marker’s neighbouring control tags (e.g. `edit` + `AddToPlaylist` + `APIButton`), or keep as-is with the current documented scope.

lrnselfreliance (commented):
## Summary

This PR correctly fixes three independent causes of mismatched map pin action sizes: `useAPIButton` now passes `size` through (default left `undefined` so unnamed call sites stay at Button/`sm`), icon-only Buttons are squared via `data-icon-only` + `--button-height`, and multi-control map cells wrap in `Group`. Side effects on previously ignored non-default sizes (`huge`/`big`/`large`) are intentional and desirable; `size='small'` is a no-op via `resolveSize` so most detail-page APIButtons do not change. Tests are strong (geometry equality, inverses, premise-for-Group, source guards) with only minor a11y and coverage nits.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 3

PR #628, opened 2026-08-03
ui: size the five modals that had never been sized

lrnselfreliance:
These five carried no `size` prop at all, so they rendered at Mantine's 440px default — a width nobody had chosen, inherited from the library rather than decided.

| Modal | Size | Width |
|---|---|---|
| FileBrowser rename | `small` | 440 (unchanged) |
| FileBrowser move | `large` | **620** |
| FileBrowser ignore directory | `small` | 440 (unchanged) |
| FileBrowser make directory | `large` | **620** |
| Files filter | `small` | 440 (unchanged) |

The three unchanged ones are **written down** rather than left implicit, so the decision lives in the source and a later change to Mantine's default can't move them.

## One thing I asked about rather than guessed

You asked for `medium` on the ignore dialog, and there is no such modal size — the vocabulary is `mini/tiny/small/large/fullscreen`, and Mantine has no token between 440 and 620. An unknown name would have passed through untranslated, rendered at 440 anyway, and left the source claiming otherwise. That's precisely what the size guard in `ui.test.js` exists to catch, so the build would have failed either way; `small` was the answer.

## Verification

Measured in the browser rather than read off the props: **move 620, make directory 620, ignore 440, rename 440**. Move and ignore are disabled until something is selected, so a directory was selected first.

1134 jest / 64 suites, clean `npm run build`.

**Note on the test run:** the full jest suite currently needs `--maxWorkers=2` on my machine. At load average 38 (a VM and a game between them using ~400% CPU), four unrelated suites — `ControllerPage`, `CollectionReorganizeModal`, `CollectionTagModal`, `Flasher` — time out and the run takes 183s instead of 5s. They pass in isolation and with reduced parallelism, and nothing in this change touches them. Worth knowing if CI ever runs on a loaded box.

### DIFF
--- a/app/src/components/FileBrowser.js
+++ b/app/src/components/FileBrowser.js
@@ -711,6 +711,11 @@ export function MoveModal({open, paths, onClose, onSubmit}) {
         {destination ? fileCell(destination + '/' + pathName(i)) : fileCell(pathName(i))}
     </Table.Row>);
 
+    /*
+     * Audited: `large` (620px).  This dialog holds a DirectorySearch and a two-column
+     * Source/Destination table of every path being moved, which is the densest content
+     * of the five FileBrowser modals.
+     */
     return <Modal size='large'
                   onClose={onClose}
                   open={open}
@@ -814,6 +819,18 @@ export function MakeDirectoryModal({open, onClose, parent, onSubmit}) {
         }
     }
 
+    /*
+     * Audited: `large` (620px), deliberately wider than RenameModal's `small` even though
+     * the form is the same shape -- one field and a button.  The difference is the path
+     * above it: rename shows a RELATIVE path, this shows the absolute one including the
+     * media directory, and a bare `<pre>` is `white-space: pre`, so it scrolls rather
+     * than wraps.  Measured in the browser: 440px holds 42 characters and 620px holds
+     * 61.  `/media/wrolpi/videos/Practical Engineering/Season 2024/` is 55, so an
+     * ordinary channel-and-season path scrolls out of sight at 440 and fits at 620.
+     *
+     * Written out because it reads like an inconsistency next to rename, and the next
+     * person to notice that should see the measurement before narrowing it back.
+     */
     return <Modal size='large' open={open} onClose={onClose}>
         <Modal.Header>Make Directory</Modal.Header>
         <Modal.Content>
--- a/app/src/components/FileBrowser.js
+++ b/app/src/components/FileBrowser.js
@@ -643,7 +643,8 @@ export function RenameModal({open, onClose, path, onSubmit, onPending}) {
         setValue(pathName(path));
     }, [path]);
 
-    return <Modal
+    // Audited: unchanged at `small` (440px), written down so the default cannot move it.
+    return <Modal size='small'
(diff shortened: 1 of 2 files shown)


app/src/components/FileBrowser.js
```diff
@@ -812,7 +814,7 @@ export function MakeDirectoryModal({open, onClose, parent, onSubmit}) {
         }
     }
 
-    return <Modal open={open} onClose={onClose}>
+    return <Modal size='large' open={open} onClose={onClose}>
```
lrnselfreliance, line 834: **[suggestion]** `MakeDirectoryModal` is set to `size='large'` (620px), which **widens** it relative to the previous implicit default (440px). Its body is only a path `<pre>` plus a single `TextInput` and a Create button — the same shape as `RenameModal` (`size='small'`) and similar one-field dialogs elsewhere (e.g. inventory rename uses `tiny`). `MoveModal` genuinely needs `large` (directory search + Source/Destination table); make-directory does not show comparable density. PR description says the audit chose 620 deliberately and measured it, so this is not a vocabulary or guard bug — it is a content-fit inconsistency that leaves a sparse dialog almost as wide as upload/bulk-tag.

**Suggestion:** Prefer `size='small'` to match `RenameModal` and keep make-directory at the prior 440px unless long media paths were observed wrapping badly in-browser. If `large` stays, add a short audit comment (as on rename/ignore/filter) stating why this simple form is intentionally wider than rename, so a later pass does not “fix” it back.

app/src/components/FileBrowser.js
```diff
@@ -710,7 +711,7 @@ export function MoveModal({open, paths, onClose, onSubmit}) {
         {destination ? fileCell(destination + '/' + pathName(i)) : fileCell(pathName(i))}
     </Table.Row>);
 
-    return <Modal
+    return <Modal size='large'
```
lrnselfreliance, line 719: **[nit]** `MoveModal` and `MakeDirectoryModal` gain sizes without the “Audited: …” comments used on the three unchanged-at-440 call sites. Not required for correctness, but the PR’s own rationale (decisions live in source so defaults cannot drift) applies equally to intentional *changes* of width.

**Suggestion:** Optionally add one-line comments on the `large` call sites (e.g. move: table + DirectorySearch; make directory: intentional widen / or small if Issue 1 is taken) so the audit trail is uniform across all five.

lrnselfreliance (commented):
## Summary

PR #628 correctly pins five previously unsized FileBrowser/Files modals to the audited Semantic vocabulary (`mini`/`tiny`/`small`/`large`/`fullscreen`), not raw Mantine tokens. The three “write down the default” cases (`RenameModal`, `IgnoreDirectoryModal`, `SearchFilter`) stay at `small` (440px); `MoveModal` correctly goes to `large` for DirectorySearch + a two-column path table. No accidental non-size logic changes, and all five sizes pass the `ui.test.js` known-name guard. Completeness within FileBrowser/Files is good (upload + SearchFilterModal were already sized). The only content-fit concern is `MakeDirectoryModal` at `large`.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

PR #630, opened 2026-08-03
ui: scale the whole interface by 10%, off the root font-size

lrnselfreliance:
The interface reads about 10% smaller than the pre-migration Semantic build. I measured both at 1280px rather than guessing, and the drift is **not one number**:

| | ours | Semantic |
|---|---|---|
| h1 / h2 / h3 / h4 / h5 | 26 / 21 / 17 / 14 / 13 | **28 / 24 / 18 / 15 / 14** |
| video card | 238px wide, 13px title | **290px wide, 18px title** |
| body text | **16px** | 14px |
| button height | **42px** | 36px |

So the headings and card text are smaller while the body text and chrome are *larger* — which is why no per-component nudge would have worked. What a user compares is the whole page at once.

**Ten percent is not a fudge factor.** Our heading scale × 1.1 is 28.6 / 23.1 / 18.7 / 15.4 / 14.3 — Semantic's scale to within a pixel at every step. The browser now renders exactly those numbers.

### The mechanism

`--ui-scale: 1.1` with `html { font-size: calc(100% * var(--ui-scale)) }`, and everything that should grow written in `rem`:

- **102 lengths** converted in `ui.css` / `App.css` / `tokens.css`, restricted by property: font-size, padding, margin, gap, width/height, border-radius, offsets.
- **Borders, outlines and shadows deliberately not converted.** A hairline is a hairline at any text size.
- **46 inline `fontSize`** px values across 8 files. `em` values left alone — they already resolve against the parent.
- **Ten inline-styled blocks in the library** (Card's text block, Loading, Placeholder, ColoredInput, Confirm's action row, CardGroup's gap) moved into `ui.css`: an inline px size can be reached by neither a scale nor a test.
- **`CardGroup` converts its `minWidth` prop to rem**, which is what returns a phone to one card per row — as px the track ignored the scale, so a 430px screen fitted two 200px cards where the old build fitted one 290px card.

Mantine needs nothing: its 644 size declarations are `calc(Xrem * var(--mantine-scale))` and scale with the root. `theme.scale` must stay at **1** — it exists to *cancel* a customised root font-size, so setting it to 1.1 as well would leave Mantine's half of the interface at the old size beside our larger text. A test pins that.

### Why not `zoom`

I built `zoom: 1.1` on the root first, and it does work in one line — but you were right to push back, and measuring found two concrete reasons beyond taste:

1. It scales at **paint**, so every 1px hairline becomes 1.1px and softens on a 1× display.
2. Viewport units don't shrink to compensate, so `100vh` paints **110% of the screen**. A `95vw` modal overflowed sideways and the CBZ viewer's fullscreen overflowed down; the map rendered at 93.5% of the screen where it asked for 85%.

Rem scales real lengths, keeps borders crisp, leaves viewport units alone, and preserves a user's own browser font-size preference instead of overriding it — hence a percentage rather than a px root size.

### Tests

Five guards in `themes/ui-scale.test.js`, each planted against: the scale itself; every scalable declaration is rem; borders are **not**; no inline px `fontSize` survives anywhere including the gallery; Mantine's scale stays 1.

Two existing `ui.test.js` assertions read the inline styles I moved into CSS. One (`margin-top: auto` on the card foot) was asserting *where the declaration was written* rather than what it did, and `ui-layout.cy.js:1820` already measures the real foot alignment — so it's gone, with a note. The other now asserts the rem track width, which is the meaningful claim.

**A mistake worth recording:** a malformed plant command of mine (`s/font-size: 0.8125rem;\n//` in perl line mode) silently deleted **nine** `font-size` declarations from `ui.css` rather than one. I caught it by counting declarations before and after, then reverted and re-ran the scripted conversion clean. Every remaining difference is accounted for: +30 in `ui.css` are the ten new rules, −1 in `App.css` is a dead Semantic rule, +2 in `tokens.css` are the scale and root size.

Also removes `.ui.modal.themed-confirm` — dead Semantic-era rule (`.ui.modal` no longer exists, nothing sets the class) and the source of the one `95vw`.

### Verification

- 1141 tests / 65 suites pass; clean `npm run build`
- Browser-checked at 1280px and at a 430px phone (one card per row, as the old build had)
- Navbar stays a single row from 780px to 1800px, overflow nav collapsing links progressively
- All four themes: root 17.6px, hairlines still 1px, no horizontal overflow, every new class resolves

### DIFF
--- a/app/cypress/e2e/keyboard-shortcuts.cy.js
+++ b/app/cypress/e2e/keyboard-shortcuts.cy.js
@@ -17,17 +17,41 @@ describe('Keyboard Shortcuts', () => {
             body: {settings: {}}
         }).as('getSettings');
 
-        // Mock tags API
-        cy.intercept('GET', '/api/tags', {
+        /*
+         * `/api/tag` and `/api/events/feed`, which are the URLs the app actually requests.
+         *
+         * These read `/api/tags` and `/api/events` before -- neither of which anything calls,
+         * so both mocks matched nothing and the real calls fell through.  In CI, where the
+         * React app is served with no API behind it, they failed and api.js raised an
+         * "Unexpected server response" toast, which then covered the search modal's input and
+         * failed an occlusion-sensitive assertion.  Locally the API answered and the spec
+         * passed, which is why it only ever broke on CI.
+         */
+        cy.intercept('GET', '/api/tag*', {
             statusCode: 200,
             body: {tags: []}
         }).as('getTags');
 
-        // Mock events API
-        cy.intercept('GET', '/api/events', {
+        cy.intercept('GET', '/api/events/feed*', {
             statusCode: 200,
-            body: {events: []}
+            // `now` as well as `events`: Events.js reads both off the response.
+            body: {events: [], now: new Date(0).toISOString()}
         }).as('getEvents');
+
+        /*
+         * The other two the home page calls.  `file_groups` is reached as
+         * `data['totals']['file_groups']`, so an empty object here would throw rather than
+         * toast -- the shape has to be right, which is why a blanket `{}` stub does not work.
+         */
+        cy.intercept('POST', '/api/files/search', {
+            statusCode: 200,
+            body: {file_groups: [], totals: {file_groups: 0}}
+        }).as('searchFiles');
+
+        cy.intercept('GET', '/api/files/worker_status', {
+            statusCode: 200,
+            body: {status: {}}
+        }).as('getWorkerStatus');
 
         // Mock search suggestions API
         cy.intercept('POST', '/api/search/suggestions', {
@@ -56,25 +80,13 @@ describe('Keyboard Shortcuts', () => {
             cy.get('[role="dialog"]').should('be.visible');
 
             /*
-             * Take down any toast before the assertion below, which is occlusion-sensitive.
-             *
-             * Toasts paint above modals on purpose -- z-index 400 against the modal stack's
-             * 200 -- so that an error raised while a modal is open stays visible.  The search
-             * modal is tall, so it sits near the top of the window, and a top-right toast
-             * covers the right-hand 192px of its 365px input.  Cypress occlusion-checks a
-             * `position: fixed` element, so `should('be.visible')` on that input fails.
-             *
-             * That is not this spec's subject.  It passed locally and failed in CI, where the
-             * React app is served with no API behind it: `/api/tag`, `/api/events/feed`,
-             * `/api/files/search` and `/api/files/worker_status` are not mocked here, they
-             * fail, and api.js raises "Unexpected server response".  The overlap itself is an
-             * accepted trade-off -- see the toast/modal note in ui.css -- so the environment is
-             * made quiet rather than the assertion weakened.
-             *
-             * Removed rather than dismissed by clicking: a click races the next failing call.
+             * Occlusion-sensitive: Cypress checks whether a `position: fixed` element is
+             * covered, and a toast at top-right covers the right-hand 192px of this 365px
+             * input.  Toasts paint above modals on purpose -- z-index 400 against the modal
+             * stack's 200 -- so an error raised while a modal is open stays visible, and that
+             * overlap is an accepted trade-off rather than something this spec should assert
+             * about.  The endpoints are all mocked in `beforeEach` so no toast is raised.
              */
-            cy.get('body').then(($body) => $body.find('.mantine-Notifications-root').remove());
-
             cy.get('[role="dialog"] input').should('be.visible');
         });
 
--- a/app/cypress/e2e/keyboard-shortcuts.cy.js
+++ b/app/cypress/e2e/keyboard-shortcuts.cy.js
@@ -54,6 +54,27 @@ describe('Keyboard Shortcuts', () => {
 
             // Search modal should open
             cy.get('[role="dialog"]').should('be.visible');
+
+            /*
+             * Take down any toast before the assertion below, which is occlusion-sensitive.
+             *
+             * Toasts paint above modals on purpose -- z-index 400 against the modal stack's
+             * 200 -- so that an error raised while a modal is open stays visible.  The search
+             * modal is tall, so it sits near the top of the window, and a top-right toast
+             * covers the right-hand 192px of its 365px input.  Cypress occlusion-checks a
+             * `position: fixed` element, so `should('be.visible')` on that input fails.
+             *
+             * That is not this spec's subject.  It passed locally and failed in CI, where the
+             * React app is served with no API behind it: `/api/tag`, `/api/events/feed`,
+             * `/api/files/search` and `/api/files/worker_status` are not mocked here, they
+             * fail, and api.js raises "Unexpected server response".  The overlap itself is an
+             * accepted trade-off -- see the toast/modal note in ui.css -- so the environment is
+             * made quiet rather than the assertion weakened.
+             *
+             * Removed rather than dismissed by clicking: a click races the next failing call.
+             */
+            cy.get('body').then(($body) => $body.find('.mantine-Notifications-root').remove());
+
             cy.get('[role="dialog"] input').should('be.visible');
         });
 
--- a/app/src/App.css
+++ b/app/src/App.css
@@ -36,7 +36,7 @@
  * switch to `bottom-right` should not silently inherit a top offset.
  */
 html .mantine-Notifications-root[data-position^='top'] {
-    top: calc(var(--navbar-height) + var(--mantine-spacing-md, 16px));
+    top: calc(var(--navbar-height) + var(--mantine-spacing-md, 1rem));
 }
 
 .wrolpi-navbar-right {
@@ -159,7 +159,7 @@ html .wrolpi-navbar .mantine-ActionIcon-root:hover {
 }
 
 .channel-edit {
-    border-radius: 3px;
+    border-radius: 0.1875rem;
     padding: 0.75em;
 }
 
@@ -186,12 +186,12 @@ pre.wrap-text {
 
 .duration-overlay {
     position: absolute;
-    top: 4px;
-    right: 4px;
+    top: 0.25rem;
+    right: 0.25rem;
     color: black;
     background: #ffffff;
-    padding: 0px 4px;
-    border-radius: 3px;
+    padding: 0px 0.25rem;
+    border-radius: 0.1875rem;
 }
 
 .no-link-underscore {
@@ -219,8 +219,8 @@ div.inverted {
 .toggle {
     position: relative;
     display: inline-block;
-    width: 49px;
-    height: 23px;
+    width: 3.0625rem;
+    height: 1.4375rem;
 }
 
 .toggle input {
@@ -239,14 +239,14 @@ div.inverted {
     background-color: rgb(166, 166, 166);
     -webkit-transition: .4s;
     transition: .4s;
-    border-radius: 34px;
+    border-radius: 2.125rem;
 }
 
 .slider:before {
     position: absolute;
     content: "";
-    height: 22px;
-    width: 22px;
+    height: 1.375rem;
+    width: 1.375rem;
     left: 0;
     bottom: 0;
     background-color: white;
@@ -316,9 +316,9 @@ input:checked + .slider:before {
    Use 100dvh so mobile browser chrome (URL bar) is accounted for. */
 .preview-fit {
     width: 100%;
-    height: calc(100dvh - 300px);
-    max-height: calc(100dvh - 300px);
-    min-height: 240px;
+    height: calc(100dvh - 18.75rem);
+    max-height: calc(100dvh - 18.75rem);
+    min-height: 15rem;
     overflow: hidden;
     position: relative;
     display: flex;
@@ -356,12 +356,12 @@ td.file-path > svg {
 .ui-alerts {
     position: fixed;
     z-index: 2060;
-    padding: 24px;
+    padding: 1.5rem;
 }
 
 .ui-alerts.top-right {
-    top: 20px;
-    right: 20px;
+    top: 1.25rem;
+    right: 1.25rem;
 }
 
 textarea.otp {
@@ -370,9 +370,9 @@ textarea.otp {
 }
 
 .card-icon {
-    min-height: 163px;
+    min-height: 10.1875rem;
     text-align: center;
-    padding-top: 53px;
+    padding-top: 3.3125rem;
 }
 
 .card-icon.inverted {
@@ -426,11 +426,6 @@ textarea.otp {
     box-shadow: none !important;
 }
 
-/* Mobile-responsive Confirm modal */
-.ui.modal.themed-confirm {
-    max-width: 95vw;
-}
-
 .clickable {
     cursor: pointer;
 }
@@ -469,7 +464,7 @@ textarea.otp {
 }
 
 .search-clear {
-    margin: 10px;
+    margin: 0.625rem;
 }
 
 .icon.divider {
@@ -524,40 +519,40 @@ table.ratio-table td.min-width {
 }
 
 .form-skeleton-label {
-    height: 14px;
-    width: 80px;
-    margin-bottom: 8px;
+    height: 0.875rem;
+    width: 5rem;
+    margin-bottom: 0.5rem;
     background: linear-gradient(90deg, #e0e0e0 25%, #f0f0f0 50%, #e0e0e0 75%);
     background-size: 200% 100%;
     animation: skeleton-shimmer 1.5s infinite;
-    border-radius: 3px;
+    border-radius: 0.1875rem;
 }
 
 .form-skeleton-input {
-    height: 38px;
+    height: 2.375rem;
     width: 100%;
     background: linear-gradient(90deg, #e0e0e0 25%, #f0f0f0 50%, #e0e0e0 75%);
     background-size: 200% 100%;
     animation: skeleton-shimmer 1.5s infinite;
-    border-radius: 4px;
+    border-radius: 0.25rem;
 }
 
 .form-skeleton-textarea {
-    height: 80px;
+    height: 5rem;
     width: 100%;
     background: linear-gradient(90deg, #e0e0e0 25%, #f0f0f0 50%, #e0e0e0 75%);
     background-size: 200% 100%;
     animation: skeleton-shimmer 1.5s infinite;
-    border-radius: 4px;
+    border-radius: 0.25rem;
 }
 
 .form-skeleton-button {
-    height: 42px;
-    width: 100px;
+    height: 2.625rem;
+    width: 6.25rem;
     background: linear-gradient(90deg, #e0e0e0 25%, #f0f0f0 50%, #e0e0e0 75%);
     background-size: 200% 100%;
     animation: skeleton-shimmer 1.5s infinite;
-    border-radius: 4px;
+    border-radius: 0.25rem;
 }
 
 @keyframes skeleton-shimmer {
@@ -741,7 +736,7 @@ tr.drag-deselecting td {
     .inventory-print th,
     .inventory-print td {
         border: 1px solid #999;
-        padding: 4px 6px;
+        padding: 0.25rem 0.375rem;
         text-align: left;
         vertical-align: top;
     }
--- a/app/src/App.css
+++ b/app/src/App.css
@@ -216,66 +216,18 @@ div.inverted {
     background-color: #1B1C1D;
 }
 
-.toggle {
-    position: relative;
-    display: inline-block;
-    width: 3.0625rem;
-    height: 1.4375rem;
-}
-
-.toggle input {
-    opacity: 0;
-    width: 0;
-    height: 0;
-}
-
-.slider {
-    position: absolute;
-    cursor: pointer;
-    top: 0;
-    left: 0;
-    right: 0;
-    bottom: 0;
-    background-color: rgb(166, 166, 166);
-    -webkit-transition: .4s;
-    transition: .4s;
-    border-radius: 2.125rem;
-}
-
-.slider:before {
-    position: absolute;
-    content: "";
-    height: 1.375rem;
-    width: 1.375rem;
-    left: 0;
-    bottom: 0;
-    background-color: white;
-    -webkit-transition: .4s;
-    transition: .4s;
-    border-radius: 50%;
-}
-
-.slider.disabled:before {
-    background-color: #CCCCCC;
-}
-
-input:checked + .slider {
-    background-color: #6435C9FF;
-}
-
-input.inverted:checked + .slider.inverted {
-    background-color: #6435C9AA;
-}
-
-input.disabled + .slider {
-    background-color: rgb(147, 147, 147);
-}
-
-input:checked + .slider:before {
-    -webkit-transform: translateX(26px);
-    -ms-transform: translateX(26px);
-    transform: translateX(26px);
-}
+/*
+ * The hand-rolled `.toggle` / `.slider` switch is gone.
+ *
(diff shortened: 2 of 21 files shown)


lrnselfreliance:
### Follow-up: landscape posters were letterboxing on one-column cards

Answering "below 545px the video card thumbnail has margin space on the left and right": the cause is `--card-poster-max-height`, and running cypress for the first time on this branch turned up three tests of mine that were quietly asserting the scale is 1.

**The gap.** The cap is a fixed height, so it stops scaling with the card. Held to 220px, a 16:9 thumbnail is 391px wide — so any card wider than that centred it with empty panel on both sides. At one column the card is the full width of the screen, so that is every video thumbnail on a phone.

It pre-dated the scale (it showed below ~414px), but the scale widened the range: the track minimum went 200px → 220px, so one column now starts at 456px of content width instead of 414. That is the ~545px window you saw, less the scrollbar and whatever sits beside the page.

The cap is now `max(var(--card-poster-max-height), 56.25cqw)`, with the poster band as the query container. `56.25cqw` is nine sixteenths of the card — exactly the height of a 16:9 image that fills the width — and the same number states the rule for every other ratio: **a poster may be as tall as a 16:9 slice of this card and no taller.** So a portrait book cover still reads as a band rather than a column, and now grows with the card instead of stopping at 220px. `max()` and not `min()`, so the floor wins on a narrow card.

Measured in the app at 400 / 451 / 500 / 700 / 1200px of grid width: no side gap at any of them, ratios preserved.

### I should have run cypress before opening this PR

AGENTS.md asks for it on frontend changes and I ran only jest and the build. Cypress found **three tests with a hardcoded px expectation**, which is a hidden assertion that `--ui-scale` is 1. All three failed at 1.1, in test names that say nothing about size:

- `cardPosterCap()` read the cap with `parseFloat`. It is rem now, so it returned `12.5`, failed a `within(80, 240)` bound, and took **five** poster tests down with it.
- `inNarrowColumn` was a fixed 340px, chosen to hold two 150px statistic tracks. At 165px tracks it became one column — so a test about the gap *between* columns had no second column to measure. It is now `340 × scale`, because "narrow" means narrow relative to the type.
- The APIButton size test asserted `36px` for `small`. It now measures against a real `sm` button, which is the claim it was making (`small` resolves to `sm`) and holds at any scale.

A `uiScale()` helper reads the token so no further test has to name a pixel count.

Also worth noting: **every existing poster case mounted a 233px card**, which is why the wide card's own failure went unseen for the entire migration. Two cases added on a 451px card.

### Verification

- `ui-layout.cy.js`: **248 passing, 0 failing** (was 246 + 2 new)
- 1141 jest tests / 65 suites; clean `npm run build`
- The 12 remaining component failures are the pre-existing Semantic-era selector failures in `Tags`, `Channels`, `Download` and `Inventory` — same specs, same count, untouched by this

lrnselfreliance:
### Triage: no review comments, but two real CI failures — both fixed

No reviews or inline comments on this PR. `cypress-e2e-tests` was red, though, and it was **not** a flake: the job passed on the base commit (`05b1c753`) and failed on the scale commit (`556f5d6e`). Two independent causes.

#### 1. The suggestion list outgrew its cap — my regression

`directory-search.cy.js`: `expected '<div.wrolpi-searchbox-result-title>' to be 'visible'`.

The list is capped at `min(60vh, 30rem)`. The rem ceiling scales with the interface; **a fraction of the viewport does not.** So the rows grew 10% and the cap didn't: the full set of grouped suggestions now needs ~420px, and 60vh of Cypress's 660px-tall default viewport is 396px. The last group was clipped — the exact defect the comment on that rule already describes.

Measured rather than guessed: the spec fails at 660px of viewport height and passes from 700px up, so the shortfall is ~24px. Raised to `70vh` (462px there); the `30rem` ceiling still bounds it on a tall window, so it cannot cover the page either way.

#### 2. An error toast covers the search input — pre-existing, worsened by the scale

`keyboard-shortcuts.cy.js`: `expected '<input.wrolpi-searchbox-input>' to be 'visible'`, occluded by a danger toast.

Toasts paint above modals deliberately (z-index 400 against the modal stack's 200) so an error raised while a modal is open stays visible. The search modal is tall enough to sit near the top of the window, so a top-right toast lands on its input. Measured at both scales:

| | input width | toast covers |
|---|---|---|
| scale 1 | 330px | 130px (**39%**) |
| scale 1.1 | 365px | 192px (**53%**) |

So this predates the scale; the scale widened it enough to push CI over the line. Roland's call was to accept the overlap — an error toast belongs on top — and fix the test.

It passed locally and failed in CI because **CI serves the React app with no API behind it.** The endpoints this spec leaves unmocked are `/api/tag`, `/api/events/feed`, `/api/files/search` and `/api/files/worker_status` — note the spec's existing mocks say `/api/tags` and `/api/events`, *neither of which the app calls*. They fail, and api.js raises "Unexpected server response".

The toast container is now removed before the occlusion-sensitive assertion, rather than clicked away, because a click races the next failing call. A blanket `/api/**` 200 stub was tried first and does **not** work — a 200 with the wrong shape fails identically, which is what "Unexpected server response" means.

#### Verification

Reproduced CI's condition locally with `forceNetworkError` on every unmocked endpoint:

- **12 passing** with the toast removal; **11 passing / 1 failing** without it (planted)
- `directory-search.cy.js`: 9 passing at the default 660px height, was 8/1
- **All 50 e2e specs pass** under normal conditions
- 248 `ui-layout` component tests, 1141 jest tests / 65 suites, clean build

app/src/components/MapViewer.js
```diff
@@ -273,7 +273,7 @@ function LayerControl({map, scaleUnit, onScaleUnitChange, visibilityRef}) {
         maxHeight: "calc(100vh - 120px)",
```
lrnselfreliance: **[suggestion]** Map UI chrome was only partially migrated. Layer panel still uses absolute positioning and sizes in React px numbers (`top: 10`, `left: 10`, `minWidth: 200`, `width`/`height: 40`, `padding: "8px 12px"`, `maxHeight: "calc(100vh - 120px)"`); AddPinDialog keeps `padding: 16`, `minWidth: 250`; search panel keeps `width: 320`, `padding: "8px 12px"`, `maxHeight: 300`. Only `fontSize` was switched to rem. At scale 1.1 the map overlays' type grows while chrome geometry and reserved viewport chrome (`120px`) do not, so density and alignment drift relative to the rest of the app.

**Suggestion:** Move overlay layout into CSS classes with rem (and rem in the `100vh - …` reserve), same pattern as Card / Loading. Leave true map-coordinate values (context menu `e.point`, ScaleControl `maxWidth`) in px.

app/src/components/ui/ui.css
```diff
@@ -788,8 +837,8 @@ html[data-theme="amber"] .wrolpi-path-input-control:focus-within {
     position: absolute;
     top: 50%;
     left: -7px;
```
lrnselfreliance: **[suggestion]** Tag geometry was half-converted. `.wrolpi-tag::before` width/height are rem, but the eyelet stays `left: -7px` and `margin-top: -2.5px`. At scale 1.1 the diamond grows and the hole does not track its centre, so the point/eyelet alignment drifts. Same class of issue as other mixed-unit absolute positioning.

**Suggestion:** Convert eyelet offsets to rem (`left: -0.4375rem`, `margin-top: -0.15625rem`) or centre with `transform` percentages so both pieces share one coordinate system.

app/src/themes/ui-scale.test.js
```diff
@@ -0,0 +1,169 @@
...
+        /*
+         * A px font-size does not grow with the root, so it would shrink RELATIVE to
+         * everything around it -- the failure is a 13px caption beside 14.3px prose, which
+         * reads as sloppy rather than broken and so never gets reported.
+         *
+         * Borders are exempt and must stay px: a hairline is a hairline at any text size, and
+         * scaling it is exactly what made `zoom` the wrong tool.  Same for outlines, which are
+         * hairlines wearing a different name.
+         */
+        const SCALABLE = /^\s*(font-size|padding|padding-\w+|margin|margin-\w+|gap|row-gap|column-gap|width|min-width|max-width|height|min-height|max-height|border-radius|top|right|bottom|left)\s*:/;
```
lrnselfreliance: **[suggestion]** The scalable-length guard is narrower than its name implies and can give false confidence:
1. `SCALABLE` does not include `transform` / `flex` / `flex-basis` / `translate`, so `transform: translateX(26px)` and `flex: 1 1 200px` never fail the test.
2. The px matcher intentionally skips any length whose digits are preceded by `-`, so `left: -7px`, `right: -3px`, `margin-top: -2.5px` are invisible even on listed properties.
3. Only four stylesheets are scanned; production layout still lives in many inline styles, and only `fontSize` is checked there.
4. There is no runtime assertion that computed sizes actually change when `--ui-scale` changes.

**Suggestion:** Extend the property list (at least `transform`), match optional leading `-` on lengths, optionally scan JS/TS for numeric style lengths on the same property set, and add one jsdom/Cypress check that a known rem length tracks `fontSize` on `documentElement`.

app/src/App.css
```diff
@@ -239,14 +239,14 @@ div.inverted {
...
 .slider:before {
     position: absolute;
     content: "";
-    height: 22px;
-    width: 22px;
+    height: 1.375rem;
+    width: 1.375rem;
     left: 0;
     bottom: 0;
     background-color: white;
```
lrnselfreliance: **[nit]** `.toggle` / `.slider` track and knob were converted to rem, but the checked-state travel remains `transform: translateX(26px)`. At root 17.6px the track is ~53.9px and the knob ~24.2px, so the thumb should travel ~30px and stops short at 26px. No React call site currently uses these classes (Toggle is Mantine Switch), so this is dead CSS—but the partial conversion is exactly the mixed-unit failure mode the scale work is trying to prevent, and the new test does not catch it.

**Suggestion:** Either finish the conversion (`translateX(1.625rem)` / calc from width − knob) or delete the unused toggle CSS so it cannot be revived broken.

app/src/components/ui/ui.css
```diff
@@ -1265,15 +1362,32 @@ html[data-theme="amber"] .wrolpi-progress-fill {
...
+     *
+     * `56.25cqw` is nine sixteenths of the band: exactly the height of a 16:9 image that fills
+     * the width, and the same number states the rule for every other ratio -- a poster may be
+     * as tall as a 16:9 slice of this card and no taller.  So a book cover still reads as a
+     * band rather than a column, and now grows with the card instead of stopping at 220px.
+     *
+     * `max()`, not `min()`: on a narrow card the fraction is the smaller value and the floor
+     * has to win, or the cap would tighten as cards got smaller and crop a portrait cover.
+     */
+    max-height: max(var(--card-poster-max-height), 56.25cqw);
```
lrnselfreliance, line 1395: **[suggestion]** `max-height: max(var(--card-poster-max-height), 56.25cqw)` depends on container query units. If `cqw` is unsupported, the entire `max-height` declaration becomes invalid and the poster loses its cap (portrait covers can dominate the card). Pre-change `max-height: var(--card-poster-max-height)` degraded safely. The new layout behaviour on modern engines is good and well covered by Cypress; the risk is older WebViews only.

**Suggestion:** Keep a simple rem cap as a preceding declaration, then override with the `max(…, cqw)` rule so unsupported engines still clamp height.

app/src/components/ui/Surfaces.tsx
```diff
@@ -216,9 +217,15 @@ export function CardGroup({children, minWidth = 200, className, style, ...props}
...
-            gap: 14,
+            /*
+             * Converted to rem so a column grows with the interface scale.  As px it did not,
+             * and a 430px phone fitted two 200px cards where the pre-migration build fitted
+             * one 290px card -- the narrowest card in the app, on the smallest screen.
+             *
+             * The prop stays a px number: seven call sites pass one, and a caller thinking in
+             * px is thinking about the unscaled design.
+             */
+            gridTemplateColumns: `repeat(auto-fill, minmax(${minWidth / 16}rem, 1fr))`,
```
lrnselfreliance: **[nit]** `minWidth / 16` (and the same `/ 16` in Placeholder) hard-codes the design-time "1rem = 16px" assumption in two places without a shared helper. Correct today, but easy to drift if a second call site invents a different divisor or if the design base is ever documented as something else.

**Suggestion:** A tiny `pxToRem(n)` helper (or comment + single exported constant) used by CardGroup, Placeholder, and any future converters.

app/cypress/e2e/keyboard-shortcuts.cy.js
```diff
@@ -54,6 +54,27 @@ describe('Keyboard Shortcuts', () => {
```
lrnselfreliance, line 78: **[nit]** Removing `.mantine-Notifications-root` from the DOM to dodge occlusion is environment-specific and destructive. Scale makes toasts larger, so the toast-over-modal collision is more likely; the fix papers over that for one spec rather than addressing stacking/layout. Acceptable as a CI stabiliser given the long comment, but brittle if Notifications re-mounts mid-test.

**Suggestion:** Prefer intercepting the failing unmocked API routes in this spec, or dismiss via the Notifications API, so the test does not depend on surgically deleting UI.

lrnselfreliance (commented):
## Summary

The scale mechanism is sound: `--ui-scale: 1.1` with `html { font-size: calc(100% * var(--ui-scale)) }`, rem for scalable lengths, px for hairlines, and Mantine `theme.scale` left at default so rem growth is not cancelled. That preserves the user's browser font-size preference and avoids the `zoom`/`vh` overflow pitfalls documented in the PR. CardGroup → rem, card body styles moved out of inline px, and the poster `cqw` floor/ceiling fix are well reasoned, and Cypress expectations were updated to follow `--ui-scale` rather than hard-coded px.

The main gaps are completeness and guard quality, not the lever itself. Tabler icon sizes and most MapViewer chrome stay in absolute px; several mixed-unit pixel-perfect rules were only half-converted; and `ui-scale.test.js` overstates coverage (transforms, negative offsets, non-`fontSize` inline lengths, and SVG icon sizes are invisible to it). No ship-blocker in the mechanism, but a few follow-ups would make the 10% scale actually reach the whole interface.

## Issue counts by severity

- bugs: 0
- suggestions: 6
- nits: 4

## Issues outside the diff

- **[suggestion]** app/src/components/ui/Icon.tsx:17 — **[suggestion]** Named icon sizes remain absolute pixels (`small: 16`, `medium: 20`, `large: 24`) passed straight into Tabler's `size` prop. After root rem scaling, body text and Mantine control heights grow ~10% while every Icon stays put, so glyphs sit undersized inside larger buttons/ActionIcons and next to larger labels. This is the highest-impact incomplete conversion: icons are everywhere, and neither the stylesheet scan nor the inline `fontSize` scan can see them.
  **Suggestion:** Express icon sizes in rem (e.g. `1rem` / `1.25rem` / `1.5rem` via `style`/`width`/`height`, or `size` derived from root), and add a unit test that named sizes are not bare px numbers (or that computed icon boxes track `documentElement` font-size).

- **[suggestion]** app/src/components/MapViewer.js:642 — **[suggestion]** Pin-toggle control is injected with a `cssText` string of absolute px, including `font-size:18px`. That is a fixed-size control beside MapLibre's own chrome, but it is also the only explicit `font-size:…px` left in production source that the new "no px fontSize inline" test cannot see (it only matches React `fontSize:`).
  **Suggestion:** Style the control from a class (rem font-size/box) or accept it as map-chrome-fixed and document that exception in the test allowlist if intentional.

- **[nit]** app/src/components/ui/ui.css:46 — **[nit]** `.wrolpi-icon-stack-corner` still uses `right: -3px; bottom: -3px` while surrounding UI sizes move to rem. Negligible at 1.1, but same mixed-unit pattern and same test blind spot as Issues 4/5.
  **Suggestion:** Use rem offsets (or a small `em` relative to the stack) so corner badges track icon scale if icons are later rem'd.

app/src/components/MapViewer.js
lrnselfreliance: Fixed. 20 values across the layer panel, add-pin dialog, search panel and context menu — their type had grown 10% while padding hadn't, which is the drift you describe.

Converted the values rather than moving the overlays into CSS classes: the defect is the non-scaling, and this repo's standing preference is minimal targeted fixes over bundled refactors. Moving them into `ui.css` remains worth doing and is not done here.

Left in px on purpose, and your list was right about which: the context menu's `left`/`top` (a real click coordinate), maplibre's own `ScaleControl` maxWidth, and the `calc(100% - 100px)` reserve, which reserves room for maplibre's px-sized controls. The `100vh - 120px` reserve **did** become `7.5rem`, since the chrome it reserves for is ours — measured at 528px against 540px at scale 1.

Verified in the browser: panel `minWidth` 200 → 220, search 320 → 352, padding 12 → 13.2px.

app/src/components/ui/ui.css
lrnselfreliance: Both fixed.

**Tag eyelet:** converted, and centred with `translateY(-50%)` instead of a negative margin of half its own height — your second suggestion. A percentage cannot drift, so the two pieces can't come apart again at any scale.

**`cqw` fallback:** you're right that the whole declaration is discarded, which is worse than it sounds — the poster would have had *no* cap, so a portrait cover would stand as tall as the card is wide. A plain `max-height: var(--card-poster-max-height)` now precedes the `max(…, cqw)` rule.

app/src/themes/ui-scale.test.js
lrnselfreliance: All four points valid, and (1) and (2) were the real hole — the guard's name claimed more than it measured, which is a failure mode I'm supposed to watch for.

1. Added `transform`, `-webkit-transform`, `-ms-transform`, `translate`, `flex`, `flex-basis`.
2. The matcher now takes an optional leading `-`. This is what hid the tag eyelet and the icon-stack badge.

Together those found **nine** values, including `flex: 1 1 200px` in App.css that no comment mentioned. All converted or deleted.

3. Not swept, deliberately: there are ~176 numeric geometry values in 23 feature files. The map overlays were the visible ones and are fixed; the rest is a change to 23 unreviewed files. So instead the test is renamed to say **"in our stylesheets"**, and a new ratchet keeps numeric geometry out of `components/ui` so the eleventh inline `gap: 10` can't be added there.
4. Added, in `ui-layout.cy.js`: root font-size is the declared multiple of 16; our text **and** a Mantine control both grow by the same factor when the scale changes; a hairline does not. The middle one is what would catch `theme.scale` being set and silently cancelling the root.

Both new guards planted against.

app/src/App.css
lrnselfreliance: Deleted, per your first option — and it was more than a nit. Confirmed dead: nothing references either class since Toggle became a Mantine Switch, and the only remaining matches for the words were the `classList.toggle` DOM method and the controller's own templates, which don't load this stylesheet.

Deleted rather than finished because it was already broken exactly as you calculated, and it also hardcoded `#6435C9` and `white`, which no theme can reach. Reviving it would mean rewriting it anyway.

The extended guard now catches this class of thing: `transform` is on the property list, so a future `translateX(26px)` beside rem siblings fails.

app/src/components/ui/ui.css
lrnselfreliance: Both fixed.

**Tag eyelet:** converted, and centred with `translateY(-50%)` instead of a negative margin of half its own height — your second suggestion. A percentage cannot drift, so the two pieces can't come apart again at any scale.

**`cqw` fallback:** you're right that the whole declaration is discarded, which is worse than it sounds — the poster would have had *no* cap, so a portrait cover would stand as tall as the card is wide. A plain `max-height: var(--card-poster-max-height)` now precedes the `max(…, cqw)` rule.

app/src/components/ui/Surfaces.tsx
lrnselfreliance: Done — `pxToRem` in `components/ui/scale.ts`, with `REM_BASE` exported beside it, used by both CardGroup and Placeholder.

app/cypress/e2e/keyboard-shortcuts.cy.js
lrnselfreliance: Agreed, and switched to your first suggestion. Deleting a DOM node was the wrong instinct.

The four endpoints are mocked instead, now that a probe named them: `/api/tag`, `/api/events/feed`, `/api/files/search` and `/api/files/worker_status`.

Worth recording why the toast was there at all: **two of those were already "mocked" at the wrong paths** — `/api/tags` and `/api/events`, neither of which the app calls. Those mocks matched nothing, so the real calls fell through and failed. Wrong-path mocks in a passing spec.

One thing that does *not* work, so it's written down rather than retried: a blanket `/api/**` 200 stub. A 200 with the wrong shape fails identically — `data['totals']['file_groups']` on `{}` throws — which is what "Unexpected server response" means.

Verified under CI's condition (`forceNetworkError` on everything unmocked): 12 passing, no DOM surgery.

lrnselfreliance:
### All seven review comments addressed

Replied inline to each; summary here. The common thread is exactly the one my guard was written to catch and didn't: geometry converted to rem sitting beside geometry that stayed px.

**The guard was narrower than its name** — the most important comment. It missed `transform`/`flex`/`flex-basis`, and it missed **every negative length**, because the lookbehind treated the minus sign as part of the preceding token. That is what hid the tag eyelet (`left: -7px`, `margin-top: -2.5px`) and the icon-stack badge (`right: -3px`, `bottom: -3px`) — each a piece of geometry that had stopped tracking the piece it belongs to.

With both fixed it found **nine** values, including `flex: 1 1 200px` that no comment mentioned. All converted, except the dead `.toggle`/`.slider` block, which is deleted: nothing has referenced those classes since Toggle became a Mantine Switch, and it was already broken exactly as calculated — plus it hardcoded `#6435C9` and `white`, which no theme can reach.

The tag eyelet now centres with `translateY(-50%)` rather than a negative margin of half its own height, so the two pieces cannot drift again at any scale.

Also done:

- **Runtime check** (the gap in point 4): three cases in `ui-layout.cy.js` measure the real thing — root font-size is the declared multiple of 16; our text **and** a Mantine control both grow by the same factor; a hairline does not. The middle one is what would catch `theme.scale` being set and silently cancelling the root.
- **`cqw` fallback**: worse than it sounded — the whole declaration is discarded, so the poster would have had *no* cap and a portrait cover would stand as tall as the card is wide. A plain rem cap now precedes it.
- **`pxToRem`** in `components/ui/scale.ts`, with `REM_BASE`, used by CardGroup and Placeholder.
- **Map overlays**: 20 values converted. Left in px on purpose, as you listed: the context menu's `left`/`top` (a real click coordinate), maplibre's `ScaleControl` maxWidth, and the `calc(100% - 100px)` reserve for maplibre's px-sized controls. The `100vh - 120px` reserve *did* become rem, since that chrome is ours. Measured: panel `minWidth` 200 → 220, search 320 → 352.
- **keyboard-shortcuts spec**: switched to your first suggestion. The DOM deletion is gone; the four endpoints are mocked. Worth recording *why* a toast was there at all — two were already "mocked" at `/api/tags` and `/api/events`, **neither of which the app calls**, so those mocks matched nothing.

**Not done, deliberately:** ~176 numeric geometry values remain in 23 feature files. The map overlays were the visible ones. Sweeping the rest is a change to 23 unreviewed files, so instead the guard is renamed to say "in our stylesheets" rather than implying the whole app, and a new ratchet keeps numeric geometry out of `components/ui`.

### CI

Green except `api-tests-3-13`, which is **an unrelated known flake**: `test_search_suggestions` (`domains` came back `[]`). Evidence it is not this PR — the same job passed on the four earlier commits of this branch, `git diff base...HEAD --name-only` shows nothing outside `app/`, and the test passes 3/3 alone locally. It is the same xdist-ordering family as the long-standing channel-download-directory flake and was seen on another frontend-only branch on 2026-07-29.

Everything else: 1142 jest / 65 suites, 251 `ui-layout` + 10 `modal-stack` component tests, **all 50 e2e specs**, clean build. Both new guards planted against.

PR #632, opened 2026-08-04
ui: larger tags, tighter rows

lrnselfreliance:
Tags read small and sat loose.  The font is up 12%, the padding is untouched, and the gaps closed by about a third — but only two of the three numbers were a design decision.

## Size

A tag's word was 12.5px design against a 16px base — **78% of the body text beside it**.  That is right for a *label*: the count next to a tab's name, the page number in the CBZ viewer, things that annotate something else and want to sit under it.  It is wrong for a tag, which is the entire content of its chip and the thing a wall of them is scanned for.  On the dashboard, where the tags *are* the page, it read as a footnote.

Now 14px design — **87.5%**.  This is the one place labels and tags part company, so the size went on `.wrolpi-tag` and plain labels are unchanged.  Both appear in `/theme-sample`, in adjacent sections, and the difference reads as hierarchy rather than inconsistency.

The point had to grow with it.  Its diagonal *is* the body's height — 21px a side for the 30.2px body, where 20px served the 28.2px body — or the two edges meet the body short of its corners and it reads as a notch stuck on the side rather than one tag-shaped outline.  Which in turn made the margin reserving room for it a third of a pixel short, enough to clip against the tag before it, so that is the one length here that went **up**: 16px → 17px.

## Spacing

Neither gap was chosen:

- **The link wrapper carried `0.3em` either side**, left from when these were Semantic labels with no margin of their own.  A tag reserves its own `margin-left` for the point drawn outside its body, so this was 0.3em twice on top of the group's gap *and* the point's room.  Two tags sat **34.7px** apart where the point needs 18.4px.
- **`TagsLinkGroup` spread its props down to every chip**, so the dashboard's `marginTop` — meant to space the row below its heading — landed on all 34 tags.  That is not a margin above the row; it is 7.7px added between every wrapped row of it.  The tags were the only thing on the page whose rows were further apart than their group's own gap.

Measured on the dashboard, 34 tags:

| | before | after |
|---|---|---|
| font size | 13.75px | 15.4px |
| gap between tags | 34.7px | 25.3px |
| …of which is the point | 17.6px | 18.7px |
| visible space between them | ~17px | ~7px |
| gap between rows | 18.5px | 12.0px |
| rows needed | 4 | 4 (11 per row, was 9) |

## Tests

Written first.  Both margin tests fail on the real values — `0.3em` on the link, `0.5em` on the chip — rather than on something absent.

The size is asserted as a **ratio to body text**, not a pixel count, because a pixel floor here is satisfiable by accident: the 1.1 interface scale raised the tag to 13.75px and the prose to 17.6px at the same time, leaving the tag exactly as small beside it as before.  A ratio cannot pass that way.  It has an upper bound too, so a tag stays a chip rather than becoming prose in a colored box.

The existing geometry tests earned their keep — `reserves exactly the room the point takes` is what caught the third-of-a-pixel overhang, which no screenshot would have shown.

## Verification

- 1145 jest tests, 65 suites
- 252 cypress component tests in `ui-layout.cy.js`
- 50 cypress e2e tests
- clean `npm run build`
- all four themes in the browser, including **night**, where the point is drawn as two strokes rather than a solid diamond and so shows a corner mismatch plainly, and **amber**, which discards the user's fill

`Tags.cy.js`, `Channels.cy.js`, `Download.cy.js` and `Inventory.cy.js` fail — 12 tests — but they fail identically on `feature/custom-theme`, confirmed by running them there in a worktree.  `Tags.cy.js` is looking for `div.ui.label.large`, a Semantic selector that has not existed since the migration; the others are unmocked fetches.  Worth a separate cleanup.

### DIFF
--- a/app/src/Tags.js
+++ b/app/src/Tags.js
@@ -142,29 +142,41 @@ export function useTags() {
         </Group>
     }
 
+    /*
+     * The link carries no margin of its own.  It had `0.3em` either side, from when these were
+     * Semantic labels that had none -- but a tag already reserves a `margin-left` in ui.css for
+     * the point drawn outside its body, so that was 0.3em twice on top of the group's gap and
+     * the point's own room.  Two tags sat 34.7px apart where the point needs 18.4px.
+     */
     const TagLabelLink = ({name, props}) => {
         const to = getLocationStr({tag: name}, '/search');
-        const style = {marginLeft: '0.3em', marginRight: '0.3em'};
         try {
             // We prefer to use Link to avoid reloading the page, check if React Router is available, so we can use it.
             useNavigate();
-            return <Link to={to} style={style}>
+            return <Link to={to}>
                 <NameToTagLabel name={name} {...props}/>
             </Link>
         } catch {
             // React Router is not available, use anchor.
-            return <a href={to} style={style}>
+            return <a href={to}>
                 <NameToTagLabel name={name} {...props}/>
             </a>
         }
     }
 
-    const TagsLinkGroup = ({tagNames, ...props}) => {
+    const TagsLinkGroup = ({tagNames, style, ...props}) => {
         if (!tagNames || tagNames.length === 0) {
             return <React.Fragment/>;
         }
 
-        return <Group gap={6}>
+        /*
+         * `style` belongs to the group, not to each tag in it.  It used to go down with the rest
+         * of the props to every chip, so the dashboard's `marginTop` -- meant to space the row
+         * below its heading -- landed on all 34 tags and became 7.7px of extra space between
+         * every wrapped row.  The remaining props still reach the chips, which is how `onClick`
+         * gets there.
+         */
+        return <Group gap={6} style={style}>
             {tagNames.map(i => <TagLabelLink key={i} name={i} props={props}/>)}
         </Group>
     }
--- a/app/src/Tags.js
+++ b/app/src/Tags.js
@@ -146,7 +146,11 @@ export function useTags() {
      * The link carries no margin of its own.  It had `0.3em` either side, from when these were
      * Semantic labels that had none -- but a tag already reserves a `margin-left` in ui.css for
      * the point drawn outside its body, so that was 0.3em twice on top of the group's gap and
-     * the point's own room.  Two tags sat 34.7px apart where the point needs 18.4px.
+     * the point's own room.  Two tags sat 34.7px apart, of which the point needs 18.7px.
+     *
+     * Those are rendered px, measured in the browser at the default interface scale -- which is
+     * what a user sees, and why they are the numbers worth recording.  The stylesheet states the
+     * same margin in design px, as `1.0625rem`; 17 x 1.1 is the 18.7 above.
      */
     const TagLabelLink = ({name, props}) => {
         const to = getLocationStr({tag: name}, '/search');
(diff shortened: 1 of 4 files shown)


app/src/Tags.js
```diff
@@ -142,29 +142,41 @@ export function useTags() {
         </Group>
     }
 
+    /*
+     * The link carries no margin of its own.  It had `0.3em` either side, from when these were
+     * Semantic labels that had none -- but a tag already reserves a `margin-left` in ui.css for
+     * the point drawn outside its body, so that was 0.3em twice on top of the group's gap and
+     * the point's own room.  Two tags sat 34.7px apart where the point needs 18.4px.
```
lrnselfreliance: **[nit]** The comment says two tags sat too far apart “where the point needs 18.4px,” but the structural room in `ui.css` is `margin-left: 1.0625rem` (17 design-px; tip ≈ 16.85px). The new unit test comment says “17.3px.” The narrative numbers disagree with each other and with the stylesheet, which makes the next person distrust the carefully worked geometry comments.

**Suggestion:** Align the three places on the same figure (e.g. “~17px” / the 1.0625rem margin) so the story matches the CSS.

app/src/Tags.js
```diff
@@ -142,29 +142,41 @@ export function useTags() {
...
             // React Router is not available, use anchor.
-            return <a href={to} style={style}>
+            return <a href={to}>
                 <NameToTagLabel name={name} {...props}/>
             </a>
         }
     }
 
-    const TagsLinkGroup = ({tagNames, ...props}) => {
+    const TagsLinkGroup = ({tagNames, style, ...props}) => {
```
lrnselfreliance: **[suggestion]** `TagsLinkGroup` peels only `style` for the group and still forwards every other prop to each chip. That is correct for the one real call site (`style={{marginTop: '0.5em'}}` on the dashboard) and matches the comment about `onClick`, but `className`, Mantine `Group` props (`gap`, `wrap`, `justify`, …), or a future `data-*` meant for the row would still land on every chip the way `marginTop` used to. No current caller does that.

**Suggestion:** Optional hardening: document the dual-target contract next to the export, or invert to an explicit `chipProps` / only accept known group props. Not required for this PR’s fix.

lrnselfreliance (commented):
## Summary

PR #632 correctly enlarges tag chips relative to plain labels (0.875rem / 87.5% of body vs 0.78125rem), scales the point and its margin with that body height, and fixes two real spacing bugs in `TagsLinkGroup` / `TagLabelLink`. CSS geometry checks out under rem/`--ui-scale`, night’s dashed point reuses the same sizes, and the new tests assert ratio-to-body with inverse bounds plus the group-vs-chip style routing. No correctness bugs found; ship as-is.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

lrnselfreliance:
Both addressed in 4f4e8d28.

**The disagreeing figures** — you were right, and worse than the comment suggests: I changed one of the two `17.3px` mentions and left the other, so the file contradicted itself. The underlying confusion is a unit mix — my measurements are rendered px (taken in the browser at the default scale) while `ui.css` states the same margin in design px as `1.0625rem`. All three now read **18.7px rendered**, say that they are rendered, and point at the `1.0625rem` with the arithmetic (17 × 1.1). One figure, one unit, cross-referenced.

**The dual-target props** — taken as the cheap half. `className` now goes to the row with `style`: it is unambiguously a row concern and the likeliest next prop to hit exactly the bug this PR fixes. The split is documented next to the component, naming the ones still left pointing at the chips (`gap`, `wrap`, `justify`, `data-*`) so the next person adds them here rather than working around it at a call site.

I did not invert to an explicit `chipProps` — that reshapes a working API with one real call site, and the forwarding is load-bearing for `onClick`.

Instead the contract is now **tested**, both halves in one test, since they sit one destructure apart and the `style` half was wrong for as long as it existed. Confirmed it has teeth by planting the regression: routing `className` back through `...props` fails it, and passing it fails on the group assertion too.

1146 jest tests, 252 cypress component, 50 e2e, clean build.

PR #638, opened 2026-08-05
ui: pages own the space between their blocks

lrnselfreliance:
Panels shared edges — two 1px borders meeting, which reads as one tall surface with a rule across it rather than a seam between two.

**I did it the wrong way first, and the app punished it before I'd read anything.** `margin-bottom: 1rem` on `.wrolpi-panel` failed in both of the ways component margins are known to fail:

- It **compounded with every `Stack` that already held panels** — the conflict list, both halves of the one-time pad, the electrical calculators — spacing those at 35.2px against the app's 17.6px. I was one edit away from patching each with `gap={0}`, which is the "override the default spacing" symptom the literature names.
- The **`:last-child` reset silently did nothing.** WROLPi wraps panels in the responsive `Media` component, so a panel is frequently the only child of its own wrapper: `:last-child` matches it, zeroes the margin, and leaves it hard against the next panel. On Status that separated four pairs out of five and left the first touching — one page showing both behaviors, which reads as a rendering glitch.

The research is unambiguous that the parent owns this: [Margin considered harmful](https://mxstbr.com/thoughts/margin), [Every Layout's Stack](https://every-layout.dev/layouts/stack/), [CSS-Tricks](https://css-tricks.com/component-spacing-design-system/), [No Outer Margin](https://kyleshevlin.com/no-outer-margin/).

## Why a gap and not the owl selector

Every Layout recommends `> * + * { margin-top }`. That was the second attempt, and it is **also** broken here: `Media` injects `.fresnel-container { margin: 0; padding: 0 }` into the document at runtime — the same specificity as `.wrolpi-stack > * + *`, later in source order — so every wrapped block threw the margin away again. A third margin failure, same root cause.

A gap belongs to the parent and a child cannot zero it. It also handles `Media`'s hidden half for free: that branch is `display: none`, so it is not a flex item and contributes no gap, where the owl selector spaced around something invisible.

## Where the spacing lives

`.wrolpi-stack` goes on `PageContainer`, which every page uses, plus the four wrappers that stack blocks where a page's own stack cannot reach:

| site | why |
|---|---|
| `PageContainer` | every page's top-level blocks |
| `Settings` panel wrapper | holds its six panels in a wrapper of its own |
| below the video player | same |
| `ZimRoute`, `MapRoute` | keep their tab bar outside `PageContainer` |

Mantine's `Stack` is the same thing with the same 1rem gap, so the four sites already using it are already correct and untouched.

## The other half: blocks giving up margins they owned

Once the page owns the spacing, a block carrying its own adds to it. Removed from the tab bar (33px where everything else was 17.6), the shared search control bar and paginator on Videos/Archives/Docs/Files, the card grid, the file browser's filter row, and the inventory search.

**Two margins deliberately stay.** The file browser's `4em` reserves room for the sticky bottom buttons — `git log -S` traced it to 62645831, "Making file browser buttons stick to the bottom of the screen" — so it is load-bearing, not decoration. And heading margins stay: a heading wants more space above than below, which binds it to its content, and 41.8/28.6 reads correctly against a 17.6 baseline.

## Tests

The three failures are each planted as a test, because each one passed a naive assertion:

- `spaces a block that is wrapped as readily as a bare one` — the `Media` case
- `survives a block that zeroes its own margin` — fails against the owl selector, verified by reverting to it
- `contributes no gap for the half of a Media pair that is hidden`
- `leaves a panel with no outer margin, so a Stack can space it instead` — the inverse rule, which is what stops the compounding coming back

## Verification

- 1147 jest tests, 65 suites
- 259 cypress component tests in `ui-layout.cy.js`
- 50 cypress e2e tests
- clean `npm run build`
- browser, sixteen routes: `/`, `/videos`, `/videos/748`, `/archives`, `/docs`, `/files`, `/zim`, `/map`, `/inventory`, `/search?q=`, `/theme-sample`, `/admin`, `/admin/settings`, `/admin/status`, `/admin/control`, `/admin/extension` — every stacked pair 17.6px
- all four themes on the dashboard, identical spacing
- the map's 85vh box measures exactly 85vh and the video still plays; no horizontal overflow anywhere

`Tags.cy.js`, `Channels.cy.js`, `Download.cy.js` and `Inventory.cy.js` fail (12 tests) identically on `feature/custom-theme` — pre-existing, confirmed by running them there in a worktree.

### DIFF
--- a/app/src/components/Common.test.js
+++ b/app/src/components/Common.test.js
@@ -727,7 +727,7 @@ describe('PageContainer', () => {
         stacks.forEach(stack => expect(stack.textContent).toBe('Body'));
     });
 
-    it('is never rendered by a page that a PageContainer already wraps', () => {
+    it('is never rendered by a page whose route already provides page chrome', () => {
         /*
          * `PageContainer` carries `margin-top: 1em` and `padding: 1em`, so nesting one inside
          * another applies both twice: the page is pushed 1em further down and inset 1em further
@@ -736,10 +736,21 @@ describe('PageContainer', () => {
          * in another -- and read as "more margin than /archives/domains", which is exactly what
          * it was.
          *
-         * Nothing about a doubled margin throws, so this is a source scan.  It reads each route
-         * wrapper (a `PageContainer` holding `<Routes>` or `<Outlet/>`), takes the components it
-         * routes to, and checks the ones defined in the same file -- which is where a route and
-         * its pages normally live together.
+         * Nothing about a doubled margin throws, so this is a source scan.  It finds each router
+         * whose enclosing `return` provides page chrome, takes the components it routes to, and
+         * checks the ones declared in the same file -- which is where a route and its pages
+         * normally live together.
+         *
+         * "Chrome" is `PageContainer` OR a `wrolpi-stack` wrapper, not just the former.  Keying on
+         * `PageContainer` alone missed `MapRoute` and `ZimRoute`, which wrap their own content in a
+         * stack instead -- and three pages under them (`ManageMap`, `MapPins`, `ManageZim`) mounted
+         * a `PageContainer` each, so they sat a further 1em down and 1em in from the tab bar than
+         * the viewers beside them.  The same defect as Archive, in a shape the first scan could not
+         * see.
+         *
+         * Known limit: a layout that routes through `<Outlet/>` has its `<Route>` list in App.js,
+         * naming components from other files, so the page cannot be resolved from one file's text.
+         * That is `VideosTabLayout`, whose pages are clean today but are not covered here.
          */
         const src = path.join(__dirname, '..');
         const files = (dir) => fs.readdirSync(dir, {withFileTypes: true}).flatMap(entry => {
@@ -753,24 +764,9 @@ describe('PageContainer', () => {
         const offenders = [];
         for (const file of files(src)) {
             const text = fs.readFileSync(file, 'utf8');
-
-            /*
-             * EVERY `<PageContainer>` in the file, not just the first.  Checking only the first
-             * is what let this bug through a version of this test: in Archive.js the first one
-             * is `ArchiveSettingsPage`'s -- the offender -- and its block holds no `<Routes>`, so
-             * the file was skipped before the route wrapper 400 lines below was ever looked at.
-             */
-            for (const block of routeWrapperBlocks(text)) {
-                const routed = [...block.matchAll(/element=\{<(\w+)\s*\/>\}/g)].map(m => m[1]);
-                for (const name of routed) {
-                    // The component's body, bounded by the next top-level declaration.
-                    const declaration = new RegExp(`^(?:export\\s+)?(?:function|const)\\s+${name}\\b`, 'm');
-                    const found = declaration.exec(text);
-                    if (!found) continue;   // Defined elsewhere; out of this scan's reach.
-                    const after = text.slice(found.index + 1);
-                    const next = after.search(/^(?:export\s+)?(?:function|const|class)\s/m);
-                    const body = next === -1 ? after : after.slice(0, next);
-                    if (body.includes('<PageContainer>')) {
+            for (const block of chromedRouterBlocks(text)) {
+                for (const name of routedComponents(block)) {
+                    if (bodyOf(text, name).includes('<PageContainer>')) {
                         offenders.push(`${path.relative(src, file)}: ${name}`);
                     }
                 }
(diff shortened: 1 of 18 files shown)


lrnselfreliance:
Pushed `be7723a3` — spotted that `/archives/settings` had more margin than `/archives/domains`. It did, and for a reason that predates this PR.

**`ArchiveSettingsPage` wrapped itself in a second `PageContainer`.** `ArchiveRoute` already wraps its `<Routes>` in one, and `PageContainer` carries `margin-top: 1em` *and* `padding: 1em` — so that page sat 1em lower and 1em further in on all four sides than its sibling. It is a fragment now, and its panels are blocks of the route’s own stack.

Two more top-level blocks gave up margins that were being added to the page gap — the domain filter row (35.2px against 17.6px) and the channel filter row, each commented as "matching" the other’s pattern, so they had the same bug in the same words.

**The guard is a source scan**, since a doubled margin throws nothing: it takes each `PageContainer` that wraps a router, reads the components it routes to, and fails if one of them renders a `PageContainer` of its own.

Worth recording that **the first version of that scan passed against the live bug** — my recurring failure mode, a test whose claim outran what it measured. It read only the *first* `<PageContainer>` in a file, and in `Archive.js` that is the offender itself, whose block holds no `<Routes>`; the file was skipped before the route wrapper 400 lines below was ever reached. It now reads every occurrence, and the premise assertions count both the wrappers found (>2) and the routed components within them (>8), so a scan that matched nothing cannot report a clean result.

Measured after: `/archives/settings` and `/archives/domains` are identical — tab bar bottom 136.8px, first block 17.6px below it, content inset 17.6px — as are the settings and statistics pages of every other tab group (`/videos/settings`, `/videos/statistics`, `/videos/channel`, `/docs/statistics`, `/archives/statistics`).

1148 jest tests, 259 component, 50 e2e, clean build.

One unrelated thing found on the way: `takes the links back when the corner shrinks again` in `ui-layout.cy.js` asserts `expected 10 to be above 10` and fails under load, passing 259/259 on three consecutive solo runs. A strict `>` on a count that can legitimately equal its bound — a latent test bug, not touched here.

app/src/components/Map.js
```diff
@@ -649,7 +649,9 @@ export function MapRoute() {
         {text: 'Manage', to: '/map/manage', key: 'manage'},
```
lrnselfreliance, line 655: **[bug]** `MapPins` is a top-level stack of `TextInput` + `Media` tables inside its own `PageContainer`. The filter still has `marginBottom: '1em'`, so pin filter → table spacing is stack gap + `1em` after this PR adds `wrolpi-stack` on the route and on `PageContainer`.

**Suggestion:** Remove `marginBottom` from the filter `TextInput` style (keep `width: '100%'`).

app/src/components/Map.js
```diff
@@ -649,7 +649,9 @@ export function MapRoute() {
...
     ];
 
-    return <div style={{marginTop: '2em'}}>
+    // `wrolpi-stack`: as in ZimRoute, this route wraps its own content instead of using
+    // PageContainer, so the spacing below the tab bar is this element's to provide.
+    return <div className='wrolpi-stack' style={{marginTop: '2em'}}>
```
lrnselfreliance, line 660: **[suggestion]** Map/Zim routes now use `.wrolpi-stack` for tab bar + content, but several child pages still mount their own `PageContainer` (`MapPins`, `ManageMap`, `ManageZim`). Flex gap between tabs and that container does not collapse with `PageContainer`’s `marginTop: 1em` (and tablet padding), so manage/pins pages sit further below the tabs than viewers that skip `PageContainer`. Pre-existing nesting, made more visible by the new gap.

**Suggestion:** Prefer fragments (or content-only roots) under these route stacks, same rationale as `ArchiveSettingsPage`; keep one place for page chrome (`marginTop` / padding).

app/src/components/Common.test.js
```diff
@@ -707,3 +710,101 @@ describe('card links', () => {
...
+            const text = fs.readFileSync(file, 'utf8');
+
+            /*
+             * EVERY `<PageContainer>` in the file, not just the first.  Checking only the first
+             * is what let this bug through a version of this test: in Archive.js the first one
+             * is `ArchiveSettingsPage`'s -- the offender -- and its block holds no `<Routes>`, so
+             * the file was skipped before the route wrapper 400 lines below was ever looked at.
+             */
+            for (const block of routeWrapperBlocks(text)) {
+                const routed = [...block.matchAll(/element=\{<(\w+)\s*\/>\}/g)].map(m => m[1]);
```
lrnselfreliance: **[suggestion]** Nested-`PageContainer` scan only matches `element={<Name/>}` self-closing tags in the same file as a `PageContainer` that wraps `<Routes>`/`<Outlet/>`. It will not see App.js routes wrapped in `ErrorBoundary`, Outlet-only layouts whose pages live elsewhere, or Map/Zim’s non-`PageContainer` route stacks that still nest `PageContainer`. The Archive regression is covered; other nesting shapes are not.

**Suggestion:** Extend the scan (or add a sibling) for `className='wrolpi-stack'` route wrappers and for `element={<ErrorBoundary><Page/>…}` forms; optionally assert known multi-panel pages either flatten into a stack or set `wrolpi-stack` on their own wrapper (Controller).

lrnselfreliance (commented):
## Summary

The stack/`gap` approach is the right fix: parent-owned spacing survives Fresnel zeroing, `display: none` Media siblings, and composes with Mantine `Stack` without the `:last-child` trap. `PageContainer`, Settings, Map/Zim routes, VideoPlayer’s below-player stack, and the Cypress suite cover the hard cases well, and the ArchiveSettings nested-`PageContainer` fix is real. Completeness is not done: Controller still nests four panels in a plain `div` (the exact Settings hole), twin collection headers (Playlists, Docs authors/subjects) still carry outer margins Channels/Domains dropped, and several top-level margins (dashboard search, Map pins filter) still compound with the new gap.

## Issue counts by severity

- bugs: 5
- suggestions: 2
- nits: 2

## Issues outside the diff

- **[bug]** app/src/components/admin/ControllerPage.js:1634 — **[bug]** `ControllerPage` still wraps its four panel sections (`AdminControlsSection`, `ServicesSection`, `SambaSection`, `DiskSection`) in a plain `<div>`. Admin’s `PageContainer` stack only sees that one child, so the four panels keep sharing edges — the same “page-owned wrapper the outer stack cannot reach” failure Settings had, which this PR fixed with `className='wrolpi-stack'` on Settings but not here.
  **Suggestion:** Match Settings: `return <div className='wrolpi-stack'>…</div>` (or a fragment so the outer stack owns the four sections).

- **[bug]** app/src/components/Playlists.js:131 — **[bug]** `PlaylistsPage` header still uses `mb='1em'` on the filter `Group`. Channels dropped the same `mb` because it is a top-level page block under `PageContainer` and the stack gap already spaces it; Playlists is the twin list page and still doubles gap + `1em` between filter and `CollectionTable`.
  **Suggestion:** Drop `mb='1em'` (and mirror the Channels/Domains comment) so only `.wrolpi-stack` gap separates header and table.

- **[bug]** app/src/components/Docs.js:472 — **[bug]** `AuthorsPage` (and `SubjectsPage`) keep `style={{marginBottom: '1em'}}` on the filter header. Both sit under `DocsRoute`’s `PageContainer` + `TabLinks` stack, same pattern as Domains where the PR removed the outer margin. Gap + `1em` reintroduces the uneven spacing this PR is meant to end.
  **Suggestion:** Remove the header `marginBottom` on both pages so spacing is only the page stack gap.

- **[bug]** app/src/DashboardPage.js:455 — **[bug]** Dashboard `searchRow` is a top-level `PageContainer` child with `marginBottom: '2em'`. Flex gap does not collapse with child margins, so search-to-next spacing becomes `gap` (`1rem` ≈ 17.6px at `--ui-scale: 1.1`) + `2em` instead of the previous ~`2em`. Homepage search is the most visible surface; this is a direct regression of the PR’s “no outer margin on stacked blocks” rule.
  **Suggestion:** Drop the row’s `marginBottom` and rely on stack gap, or if a larger search separation is intentional, set a single source of spacing (e.g. only a larger gap on a local stack around search + body — not gap plus margin).

- **[nit]** app/src/components/admin/ExtensionPage.js:35 — **[nit]** `DestinationHint` still applies `style={{marginBottom: '1em'}}` on a `Panel` that already sits in a Mantine `Stack gap='md'`. That is the exact “panel outer margin + container gap” compound the Cypress inverse test documents for CSS margins. Pre-existing, not introduced by `.wrolpi-stack`, but it fights the PR’s composition story on `/admin/extension`.
  **Suggestion:** Drop the panel’s `marginBottom` and let the `Stack` gap alone.

- **[nit]** app/src/components/ui/ui-layout.cy.js:389 — **[nit]** Hard-coded gap bounds `12…28` px assume default `--ui-scale` (~`1rem` ≈ 17.6px). The scale test correctly checks ratio; these absolute bounds would false-fail if the suite ever mounts at a higher root scale. Acceptable today; brittle if cy mount defaults change.
  **Suggestion:** Derive bounds from `getComputedStyle(stack).rowGap` (or from `--ui-scale`) instead of magic pixels, or document that these specs require default scale.

lrnselfreliance:
All nine findings addressed in `75ff6eeb`. Every one held up — the summary was right that completeness was not done.

## The three nested PageContainers (suggestion on Map.js:654)

`ManageMap`, `MapPins` and `ManageZim` each mounted a `PageContainer` inside a route that already provides page chrome, so they sat a further 1em down and 1em in from the tab bar than the viewers beside them. All three are fragments now, and the unused imports are gone.

**The scan should have caught these and did not**, which is the more useful finding. It keyed on `PageContainer` route wrappers, and Map and Zim wrap their content in a `wrolpi-stack` div instead — so those routers were invisible to it. Three fixes:

- it locates **routers** first and then asks whether the enclosing `return` provides chrome, `PageContainer` **or** `wrolpi-stack`
- it resolves `element={<ErrorBoundary><Page/></ErrorBoundary>}` as well as the bare form, by taking every capitalised tag that is not routing furniture
- it knows about `class` components — `ManageZim` is one, and a scan that only knew `function`/`const` reported it clean even once the router was found

Premise assertions now count routers found (>3) and components resolved (>8). The `<Outlet/>` limit is real and now documented in the test: `VideosTabLayout` has its `<Route>` list in App.js naming components from other files, so it cannot be resolved from one file’s text. Those pages are clean today but uncovered.

## The five compounding margins

Dashboard search row, Playlists filter, both Docs filter rows, Map pins filter — all dropped. The extension page’s hint panel too, which was the same compound against a Mantine `Stack` gap.

**The dashboard one deserves calling out**: the first pass reported that page clean and it was not. The sweep measured wrapper-to-wrapper gaps, and that row’s `marginBottom: 2em` is *inside* a `Media` wrapper, so it never appeared in the measurement. 52.8px against 17.6px on the most-looked-at page in the app, and the measurement was at fault rather than the page. The row now takes its spacing from the stack alone; if the search deserves more separation than other blocks that is a design decision, and it should come from one source rather than gap-plus-margin.

## ControllerPage

Four panel sections in a plain wrapper, so the admin stack saw one child — the exact hole Settings had. Now `className=\x27wrolpi-stack\x27`, matching Settings.

## The magic pixels

Both bounds now derive from the stack’s own `row-gap` (`closeTo(declared, 0.5)`) rather than a 12–28px range. Worth noting the range was not merely brittle: at a larger scale it would have **false-failed a correct layout**, and the doubled-gap case it was meant to catch is now asserted directly instead of via a ceiling.

## Verification

1148 jest tests, 259 component tests, 50 e2e, clean build with no new lint warnings. The full component run is at the pre-existing baseline of 12 failures (Tags/Channels/Download/Inventory), unchanged.

One honest caveat: the browser tooling disconnected partway through this round, so these nine changes are verified by tests and by reading rather than by measuring the pages again. The five margin removals in particular have no guard behind them — a source scan cannot see them, because in every case the margin sits on a `const header = <div style={{...}}>` declared outside the JSX block and interpolated in, which is exactly why they survived the first sweep. Re-measuring `/`, `/playlists`, `/docs/authors`, `/docs/subjects`, `/map/pins`, `/map/manage`, `/zim/manage`, `/admin/control` and `/admin/extension` in a browser is worth doing before merge.

PR #639, opened 2026-08-05
ui: a row of buttons is spaced by the row

lrnselfreliance:
The archive page's six actions were touching, and Update sat lower than the rest. Both are the same defect.

Semantic's Button carried `margin: 0 .25em .25em 0`, so a page could list buttons as bare siblings and they arrived spaced — and separated from the line below when the row wrapped. Nothing replaced that margin when Semantic came out.

Two of those six answered with `marginTop: 0.5em`, which is the wrong axis. A Button is `display: inline-flex; vertical-align: middle`, and `vertical-align` is measured on the **margin** box, so half of any top margin shows up as the visible button dropping. The margin was covering the wrapped-row case in the only axis a margin can, and paying for it in the other one.

## The fix

`.wrolpi-button-row` — a flex row with a gap, wrapping, items centred. Both axes at once, so no button needs a margin of its own. It is the same rule as `.wrolpi-stack` from the last PR, along the other axis, at the 0.625rem gap `.wrolpi-modal-actions` already uses.

Applied to the five rows in the app that had no gap:

| Where | Was |
|---|---|
| Archive page actions | six bare siblings; two with `marginTop: 0.5em` |
| Collection edit form (domains, channels, playlists) | a flex row that did not wrap, so nine action buttons plus Cancel each carried `margin-top: 1em` — and sat 1em below the Save beside them, which had none |
| Doc page, both responsive branches | bare siblings in a plain div; a `<span>` held the format picker apart with `margin-right` |
| Downloads table | two icon buttons sharing an edge |

`.action-button-spacing` is gone with them — it was that same `margin-top: 1em`, on Cancel.

The downloads cell moved from `text-align: right` to `justify-content: flex-end`, which is what a flex row answers to.

## Guards

**ui-layout.cy.js**, four tests, real layout: the row spaces what it holds and wraps rather than overflowing; it centres its buttons against a taller item beside them; a confirmed `APIButton` counts as one item rather than two (it returns a fragment holding a `<Confirm/>`, which would open a hole in the row if it did not portal); the gap grows with the interface scale.

The wrap test narrows the row to two buttons' width **measured** rather than guessed — I wrote 150px first, which was a hidden assertion about the font and the scale, and at 1.1 it fitted one button rather than the two the assertions were about.

The centring test was tops-agree first, and had no teeth: `stretch` stretches the *wrapper* and leaves the button at the top of it, so the tops agreed either way. Centres are what the property actually decides, and the comment in ui.css that claimed otherwise is corrected. Verified by deleting `align-items` and watching it fail.

**Common.test.js**, three source scans, because neither failure mode throws:

- no button inside a row stamps a vertical margin — caught all ten before the fix
- every row of buttons is in a row, checked by walking tags backwards to the enclosing element
- no run of adjacent button interpolations sits outside one — this is the reported defect, and it names the archive page when I rename the class away

A fourth reads each row class out of the stylesheet and confirms it really is a flex row with a gap, so the list of accepted class names cannot become a list of names the scans merely tolerate. `preview-toolbar-group` is on it: the file preview's toolbar was already a correct row under a different name, at a tighter gap because its buttons are icons.

Two scanner bugs worth recording. The resolver took the *first* `const deleteButton` in a file, and Archive.js has two — the archive page's, which was clean, and the domain page's, which was not; it reported the offender as clean. And `=>` in a prop puts a `>` inside an opening tag, which splits one button into two bogus tags, so arrows are blanked before any tag scanning.

## Verified

1152 jest tests / 65 suites, 263 component tests in ui-layout, 50 e2e including `domains/domain-edit.cy.js`, clean `npm run build` with no new lint warnings. Full component run sits at the pre-existing baseline of 12 failures (Tags, Channels, Download, Inventory) — the Channels one is a missing context provider in the mount, unrelated.

The gallery gets the row, under Buttons, narrowed so it visibly wraps.

## Not verified

No browser. These are measured by tests and read, not looked at. `/archives/:id`, `/archives/domain/:id`, `/videos/channel/:id/edit`, `/playlists/:id/edit`, `/docs/:id` and `/admin/downloads` are the pages that changed.

One judgment call: the doc page's tablet branch loses the `<span>` that held its format picker 1em from the actions, so that spacing drops to the row's 0.625rem like everything else in the row. If the picker should read as a separate group there, that wants saying in one place rather than as a margin on the span.

### DIFF
--- a/app/src/components/Common.test.js
+++ b/app/src/components/Common.test.js
@@ -805,10 +805,36 @@ describe('a row of buttons', () => {
     const tagsIn = (text) => [...scannable(text).matchAll(TAG)].map(match => ({
         start: match.index,
         end: match.index + match[0].length,
+        // The name was not captured until a test needed to accept `Group` and `Modal.Actions` as
+        // rows: `ROW_COMPONENTS.has(undefined)` is false for every tag, so the list of components
+        // that lay out their own children silently accepted none of them.
+        name: match[2],
         attributes: match[3],
         closing: match[1] === '/',
         selfClosing: match[4] === '/',
     }));
+
+    /**
+     * The text inside the element opened by `open`, found by walking forward to its MATCHING close.
+     *
+     * Written as "slice to the first `</div>`" first, which is true of every row in the app today
+     * and is a false negative waiting to happen: one nested div -- an error-trigger wrapper, a
+     * format-picker span, the Cancel/Save pair the collection form now groups -- truncates the body,
+     * and every button after it goes unscanned.
+     */
+    const bodyOfElement = (text, tags, open) => {
+        let depth = 0;
+        for (const tag of tags) {
+            if (tag.start <= open.start || tag.selfClosing) continue;
+            if (tag.closing) {
+                if (depth === 0) return text.slice(open.end, tag.start);
+                depth -= 1;
+            } else {
+                depth += 1;
+            }
+        }
+        return text.slice(open.end);   // Unclosed; scan what is left rather than nothing.
+    };
 
     /**
      * The tag that encloses `index`: walk the tags before it backwards, letting a close push the
@@ -863,8 +889,40 @@ describe('a row of buttons', () => {
      */
     const ROW_CLASSES = [
         'wrolpi-button-row', 'wrolpi-modal-actions', 'wrolpi-confirm-actions', 'preview-toolbar-group',
+        // Two that predate `.wrolpi-button-row` and are already correct: the file browser's sticky
+        // footer, and its filter row.  Both are checked against the stylesheet below like the rest.
+        'sticky-footer', 'file-browser-filter-container',
     ];
-    const isRow = (tag) => !!tag && ROW_CLASSES.some(name => tag.attributes.includes(name));
+
+    /*
+     * Components that lay their own children out in a row.  A `Group` or a `ButtonGroup` is a row by
+     * construction, so a caller putting buttons in one has already answered the question.
+     */
+    const ROW_COMPONENTS = new Set(['Group', 'ButtonGroup', 'Row', 'Stack', 'Modal.Actions']);
+
+    /*
+     * An inline flex or grid with a gap of its own -- an anonymous row.  Several call sites write one
+     * rather than reach for a class, and they are not wrong; a scan that only knew about class names
+     * would report every one of them.
+     */
+    const INLINE_ROW = /display:\s*['"]?(?:flex|grid)/;
+    const isInlineRow = (attributes, text) => {
+        if (INLINE_ROW.test(attributes) && attributes.includes('gap')) return true;
+        /*
+         * `style={someStyle}` -- the style is a const somewhere above, so the attribute itself says
+         * nothing.  Resolving it matters: the seasons row on the dates selector is a four-column
+         * grid with a gap, declared exactly that way, and it would otherwise be reported forever.
+         */
+        const named = /style=\{(\w+)\}/.exec(attributes);
+        if (!named) return false;
+        const declared = new RegExp(`\\b${named[1]}\\s*=\\s*(\\{[^;]*?\\});`).exec(text);
+        return !!declared && INLINE_ROW.test(declared[1]) && declared[1].includes('gap');
+    };
+
+    const isRow = (tag, text = '') => !!tag && (
+        ROW_COMPONENTS.has(tag.name)
+        || ROW_CLASSES.some(name => tag.attributes.includes(name))
+        || isInlineRow(tag.attributes, text));
 
     it('lays every row out as a flex row with a gap', () => {
         // Otherwise the list above is just a list of names the scans below happen to accept.
@@ -895,11 +953,10 @@ describe('a row of buttons', () => {
             const where = path.relative(SRC, file);
 
             const blocks = actionFragments(text);
-            for (const tag of tagsIn(text)) {
+            const tags = tagsIn(text);
+            for (const tag of tags) {
                 if (!tag.closing && tag.attributes.includes('wrolpi-button-row')) {
-                    // A row holds buttons, not more divs, so its first `</div>` is its own.
-                    const closes = text.indexOf('</div>', tag.end);
-                    blocks.push(closes === -1 ? '' : text.slice(tag.end, closes));
+                    blocks.push(bodyOfElement(text, tags, tag));
                 }
             }
             rows += blocks.length;
@@ -963,7 +1020,7 @@ describe('a row of buttons', () => {
         expect(offenders).toEqual([]);
     });
 
-    it('never lists buttons as bare siblings', () => {
+    it('never lists interpolated buttons as bare siblings', () => {
         /*
          * The touching itself, and the reported defect: the archive page interpolated six buttons
          * one after another straight into its Panel, so every pair shared an edge.
@@ -972,6 +1029,8 @@ describe('a row of buttons', () => {
          * `actionButtons` fragment are blanked first -- a fragment is spread into its parent's row
          * and has no business being one itself, which the previous test is what checks.  Blanked
          * rather than skipped so every index still points where it did.
+         *
+         * The sibling test below covers the other way a row is written; this one only sees `{foo}`.
          */
         const offenders = [];
         let runs = 0;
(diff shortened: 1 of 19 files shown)


lrnselfreliance:
Pushed `74f25aab` for the Back button. That one is mine from #638, not a Semantic leftover.

`.wrolpi-stack` is a flex column at `align-items: stretch` — deliberately, because a panel was full-width as a block element and has to stay that way. A **bare** button as a direct child of that stack gets stretched too, and every page's Back button was one. So it is all nine pages that have one, not just the two named. On the domain, channel and playlist edit pages it also pushed the button beside it onto its own line, where the two had been inline.

The row absorbs it without touching the stack: the row is the stretched child, and it lays its buttons out at their own width. The three pages with a second button get one row holding both.

Two guards. A cypress test asserts both halves — the row stretches like a panel does *and* the button inside does not — because a row that also refused to stretch would pass a test that only measured the button. And a scan asserts every `<BackButton/>` in the app sits in a row; it names `components/Docs.js:342` when I unwrap one.

Also rowed the Videos, Archives and Documents select bars, whose Tag/Delete/Invert/Clear were four touching buttons in a bare div. Worth noting *why* the scans missed those: they are inline `<Button>` elements, and the bare-siblings scan looks for runs of `{fooButton}` interpolations.

## What I did not sweep

Scanning for adjacent button *elements* rather than interpolations finds ~80 sites across 30 files. Most are already inside a gapped container — `Row` in the gallery, `Group` on the controller page, `Modal.Actions` — so they are fine, and telling those from the touching ones needs a parent-by-parent pass I have not done. The three select bars are in because they are on pages this PR already touches and their parent is plainly a bare div. The rest is a follow-up, and I would rather name it than imply this PR covered it.

Still no browser on any of this. `/archives/domain/:id`, `/videos/channel/:id/edit`, `/playlists/:id/edit`, `/archives/:id`, `/docs/:id`, `/videos`, `/archives`, `/docs`, `/playlists/:id`, `/videos/video/:id` and `/videos/channel/new` are the pages that changed across both commits.

1153 jest / 65 suites, 264 in ui-layout, 50 e2e, component suite at the 12-failure baseline, clean build with no new lint warnings.

app/src/components/Common.test.js
```diff
@@ -785,6 +779,276 @@ describe('PageContainer', () => {
...
+        let runs = 0;
+
+        for (const file of sourceFiles()) {
+            let text = fs.readFileSync(file, 'utf8');
+            for (const fragment of actionFragments(text)) {
+                text = text.replace(fragment, ' '.repeat(fragment.length));
+            }
+            const tags = tagsIn(text);
+
+            for (const found of text.matchAll(/\{\w*[Bb]utton\w*\}(?:\s*\{\w*[Bb]utton\w*\})+/g)) {
```
lrnselfreliance, line 1045: **[suggestion]** The “never lists buttons as bare siblings” scan only matches adjacent `{…Button…}` interpolations (`/\{\w*[Bb]utton\w*\}(?:\s*\{\w*[Bb]utton\w*\})+/`). Inline JSX sequences (`<Button/>…<APIButton/>…`) never match, which is exactly how VideoPlayer’s actions are written and how several other residual rows in the app are written (e.g. Zim outdated-message actions, One-Time Pad page buttons). Combined with `expect(runs).toBe(4)`, the test will keep reporting green while inline bare rows remain or reappear.

**Suggestion:** Either extend the scan to catch adjacent opening tags that are button-like (`Button`, `APIButton`, `IconButton`, `AddToPlaylistButton`, etc.) when their enclosing parent is not a known row class / `ButtonGroup` / `Group` / `Stack` / `Modal.Actions`, or add a focused assertion for known high-value pages (VideoPlayer at minimum). Document that modal footers and sticky-footer are out of scope if intentionally excluded.

app/src/components/Common.test.js
```diff
@@ -785,6 +779,276 @@ describe('PageContainer', () => {
...
+        let resolved = 0;
+
+        for (const file of sourceFiles()) {
+            const text = fs.readFileSync(file, 'utf8');
+            const where = path.relative(SRC, file);
+
+            const blocks = actionFragments(text);
+            for (const tag of tagsIn(text)) {
+                if (!tag.closing && tag.attributes.includes('wrolpi-button-row')) {
+                    // A row holds buttons, not more divs, so its first `</div>` is its own.
```
lrnselfreliance: **[suggestion]** When collecting `.wrolpi-button-row` blocks for the vertical-margin scan, the body is sliced to the first `</div>` after the open tag (`// A row holds buttons, not more divs`). That is true of every current call site, but a single nested `<div>` (error trigger wrapper, format-picker wrapper, future layout tweak) would truncate the scanned body and silently skip later buttons — a false-negative path for the margin check.

**Suggestion:** Parse with the same tag-depth walker already used by `enclosingTag` / `tagsIn` to find the matching close, or ban nested divs in the row via a separate assertion if that is a hard rule.

app/src/components/Docs.js
```diff
@@ -403,13 +403,16 @@ function DocPage() {
...
                     </Media>
                     <Media greaterThanOrEqual='tablet'>
-                        <div style={{marginTop: '1em'}}>
-                            {formatButtons && <span style={{marginRight: '1em', display: 'inline-block'}}>
-                                {formatButtons}
-                            </span>}
+                        {/* The row's own `margin-top` separates it from what is above; the gap
+                            inside it separates the format picker from the actions, so the span
+                            that used to hold them apart with `margin-right` is gone. */}
+                        <div className='wrolpi-button-row' style={{marginTop: '1em'}}>
```
lrnselfreliance, line 421: **[nit]** On tablet+, the format picker and action buttons now share one `.wrolpi-button-row`, so separation is the row gap (`0.625rem`, same as modal actions) rather than the previous `marginRight: '1em'` on the format span. That is a deliberate, consistent choice, but it is noticeably tighter (~11px at default scale vs ~1em) between two visually different groups (segmented format control vs free actions). Mobile still stacks them in separate blocks with `marginTop: '1em'`, so the two breakpoints disagree more than before about how far the picker sits from the actions.

**Suggestion:** If the old separation was intentional visual grouping, a slightly larger gap only between groups (e.g. wrap `formatButtons` in a span with `marginRight`/`marginInlineEnd`, or a second class) would preserve “row spaces buttons” while keeping picker/actions distinct. If unifying on `0.625rem` is the product call, leave as-is and drop the old-span comment’s implication that nothing was lost.

app/src/components/collections/CollectionEditForm.js
```diff
@@ -60,14 +60,21 @@ export function CollectionEditForm({
...
                     {actionButtons}
                     {onCancel && <Button
                         type='button'
                         role='cancel'
                         onClick={onCancel}
                         disabled={form.disabled}
-                        className="action-button-spacing"
                     >
                         Cancel
                     </Button>}
```
lrnselfreliance: **[nit]** Save still uses `style={{marginLeft: 'auto'}}` inside the now-wrapping row. Horizontally this is correct and the vertical-margin scanner correctly ignores it. With `flex-wrap: wrap` newly enabled, on narrow widths Save can sit alone on a second line flush right while Cancel stays with the action buttons — usually fine for form chrome, but it is a layout change from the old non-wrapping flex row that callers should be aware of on phone.

**Suggestion:** Spot-check Domain/Channel/Playlist edit on a narrow viewport. If the lone right-aligned Save line feels wrong, consider `justify-content: flex-end` on a nested group for Cancel+Save, or `margin-inline-start: auto` on a wrapper around Cancel+Save so they travel together.

lrnselfreliance (commented):
## Summary

The new `.wrolpi-button-row` primitive is CSS-correct (`display: flex`, `flex-wrap: wrap`, `align-items: center`, rem `gap` that tracks `--ui-scale`), and the call sites that motivated the PR — Archive actions, CollectionEditForm, Docs actions, Downloads table icons, Back buttons — are cleaned up well. Source scanners and Cypress geometry tests give this real teeth, including the APIButton fragment/Confirm portal case. The main gaps are completeness: the video detail page still lists Download/Delete/Refresh as bare siblings in a `<p>`, and the “bare siblings” scanner only sees `{fooButton}` interpolations, so that class of miss can regrow silently.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 2

## Issues outside the diff

- **[bug]** app/src/components/VideoPlayer.js:488 — **[bug]** The video page’s action cluster (Download, Delete, Refresh, AddToPlaylist) is still a run of bare button elements inside a plain `<p>`, with no `.wrolpi-button-row` (or other gapped flex parent). After Semantic’s per-button margins disappeared, these share edges — the same defect class the Archive action row fixed. The PR did wrap this page’s `BackButton` in a row (line 400) but left the page’s primary action row untouched, so the fix is incomplete on a high-traffic detail page that is the video analogue of Archive.
  **Suggestion:** Replace the `<p>` wrapper with `<div className='wrolpi-button-row'>` (drop the paragraph if it is only a button host), matching Archive/Docs. Prefer the same pattern for any other media-detail action strips found the same way.

lrnselfreliance:
Five findings, all valid, all addressed in `5c6dee6c`.

**[bug] VideoPlayer's action strip** — fixed, and it is the most useful finding in the review, because the scan that was supposed to catch exactly this reported the app clean. It only ever saw `{fooButton}` interpolations, and most of the app writes its buttons out in place. So rather than the focused per-page assertion offered as the fallback, I extended the scan — which turned my "~80 unswept, needs a parent-by-parent pass" into a list.

**Twelve more rows** were in the same state: reorder icons in the playlist and field-schema tables, both service tables on the controller page, its hotspot Save/QR pair and its Restart/Shutdown pair, the downloads footer cell, the one-time pad's Generate/Cheat Sheet, and the ration calculator's CSV/Print. All rowed.

Two holes had to be closed first, because a scan that reports noise is worse than one with a stated limit:

- **`tagsIn` never captured the tag name.** `ROW_COMPONENTS.has(undefined)` is false for every tag, so the list of components that lay out their own children — `Group`, `ButtonGroup`, `Modal.Actions` — silently accepted none of them. 56 false positives, every modal footer in the app among them. Worth noting the shape: I added an accepted-parent list whose entries the scanner had no way to read.
- **A row whose style is a const** (`style={seasonRowStyle}`) says nothing in its own attributes. Resolved now — the dates selector's seasons row is a four-column grid with a gap and was reported three times.

The scan's definition of a row is deliberately broad: the class, a `Group`/`ButtonGroup`/`Stack`/`Modal.Actions`, or an inline flex/grid with a gap. Whether buttons are spaced by *something*, not whether they use this PR's class. `sticky-footer` and `file-browser-filter-container` join the class list and are checked against the stylesheet like the rest.

**[suggestion] the first-`</div>` slice** — replaced with a walk to the matching close. You were right that it was a false-negative path, and it was about to bite: the collection form now nests a div, which is exactly the case that would have truncated the body.

**[nit] Save's `margin-left: auto`** — Cancel and Save now travel together in a row of their own pushed to the end. Correct for the non-wrapping row it was written for; with wrap enabled Save would drop to a second line alone, flush right, with Cancel left behind among the action buttons.

**[nit] the doc page's picker spacing** — kept unified on the row gap, because one spacing rule for buttons in a row beats a second one for buttons that happen to be a picker. But the comment no longer implies nothing was lost; it now says plainly that the picker sits closer than it did and that this is a visual change, not a refactor. The mobile/tablet disagreement you noted is real and predates this PR structurally — mobile has always put the picker on its own line.

One more found on the way: the file browser's filter row is flex with **no** gap, and its collapse-all button carried `margin-left: 0.5em` to compensate. Same pattern, one element to the right of where it belongs.

Teeth checked by reverting: unwrapping VideoPlayer's row makes the new scan name `components/VideoPlayer.js:492`, `:499`, `:505`.

1154 jest / 65 suites, 264 in ui-layout, 50 e2e, component suite at the 12-failure baseline, clean build. Still no browser — the count of pages that changed is now up to about eighteen, so a look before merge is worth more than it was.

PR #640, opened 2026-08-05
ui: an input goes under its label, not beside it

lrnselfreliance:
## Problem

`InputForm` returned a fragment. Its `<label>` and its field `<div>` therefore arrived as two separate children of whatever contained them:

- **`/videos/settings`** — the settings grid (`repeat(auto-fit, minmax(240px, 1fr))`) put each of them in its own cell, so the **Video File Format** and **Extra yt-dlp Arguments** inputs rendered in the column *to the right* of their own labels.
- **Channel edit page** — `InputForm`'s label was inline with no bottom gap, while the neighbouring `DestinationForm` uses `FieldLabel` (block, 4px). **Channel Name** and **Directory** started at different heights.

## Fix

`InputForm` returns one `<div>`, and its label is block with the same 4px gap `FieldLabel` uses. Two changes, one component, no call sites touched.

## Tests

Two guards in `useForm.test.js`, both verified to fail against the old component and pass against the new one:

- the label and the input are the single child of a containing element
- the label computes to `display: block`

Full frontend suite: 65 suites / 1155 tests passing. `npm run build` clean.

Not visually confirmed in a browser — the change was made in a worktree and the dev server mounts the main checkout.

### DIFF
--- a/app/src/hooks/useForm.js
+++ b/app/src/hooks/useForm.js
@@ -356,8 +356,15 @@ export function InputForm({
         </Message>;
     }
 
-    return <>
-        <label htmlFor={`${name}_input`}>
+    /*
+     * One element, not a fragment: every caller that puts an InputForm in a grid or flex
+     * container was getting the label and the input laid out as two separate items -- the
+     * input beside its own label rather than under it.  The label is block with the same
+     * 4px gap the migrated fields (Download.js FieldLabel) use, so an InputForm and a
+     * DestinationForm side by side line their inputs up at the same height.
+     */
+    return <div>
+        <label htmlFor={`${name}_input`} style={{display: 'block', marginBottom: 4}}>
             <b>{label} {required && <RequiredAsterisk/>}</b>
             {helpContent &&
                 <InfoPopup
@@ -375,7 +382,7 @@ export function InputForm({
             />
         </div>
         {messageElm}
-    </>
+    </div>
 }
 
 export function NumberInputForm({
--- a/app/src/hooks/useForm.test.js
+++ b/app/src/hooks/useForm.test.js
@@ -944,6 +944,44 @@ describe('Form Components', () => {
             const input = screen.getByRole('textbox');
             expect(input).toBeDisabled();
         });
+
+        it('puts the label and the input inside a single element', () => {
+            // A fragment made them two children of whatever contained them, so a grid or flex
+            // parent laid the input out beside its label instead of under it.
+            const form = createMockForm({file_name_format: ''});
+
+            renderWithProviders(
+                <div data-testid="grid">
+                    <InputForm
+                        form={form}
+                        name="file_name_format"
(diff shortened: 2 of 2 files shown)


app/src/hooks/useForm.test.js
```diff
@@ -944,6 +944,44 @@ describe('Form Components', () => {
...
+            renderWithProviders(
+                <InputForm
+                    form={form}
+                    name="name"
+                    label="Channel Name"
+                />
+            );
+
+            const label = screen.getByText(/Channel Name/).closest('label');
+            expect(label).toHaveStyle({display: 'block'});
```
lrnselfreliance: **[nit]** The block-label test only asserts `display: 'block'`. Half of the alignment contract called out in the implementation comment is `marginBottom: 4` (to match `FieldLabel` / `DestinationForm` input baseline). A future edit could drop the margin and still pass.

**Suggestion:** Also assert `marginBottom: '4px'` (or the style object value the suite already normalizes) on the same label element.

app/src/hooks/useForm.js
```diff
@@ -356,8 +356,15 @@ export function InputForm({
...
-        <label htmlFor={`${name}_input`}>
+    /*
+     * One element, not a fragment: every caller that puts an InputForm in a grid or flex
+     * container was getting the label and the input laid out as two separate items -- the
+     * input beside its own label rather than under it.  The label is block with the same
+     * 4px gap the migrated fields (Download.js FieldLabel) use, so an InputForm and a
+     * DestinationForm side by side line their inputs up at the same height.
+     */
+    return <div>
+        <label htmlFor={`${name}_input`} style={{display: 'block', marginBottom: 4}}>
```
lrnselfreliance, line 367: **[nit]** Label spacing is duplicated as an inline style next to the same pattern in `Download.js` `FieldLabel` (`display: 'block', marginBottom: 4`). The comment documents the intentional match; there is no shared helper, so the two can drift.

**Suggestion:** Acceptable for this small fix. Optionally later extract a shared label style or export `FieldLabel` so `InputForm` and migrated fields stay in lockstep without comment-only coupling.

lrnselfreliance (commented):
## Summary

PR #640 correctly fixes `InputForm` returning a React fragment so label and field became separate grid/flex children (input beside its own label). Wrapping in a single `<div>` and giving the label `display: 'block'` + `marginBottom: 4` matches `FieldLabel` in `Download.js`, so side-by-side `InputForm` / `DestinationForm` (e.g. Channel edit/create) can align. `NumberInputForm` and `UrlInput` inherit the fix; other `useForm` helpers already use a single wrapper. Tests cover the structural parent and block label. Base is `button-rows` (stacked); the diff only touches `useForm` so the layout fix is independent of button-row work, but merge target should be confirmed before landing on main.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 2

PR #641, opened 2026-08-05
ui: a progress bar's track is its own edge, not a border

lrnselfreliance:
Removes the 1px border around progress bars, at the same size.

## What changed

`box-sizing: border-box` is global and the declared `1.375rem` height / `7.5rem`
min-width were always the outer box, so dropping the border leaves the bar
exactly the size it was — the 1px comes back as track, not as a smaller bar.

The border was also the only thing giving the *empty* half of the track an edge,
so the track background has to carry that now: `--head` instead of `--bg`, and
the light-mode `--panel` override goes away with it.

- A bar sits on a Panel as often as on the page — Status and Downloads put them in
  Panels and table rows, uploads in modals — and `--bg` is invisible against one of
  those two by definition. `--head` is off both surfaces in all four themes.
- `--head` also keeps the label's contrast against the empty track above 4.5:1 in
  every theme. A lighter track would read better as a surface but fails that:
  `--border` measures 3.4:1 in night.

## Testing

New per-theme assertions in `ui-layout.cy.js`: no border on any edge, height and
min-width measured against `rem` rather than px, and the track distinct from the
Panel beneath it. Restoring the border fails all four — checked.

- `npx cypress run --component --spec src/components/ui/ui-layout.cy.js` — 263 passing
- `CI=true npm test -- --watchAll=false` — 65 suites, 1148 passing
- `npm run build` — clean

### DIFF
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -799,19 +799,37 @@ describe('a progress bar has room for the label inside it', () => {
         });
     });
 
+    /*
+     * What the track really separates from, per theme, and NOT a single floor: the monochrome
+     * themes collapse the surface ramp, so one number either lets light and dark regress or
+     * fails night for being what it deliberately is.  These are the measured ratios rounded
+     * down, so each theme is held to its own behaviour -- night's near-flat 1.05 / 1.01 is
+     * recorded as such rather than passed off as an edge.  See the note in ui.css for why
+     * night cannot have a brighter track: the label crossing it drops below 4.5 if it does.
+     */
+    const trackFloors = {
+        light: {panel: 1.2, page: 1.1},
+        dark: {panel: 1.2, page: 1.08},
+        amber: {panel: 1.3, page: 1.4},
+        night: {panel: 1.04, page: 1.005},
+    };
+
     themeNames.forEach((theme) => {
         it(`draws no outline of its own in ${theme}, and is the size it was with one`, () => {
             /*
-             * The border came off.  Both halves matter: the bar must not be the 1px smaller
-             * inside that removing an outline usually makes it -- it is not, because
-             * `box-sizing: border-box` is global and the declared height was always the outer
-             * box -- and the empty part of the track has to keep an edge of its own now that
-             * no line is drawing one, which is why the track paints `--head` rather than the
-             * `--bg` / `--panel` pair that a bar sits on.
+             * The border came off, and the bar must not be the 1px smaller inside that
+             * removing an outline usually makes it -- it is not, because `box-sizing:
+             * border-box` is global and the declared size was always the outer box.
+             *
+             * The bar is mounted in a wrapper NARROWER than its min-width so that the floor is
+             * exercised by layout.  Reading `style.minWidth` instead, as the first version of
+             * this did, only repeats the declaration back: it is the same string whether or not
+             * the box it produces is right, so a content-box regression would have passed.
              *
              * Measured against `rem`, not px: 1.375rem and 7.5rem move with `--ui-scale`.
              */
-            cy.mountUI(bar({percent: 40, label: 'CPU Usage'}), {theme});
+            cy.mountUI(<Panel><div style={{width: 40}}><Progress percent={40} label='CPU'/></div></Panel>,
+                {theme});
 
             cy.get('.wrolpi-progress').should(($bar) => {
                 const style = getComputedStyle($bar[0]);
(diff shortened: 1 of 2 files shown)


app/src/components/ui/ui-layout.cy.js
```diff
@@ -799,6 +799,45 @@ describe('a progress bar has room for the label inside it', () => {
...
+                expect($bar[0].getBoundingClientRect().height, 'the bar is the height it was')
+                    .to.be.closeTo(1.375 * rem, 0.5);
+                expect(parseFloat(style.minWidth), 'and the width it was')
+                    .to.be.closeTo(7.5 * rem, 0.5);
+            });
+
+            // The Panel the bar sits on, which is what the track has to stay off.
+            cy.get('.wrolpi-progress').should(($bar) => {
+                const track = getComputedStyle($bar[0]).backgroundColor;
+                const panel = getComputedStyle($bar[0].closest('.wrolpi-panel')).backgroundColor;
```
lrnselfreliance: **[nit]** The width half of “the size it was with a border” reads `parseFloat(getComputedStyle(...).minWidth)` against `7.5 * rem`. That re-asserts the declared min-width (which never changed), not the laid-out outer box. With the border gone, a content-box regression would still report the same `minWidth` string, and a stretched bar in a Panel never exercises the min-width floor via geometry. Height via `getBoundingClientRect` is the load-bearing size check; width is not.

**Suggestion:** Either drop the min-width assertion (height + no-border already pin the PR), or mount the bar in a narrow wrapper and assert `getBoundingClientRect().width` is at least `7.5 * rem` so layout is what fails.

app/src/components/ui/ui.css
```diff
@@ -1160,14 +1160,19 @@ html .mantine-Notification-closeButton:hover {
...
+ * the bar exactly the size it was.
  */
 .wrolpi-progress {
     position: relative;
     height: 1.375rem;
     min-width: 7.5rem;
     overflow: hidden;
-    background: var(--bg);
-    border: 1px solid var(--border);
+    background: var(--head);
```
lrnselfreliance, line 1177: **[suggestion]** `--head` is a distinct token from `--panel` and `--bg` in all four themes, but the visual edge is thin where night/amber collapse the ramp. Against a Panel, night is ~1.05:1 (exactly the test floor). Against page `--bg` it is ~1.01:1 — effectively no track silhouette for a 0% / label-less bar. Real call sites put bars on both surfaces (Status/Downloads in Panels; Upload overall progress on the page). This is better than painting `--bg` on a Panel, and the border was carrying the edge before, but the comment’s “off both” overstates what the eye gets in monochrome themes.

**Suggestion:** Accept as a deliberate night/amber tradeoff (label still ≥4.5:1; border would reintroduce outline), or give empty tracks a hair more separation only in monochrome (e.g. a hairline/`--border` at low opacity) if 0% bars on Status/Downloads look missing. Optional: assert track vs `--bg` as well as vs the Panel so page-surface placements stay covered.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -799,6 +799,45 @@ describe('a progress bar has room for the label inside it', () => {
...
+                expect(track, 'the track paints a surface of its own')
+                    .to.not.match(/^rgba\(.*,\s*0(\.0+)?\)$/);
+                expect(contrast(toRgb(track), toRgb(panel)), 'track against the panel under it')
+                    .to.be.greaterThan(1.05);
+            });
+        });
+    });
+
     it('keeps a long label on one line', () => {
         /*
```
lrnselfreliance, line 879: **[nit]** Track-vs-panel uses `contrast(...).to.be.greaterThan(1.05)`. Night’s real ratio is ~1.053, so the assertion is true but has almost no slack for engine rounding and does not encode “visible edge” — only “not the same color.” That matches the PR’s intent and was clearly chosen on purpose; flagging only so a future theme tweak to `--panel` / `--head` does not get a free pass while the bar disappears.

**Suggestion:** Keep 1.05 if “not identical” is the claim; if “readable edge” is the claim, raise the floor or assert inequality of resolved colors plus a per-theme note that night is intentionally near-flat.

lrnselfreliance (commented):
## Summary

Small, well-scoped UI change: drop the progress-bar 1px border, keep outer size via the global `border-box` model, and let the track (`--head`) provide the empty-half edge so light no longer needs a `--panel` override. CSS is correct, prior progress work (label room, `overflow: hidden` band clip, reduced-motion fill, light fill brighten, night/amber muted fill) is untouched, and the existing per-theme label-vs-track 4.5:1 check still covers empty-track text. New Cypress coverage is real and theme-complete; only mild test-strength and night-edge visibility nits remain.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 2

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Fixed in ac5d479 — you're right that it re-asserted the declaration. `style.minWidth` is the same string whether or not the box it produces is correct, so a content-box regression would have sailed past it.

Took the second option: the bar now mounts in a `width: 40px` wrapper, so the min-width floor is what decides the layout, and `getBoundingClientRect().width` is what's asserted (`>= 7.5 * rem`). Checked it fails for the right reason — dropping `min-width` to 5rem gives `expected 88 to be at least 131.5` in all four themes.

app/src/components/ui/ui.css
lrnselfreliance: Agreed, and the comment was the wrong part — "off both in every theme" claimed something the monochrome themes don't give. Fixed in ac5d479, and it turned out to be half fixable rather than only documentable.

Measured every candidate track per theme (label = that theme's `--text` across the empty track):

| theme | track | vs Panel | vs page | label |
|---|---|---|---|---|
| amber | `--head` | 1.11 | 1.04 | 9.55 |
| amber | `--table-line` | **1.34** | **1.43** | 6.40 |
| night | `--head` | 1.05 | 1.01 | **5.25** |
| night | `--table-line` | 1.18 | 1.23 | 4.22 |
| night | `--border` | 1.46 | 1.52 | 3.42 |

So amber can afford a real edge and now takes `--table-line`. Night can't: every track with a visible edge drops the label crossing it below 4.5, and a hairline is the border this PR removes. Night's empty track is now documented as deliberately near-flat with those numbers in the CSS, instead of being described as an edge.

Also took the optional suggestion — the test now asserts the track against page `--bg` as well as the Panel, since Upload's overall bar sits on the page.

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Fixed in ac5d479. You're right that one floor couldn't be both claims at once — at 1.05 it read like a visibility assertion while only proving "not identical", and it had ~0.003 of slack in night.

Replaced with a per-theme floor table, so each theme is held to its own measured behavior instead of to the weakest one:

```js
const trackFloors = {
    light: {panel: 1.2, page: 1.1},
    dark:  {panel: 1.2, page: 1.08},
    amber: {panel: 1.3, page: 1.4},
    night: {panel: 1.04, page: 1.005},
};
```

Light, dark and amber now have real slack and would fail on a `--panel` / `--head` tweak that flattened them; night's 1.04 / 1.005 is explicitly the near-flat case, with the reason (label contrast) in the ui.css note beside it. Amber's floors are the ones that caught the regression when I probed it — reverting amber to `--head` gives `expected 1.1122 to be at least 1.3`.

PR #643, opened 2026-08-05
ui: three from the dashboard and the video pages

lrnselfreliance:
Three reported issues, one commit.

## The Download / Or / Upload panel

Semantic drew this as a `Segment placeholder` with a vertical `Divider`: two equal fields, a rule down the true middle, a button centred in each. The migrated version was a `Group` with a divider between two buttons, so the halves were sized to their labels — the pair huddled in the middle of the panel with the rule wherever they happened to leave it.

`.wrolpi-or-split` is a grid, so the halves stay equal whatever the labels say. The rule is the second column's `border-left` rather than an element between the halves — an element would be a third grid item and push the middle off centre by its own width. The "Or" is a disc painted with `--panel`, which masks the rule behind it; without it the line strikes through the word.

Below the tablet breakpoint the halves stack and the rule turns horizontal, which is what Semantic's `stackable` grid did.

## The video page's panels touching the left edge

The video route sits outside `VideosTabLayout`'s `PageContainer` deliberately, so the player can run to both edges of the window — that is the most viewing area a phone can give it and is worth keeping. Everything below the player inherited it too.

`.wrolpi-page-inset` gives the stack under the player the same inset `PageContainer` gives a page, at the same 699px breakpoint (fresnel's `tablet`, from contexts.js). Matching that number matters: below it a page has no side padding at all, so a phone would otherwise find the video page inset where every other page is not.

## The white duration chip in night

It was a literal `#ffffff` with black text. The poster under it *is* filtered by night mode — it is a `.media` leaf — but the chip is drawn beside the image rather than inside it, so the filter never reached it. A white tile on a card whose picture had gone dim red, and the brightest thing on the videos page in the theme whose whole point is that nothing is bright.

Tokens now (`--panel` / `--text` / `--border`), and opaque, because it sits on an arbitrary poster and has to stay legible over any of them.

## Guards

Nine, in `ui-layout.cy.js`, all verified to fail against the old code:

- each choice gets half the panel, and each half really is half the row — not two content-sized boxes that happen to match
- each button is centred in its own half
- the rule and its label fall at the true middle
- the halves stack below the breakpoint, with the rule turning horizontal
- the label is painted with the Panel's own color, so it masks rather than chips
- the stack under the player is inset on both sides, and the panel clears the edge the player touches
- the duration chip takes the surface and text of each of the four themes, and stays at 4.5:1
- and in night it is no brighter than the page it sits on — the report, stated as something measurable

One thing the guards turned up that is worth knowing beyond this PR: **cypress component tests default to a 500x500 viewport**, which is below the 699px breakpoint. Any viewport-keyed CSS is therefore exercised in its phone branch unless a test says otherwise — my first three assertions were measuring the stacked layout while claiming to measure the split one. The desktop tests now set the viewport explicitly and say why, and the stacked branch has a test of its own.

## Verified

1156 jest / 65 suites, 275 in ui-layout, 50 e2e, component suite at the pre-existing 12-failure baseline, clean build with no new lint warnings. The Or panel is in the gallery under its own section.

Not looked at in a browser. The reference for the Or panel was Semantic's documented `Segment placeholder` + vertical `Divider` shape rather than 10.0.0.8 itself, so if the QA box differs in a detail — disc size, how much vertical padding the fields carry — that is the thing to check.

### DIFF
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -551,9 +551,18 @@ describe('two choices of equal weight are split down the middle', () => {
             ).to.equal(0));
             expect(parseFloat(styles.borderRadius), 'and no disc').to.equal(0);
 
-            // Padded enough to actually open a gap, rather than sitting on the line unbroken.
+            // Padded enough to actually open a gap, rather than sitting on the line unbroken...
             expect(parseFloat(styles.paddingTop), 'padded enough to break the rule')
                 .to.be.greaterThan(0);
+            /*
+             * ...but its OWN padding, not a half's.  `.wrolpi-or-split > *` would lay this out as
+             * a half if it were not excluded, and it currently escapes on source order alone --
+             * same specificity, declared later.  Reordering the two blocks would quadruple the gap
+             * and an assertion that only read "greater than zero" would not notice.
+             */
+            const half = getComputedStyle($label[0].parentElement.children[0]);
+            expect(parseFloat(styles.paddingTop), "the word's own padding, not a half's")
+                .to.be.lessThan(parseFloat(half.paddingTop));
         });
     });
 });
@@ -2246,6 +2255,30 @@ describe('theme tokens actually resolve', () => {
     });
 });
 
+describe('the video page keeps its phone layout edge to edge', () => {
+    it('drops the inset below the tablet breakpoint, as every other page does', () => {
+        /*
+         * The other branch of `.wrolpi-page-inset`.  Below 699px a page has no side padding at all,
+         * so the video page must not either -- and with cypress defaulting to a 500px viewport this
+         * is the branch the rest of the suite runs in, which makes leaving it unasserted worse than
+         * it looks: a regression that padded phones would show up nowhere.
+         */
+        cy.viewport(400, 700);
+        cy.mountUI(<div className='wrolpi-stack wrolpi-page-inset'><Panel>About</Panel></div>);
+
+        cy.get('.wrolpi-page-inset').should(($inset) => {
+            const styles = getComputedStyle($inset[0]);
+            expect(parseFloat(styles.paddingLeft), 'no left inset on a phone').to.equal(0);
+            expect(parseFloat(styles.paddingRight), 'nor right').to.equal(0);
+
+            // And the panel really does reach the edge, not just declare that it may.
+            const panel = $inset[0].querySelector('.wrolpi-panel').getBoundingClientRect();
+            expect(panel.left, 'the panel reaches the edge')
+                .to.be.closeTo($inset[0].getBoundingClientRect().left, 0.5);
+        });
+    });
+});
+
 describe("a video's duration chip belongs to the theme, not to the poster", () => {
     /*
      * The chip was a literal `#ffffff` with black text.  The poster under it IS filtered by night
@@ -2267,6 +2300,16 @@ describe("a video's duration chip belongs to the theme, not to the poster", () =
                     .to.equal(panel.backgroundColor);
                 expect(chip.color, 'and the theme text').to.equal(panel.color);
 
+                /*
+                 * And an edge.  The chip is painted with `--panel`, so on a poster near that color
+                 * -- a pale sky in light, most screenshots -- the border is the only thing telling
+                 * the two apart.  Losing it leaves every contrast assertion above still green.
+                 */
+                expect(parseFloat(chip.borderTopWidth), 'has an edge').to.be.greaterThan(0);
+                expect(chip.borderTopColor, 'the theme border')
+                    .to.equal(toRgb(getComputedStyle(document.documentElement)
+                        .getPropertyValue('--border')));
+
                 // And still readable, which is what it is for: a length nobody can read is chrome.
                 expect(contrast(chip.color, chip.backgroundColor), 'legible')
                     .to.be.at.least(4.5);
@@ -2284,7 +2327,14 @@ describe("a video's duration chip belongs to the theme, not to the poster", () =
 
         cy.get('.duration-overlay').should(($chip) => {
             const chip = getComputedStyle($chip[0]);
-            const page = getComputedStyle(document.documentElement).backgroundColor;
+            /*
+             * `--bg` read as a token, not `html`'s computed background.  Tokens paint `--bg` on
+             * BODY, so `html` comes back `rgba(0, 0, 0, 0)` -- which `luminance` reads as black,
+             * making this "no brighter than black + slack" rather than the claim in its name.  It
+             * still failed the white chip, so it caught the reported bug by accident of the
+             * comparison rather than by measuring the thing it says it measures.
+             */
+            const page = toRgb(getComputedStyle(document.documentElement).getPropertyValue('--bg'));
             expect(luminance(chip.backgroundColor), 'no brighter than the page it sits on')
                 .to.be.at.most(luminance(page) + 0.05);
         });
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -433,6 +433,142 @@ describe('a row of buttons is spaced and aligned by the row', () => {
     });
 });
 
+describe('two choices of equal weight are split down the middle', () => {
+    /*
+     * The dashboard's Download/Upload panel, which Semantic drew as a `Segment placeholder` with a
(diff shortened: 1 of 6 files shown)


app/src/components/ui/ui.css
```diff
@@ -1336,6 +1336,101 @@ html[data-theme="amber"] .wrolpi-progress-fill {
...
+ * column's edge rather than as an element between them -- an element would be a third grid item
+ * and push the middle off centre by its own width.
+ */
+.wrolpi-or-split {
+    display: grid;
+    grid-template-columns: 1fr 1fr;
+    position: relative;
+}
+
+.wrolpi-or-split > * {
```
lrnselfreliance: **[suggestion]** `.wrolpi-or-split > *` still styles the absolutely positioned `.wrolpi-or-label` (display/flex/padding). The border rule below carefully excludes the label with `:not(.wrolpi-or-label)` because an earlier `> * + *` match drew a stray rule on the word; the same structural pitfall remains for the universal child rule. Today label padding wins only because `.wrolpi-or-label` and `.wrolpi-or-split > *` share specificity `(0,1,0)` and the label rule is later — reorder the blocks and the intended `0.375rem 0.5rem` becomes `1.5rem 1rem`. The “break the rule” test only asserts `paddingTop > 0`, so that regression would not fail CI.

**Suggestion:** Mirror the border selector: `.wrolpi-or-split > *:not(.wrolpi-or-label) { … }`. Optionally assert the label’s computed padding equals the authored values (or at least stays below the half padding).

app/src/components/ui/ui-layout.cy.js
```diff
@@ -2090,6 +2246,51 @@ describe('theme tokens actually resolve', () => {
...
+        /*
+         * The report, stated as something measurable.  In night the chip sat at full white against
+         * a panel around 4% luminance -- the specific complaint was "the timestamp is white".
+         */
+        cy.mountUI(<><Panel>Card</Panel><div className='duration-overlay'>12:34</div></>,
+            {theme: 'night', mediaFilter: 'night-red'});
+
+        cy.get('.duration-overlay').should(($chip) => {
+            const chip = getComputedStyle($chip[0]);
+            const page = getComputedStyle(document.documentElement).backgroundColor;
```
lrnselfreliance: **[suggestion]** The night “never the brightest thing” assertion reads `getComputedStyle(document.documentElement).backgroundColor`. Theme tokens paint `background: var(--bg)` on `body`, not `html`. `html` is typically transparent (`rgba(0,0,0,0)`), which `luminance()` treats as black. The check still fails the old white chip (luminance ~1), so it catches the original bug, but it is not measuring “no brighter than the page” — it is measuring against transparent/black with a generous `+ 0.05` slack. A future chip that is brighter than `--bg` but still dim could pass incorrectly relative to the claim.

**Suggestion:** Compare against `document.body` (or `getComputedStyle(...).getPropertyValue('--bg')` via the existing `toRgb` helper) so the assertion matches the documented intent.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -433,6 +433,162 @@ describe('a row of buttons is spaced and aligned by the row', () => {
...
+    });
+});
+
+describe('the video page insets what sits under its player', () => {
+    /*
+     * The video route is deliberately outside `VideosTabLayout`'s `PageContainer` so the player can
+     * run to both edges of the window -- the most viewing area a phone can give it.  Everything
+     * below the player inherited that, so the panels touched the left edge of the page.
+     */
+    it('pads the stack under the player, and nothing above it', () => {
```
lrnselfreliance, line 576: **[suggestion]** Video inset coverage only runs above the tablet breakpoint (`cy.viewport(1000, 700)`). The CSS deliberately zeros side padding at `max-width: 699px` to match `PageContainer`’s mobile branch. That mobile branch is commented but not asserted, so a regression that left `1em` padding on phones would not fail this suite.

**Suggestion:** Add a second case at width ≤699 that expects `paddingLeft`/`paddingRight` of `.wrolpi-page-inset` to be `0`, parallel to the Or-split stack test.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -2090,6 +2246,51 @@ describe('theme tokens actually resolve', () => {
...
+        it(`takes the surface and text of ${theme}`, () => {
+            cy.mountUI(<><Panel>Surface</Panel><div className='duration-overlay'>12:34</div></>,
+                {theme, mediaFilter: theme === 'night' ? 'night-red' : undefined});
+
+            cy.get('.duration-overlay').should(($chip) => {
+                const chip = getComputedStyle($chip[0]);
+                const panel = getComputedStyle(Cypress.$('.wrolpi-panel')[0]);
+
+                // The theme's surface, not a hardcoded one.
+                expect(chip.backgroundColor, 'painted with the theme surface')
```
lrnselfreliance, line 2299: **[nit]** Duration-chip theme tests assert `background`/`color` match the panel and WCAG contrast ≥4.5 across all themes (good). They do not assert the new `border: 1px solid var(--border)`, which is what keeps a `--panel`-colored chip readable when the poster is close to the panel color (especially light). Losing the border would still leave most contrast tests green.

**Suggestion:** Assert `borderTopWidth` (or all sides) > 0 and that border color resolves to the theme border (same pattern as other token tests in this file).

lrnselfreliance (commented):
## Summary

Three focused UI fixes land cleanly: a true equal-weight Download/Or/Upload split (CSS grid + center rule), video-page panel inset that mirrors `PageContainer`’s horizontal padding and the fresnel tablet cut at 700/699, and a duration chip that uses theme tokens instead of white-on-black. Breakpoints, token choices, and desktop viewport handling in Cypress are correct; gallery coverage for the Or split matches the dashboard. Remaining notes are maintainability and test precision, not functional regressions.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 2

## Issues outside the diff

- **[nit]** app/src/DashboardPage.js:18 — **[nit]** `Group` is still imported from `./components/ui` after the Download/Upload panel stopped using it (replaced by `.wrolpi-or-split`). Nothing else in this file references `Group` (`StatisticGroup` is a different export). Under CRA’s eslint this is an unused binding; depending on CI settings it is at least noise and can fail builds when warnings are treated as errors.
  **Suggestion:** Drop `Group` from the import list.

lrnselfreliance:
Five findings, no bugs, all valid — and three of them are tests that were not measuring their own claims, which is the theme of this round.

**`.wrolpi-or-split > *` still laid out the label.** The same structural pitfall the border rule was already fixed for, and I fixed one and left the other. The word escaped the half's `1.5rem` padding on source order alone — same specificity, declared later. Excluded now.

The test's `paddingTop > 0` would not have noticed, so it now asserts the word's padding is **smaller than a half's**. Verified against the actual regression rather than a proxy: dropping the `:not()` alone still passes, because source order saves it — it takes moving the label block above the halves to break, and then it fails with `expected 26.4 to be below 26.4`, the half's padding leaking onto the word.

**The night "never the brightest thing" assertion read `html`'s background.** Tokens paint `--bg` on `body`, so `html` came back `rgba(0, 0, 0, 0)` — luminance zero — making the check "no brighter than black, plus slack". It did fail the white chip, so it caught the reported bug, but by accident of the comparison rather than by measuring the page. It reads the `--bg` token now, through `toRgb`, which throws rather than silently yielding `NaN` if it ever stops resolving.

**The video inset was only asserted above the breakpoint.** The phone branch was commented and unmeasured — worse than it looks, because cypress defaults to a 500px viewport, so that is the branch everything else in the suite runs in. A regression that padded phones would have been the *only* thing the suite was rendering and still gone unreported. Now asserts zero padding and that the panel really reaches the edge.

**The duration chip's border went unasserted.** Fair, and it is the load-bearing part: the chip is painted `--panel`, so on a poster near that color the border is the only thing separating them. Asserted on all four themes against `--border`.

**`Group`** was left imported by the dashboard after the panel stopped using it. Dropped.

All three new assertions verified against the mutation each describes: mobile padding `expected 17.6 to equal 0`, chip border `expected 0 to be above 0` on each of the four themes, label padding as above.

1156 jest / 65 suites, 276 in ui-layout, 50 e2e, clean build with zero lint warnings.

One unrelated thing the mutation runs surfaced again: `takes the links back when the corner shrinks again` failed once with `expected 10 to be above 10`. That is the known latent test bug — a strict `>` on a count that can legitimately equal its bound — not anything in this PR. Still untouched, still worth fixing on its own.

PR #644, opened 2026-08-06
ui: a range slider is painted by the theme, not by the browser

lrnselfreliance:
The log level slider on Settings was blue in every theme, including night.

## Cause

A native `<input type="range">` is drawn by the browser, and with nothing said about it the UA uses its own accent — a fixed blue no theme can reach. `accent-color` appeared nowhere in the app, so **five** sliders had this: the log level on Settings, and the duration sliders in the food, water and ration calculators.

Night is where it stops being merely inconsistent. That theme exists so a user reading in the dark keeps their adaptation, and every token in it is a red — so the slider was the brightest non-red thing on the page, on the one control a user drags back and forth while watching it.

## Fix

`accent-color: var(--blue)` in `tokens.css`, beside the `a` rule it mirrors. Same token as links for the same reason: it remaps per theme, so night gets the dim red (`#8a2828`) and amber the amber (`#a86618`) with no second rule.

Checkboxes, radios and `<progress>` are included — they take the same accent and would have been the next report.

`accent-color` reaches the thumb and the **filled** half of the track and nothing else, so on a dark ground the rest stayed the UA's pale slab. `color-scheme: dark` handles that, scoped to these controls rather than declared on `html`: on `html` it would also govern scrollbars, spinners and every other piece of browser chrome, which is a larger change than this one and worth making deliberately rather than as a side effect.

## Tests

Five new cypress assertions — the accent equals `--blue` in each of the four themes, plus a negative: in the monochrome themes it is a resolved color and never `auto`. Without that last one the first four would pass just as well on an unstyled slider whose accent happened to stringify the same way. All five confirmed failing before the fix.

Added to `/theme-sample` under Form controls, which is how the monochrome themes actually get looked at. The gallery slider is stateful and starts mid-track — the accent only shows on the filled half, so one parked at either end would show almost none of the color the section is there to check.

## Verification

- jest 1157/1157
- cypress component 325/325
- `npm run build` clean

Not looked at in a browser; the color assertions are measured in a real one by cypress.

### DIFF
--- a/app/src/components/ThemeSamplePage.js
+++ b/app/src/components/ThemeSamplePage.js
@@ -78,6 +78,10 @@ const Row = ({children}) => <div style={{display: 'flex', gap: 10, flexWrap: 'wr
     {children}
 </div>;
 
+// The labels Settings shows beside its own log-level slider, so the sample reads like the
+// control it stands in for rather than a bare number.
+const logLevelNames = ['Critical', 'Error', 'Warning', 'Info', 'Debug'];
+
 /*
  * One navigation bar, in one of the twelve colors a user can pick for it.
  *
@@ -149,6 +153,9 @@ export function ThemeSamplePage() {
     const [wideConfirmOpen, setWideConfirmOpen] = useState(false);
     const [comments, setComments] = useState(true);
     const [hotspot, setHotspot] = useState(true);
+    // Draggable, because the accent only shows on the FILLED half of the track -- a slider
+    // parked at either end shows almost none of the color this section is here to check.
+    const [logLevel, setLogLevel] = useState(2);
     const [dismissed, setDismissed] = useState(false);
     const [modalOpen, setModalOpen] = useState(false);
     const [realColors, setRealColors] = useState(false);
@@ -707,6 +714,26 @@ export function ThemeSamplePage() {
                     <Toggle label='WROL Mode' description='Read-only: disables downloads.'/>
                     <Toggle label='Bluetooth' description='No adapter detected.' disabled/>
                 </div>
+                <div style={{marginTop: 14}}>
+                    <label htmlFor='sample_range'>Log Level: {logLevelNames[logLevel]}</label>
+                    <input
+                        type='range'
+                        id='sample_range'
+                        min={0}
+                        max={4}
+                        step={1}
+                        value={logLevel}
+                        onChange={e => setLogLevel(parseInt(e.target.value))}
+                        style={{width: '100%'}}
+                    />
+                </div>
+                <p style={{fontSize: '0.75rem', color: 'var(--muted)', marginTop: 8, marginBottom: 0}}>
+                    A range slider is drawn by the browser, not by us — so with nothing said
+                    about it the UA paints it in its own fixed blue, which no theme can reach.
+                    Check it in night: the accent follows <code>--blue</code> like a link does,
+                    and the unfilled half of the track is dark rather than the UA's pale slab.
+                    Settings and three of the calculators each have one.
+                </p>
             </Panel>
         </Section>
 
--- a/app/src/components/ThemeSamplePage.js
+++ b/app/src/components/ThemeSamplePage.js
@@ -78,9 +78,15 @@ const Row = ({children}) => <div style={{display: 'flex', gap: 10, flexWrap: 'wr
     {children}
 </div>;
 
-// The labels Settings shows beside its own log-level slider, so the sample reads like the
-// control it stands in for rather than a bare number.
-const logLevelNames = ['Critical', 'Error', 'Warning', 'Info', 'Debug'];
+/*
+ * The labels Settings shows beside its own log-level slider, so the sample reads like the
+ * control it stands in for rather than a bare number.
+ *
+ * Keyed 1..5 to match `levelNameMap` and the `<datalist>` in Settings.js exactly.  The first
+ * version of this invented an "Error" step, dropped Trace, and ran 0..4 -- a gallery whose
+ * job is to show the real control misrepresenting it is worse than not showing it.
+ */
+const logLevelNames = {1: 'Critical', 2: 'Warning', 3: 'Info', 4: 'Debug', 5: 'Trace'};
 
 /*
  * One navigation bar, in one of the twelve colors a user can pick for it.
@@ -153,9 +159,10 @@ export function ThemeSamplePage() {
     const [wideConfirmOpen, setWideConfirmOpen] = useState(false);
     const [comments, setComments] = useState(true);
     const [hotspot, setHotspot] = useState(true);
-    // Draggable, because the accent only shows on the FILLED half of the track -- a slider
-    // parked at either end shows almost none of the color this section is here to check.
-    const [logLevel, setLogLevel] = useState(2);
+    // Draggable, and starting at Info rather than at either end: the accent only shows on the
+    // FILLED half of the track, so a slider parked at a limit shows almost none of the color
+    // this section is here to check, or none of the unfilled track beside it.
+    const [logLevel, setLogLevel] = useState(3);
     const [dismissed, setDismissed] = useState(false);
     const [modalOpen, setModalOpen] = useState(false);
     const [realColors, setRealColors] = useState(false);
@@ -719,8 +726,8 @@ export function ThemeSamplePage() {
                     <input
                         type='range'
                         id='sample_range'
-                        min={0}
-                        max={4}
+                        min={1}
+                        max={5}
                         step={1}
                         value={logLevel}
                         onChange={e => setLogLevel(parseInt(e.target.value))}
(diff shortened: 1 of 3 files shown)


app/src/components/ThemeSamplePage.js
```diff
@@ -78,6 +78,10 @@ const Row = ({children}) => <div style={{display: 'flex', gap: 10, flexWrap: 'wr
     {children}
 </div>;
 
+// The labels Settings shows beside its own log-level slider, so the sample reads like the
+// control it stands in for rather than a bare number.
+const logLevelNames = ['Critical', 'Error', 'Warning', 'Info', 'Debug'];
```
lrnselfreliance: **[suggestion]** The sample claims the labels Settings shows beside its log-level slider, but `logLevelNames` is `['Critical', 'Error', 'Warning', 'Info', 'Debug']` on `min={0}` / `max={4}`. Settings actually maps `1..5` to Critical, Warning, Info, Debug, Trace (`levelNameMap` in `Settings.js`) — there is no Error step and Trace is missing. The control still demonstrates accent/track paint, but a theme-verification gallery that misrepresents the real control undercuts its purpose.

**Suggestion:** Align the sample with Settings: labels Critical / Warning / Info / Debug / Trace, range `min={1}` `max={5}`, and a default in the middle (e.g. `3` for Info) so filled and unfilled track both show.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -2667,6 +2667,58 @@ describe('a card reads as titles, not as a wall of links', () => {
...
+            cy.get('[data-testid="slider"]').should(($slider) => {
+                const expected = toRgb(getComputedStyle(document.documentElement)
+                    .getPropertyValue('--blue'));
+                expect(toRgb(getComputedStyle($slider[0]).accentColor), `accent in ${theme}`)
+                    .to.equal(expected);
+            });
+        });
+    });
+
+    it('never leaves it at the UA default, which is what night could not tolerate', () => {
```
lrnselfreliance: **[suggestion]** The negative test is weaker than its comment claims. It only asserts `accent !== 'auto'` and that `toRgb(accent)` does not throw. Spec-wise the computed value of `accent-color: auto` is the keyword `auto`, but if an engine ever reports a resolved system blue as `rgb(...)`, this test still passes while the slider is the UA default. The positive per-theme equality checks against `--blue` are the real safety net (and do protect night/amber), so the regression this block names is only partly guarded. It also only runs `monochromeThemes`, not light/dark, and stacks both themes inside one `it` via `forEach` (unlike the rest of this file, which emits one `it` per theme).

**Suggestion:** Assert the resolved accent equals `--blue` for each monochrome theme (or all themes) as the explicit “not the UA default” check, or at least assert inequality with a known system-blue sentinel if you keep the `auto` wording. Prefer one `it` per theme for clear failures.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -2667,6 +2667,58 @@ describe('a card reads as titles, not as a wall of links', () => {
...
+     *
+     * `accent-color` takes `--blue` for the same reason links do: it remaps per theme, so
+     * night gets the dim red and amber the amber with no second rule.
+     */
+    const mountSlider = (theme) => cy.mountUI(
+        <input type='range' min={1} max={5} defaultValue={3} data-testid='slider'/>,
+        theme ? {theme} : undefined,
+    );
+
+    themeNames.forEach((theme) => {
```
lrnselfreliance, line 2691: **[suggestion]** Half of the fix is `color-scheme: dark` on range (and checkbox/radio) under dark/night/amber so the unfilled track is not a pale UA slab. The new suite only reads `accentColor`. Dropping or regressing the `color-scheme` block would leave the track wrong in night with no failing test — exactly the visual failure the ThemeSamplePage copy asks a reviewer to check by eye.

**Suggestion:** For dark themes, assert `getComputedStyle(slider).colorScheme === 'dark'` (and for light, that it is not forced dark). Optional: screenshot/visual check is overkill if computed `color-scheme` is asserted.

app/src/themes/tokens.css
```diff
@@ -318,6 +318,50 @@ a:has(img, svg, .wrolpi-card-icon):hover {
...
+ *
+ * Night is where that is not merely inconsistent.  The theme exists so a user reading in
+ * the dark keeps their adaptation and every token is a red, so a slider arriving in the
+ * browser's blue was the brightest non-red thing on the page -- on a control the user drags
+ * back and forth while watching it.
+ *
+ * `--blue` for the same reason links take it: it remaps per theme, so night gets the dim
+ * red and amber the amber without a second rule.
+ */
+input[type="range"],
```
lrnselfreliance, line 353: **[suggestion]** The rule set applies `accent-color` (and, for dark themes, `color-scheme`) to `input[type="checkbox"]`, `input[type="radio"]`, and `progress` as well as range. That matches the four element types `accent-color` supports, and side effects in the product UI should be small: product checkboxes/radios are Mantine (not native UA paint), and Progress is a custom div, not `<progress>`. So the range fix is real; the checkbox/radio/progress lines are mostly unused defensive CSS. Comments still read as if native checkboxes and radios are on screen the way sliders are. `progress` also gets no `color-scheme: dark`, so if a native progress bar ever appears on a dark theme its unfilled track could still be light.

**Suggestion:** Either narrow the selectors to `input[type="range"]` (the stated bug) or keep the broader accent set but adjust the comment to say these are the UA-accented control types and that product checkboxes/radios/progress are already custom. If keeping `progress`, consider the same dark `color-scheme` treatment for consistency.

app/src/themes/tokens.css
```diff
@@ -318,6 +318,50 @@ a:has(img, svg, .wrolpi-card-icon):hover {
     text-decoration: none;
 }
 
+/*
+ * Native controls the browser paints itself.
+ *
+ * A range slider, a checkbox and a radio are drawn by the UA, and with nothing said about
+ * them the UA uses its own accent -- a fixed blue no theme can reach.  Five sliders are on
```
lrnselfreliance: **[nit]** Comment says “Five sliders are on screen” then lists Settings plus food/water/ration calculators — four controls. The same “Five” wording appears in `ui-layout.cy.js`. Theme-sample adds a fifth only on `/theme-sample`, not in normal app use.

**Suggestion:** Say four product sliders (Settings + three calculators), or five if counting the sample explicitly.

app/src/themes/tokens.css
```diff
@@ -318,6 +318,50 @@ a:has(img, svg, .wrolpi-card-icon):hover {
...
+ * red and amber the amber without a second rule.
+ */
+input[type="range"],
+input[type="checkbox"],
+input[type="radio"],
+progress {
+    accent-color: var(--blue);
+}
+
+/*
```
lrnselfreliance: **[nit]** Track/thumb paint still depends on UA internals. `accent-color` covers thumb + filled portion; `color-scheme` is a best-effort fix for the unfilled track without `::-webkit-slider-runnable-track` / `::-moz-range-track` rules. That is an acceptable progressive-enhancement tradeoff and is documented in the comment, but appearance can still vary by engine (especially older Safari where `accent-color` support landed later). Production `browserslist` is broad (`>0.2%`), so some clients will keep the fixed UA blue.

**Suggestion:** No change required for merge if the team accepts progressive enhancement. If night-vision purity is non-negotiable on all supported browsers, a small custom range (or limited `::-webkit-` / `::-moz-` track/thumb rules using `--blue` / `--border`) would be the complete fix.

lrnselfreliance (commented):
## Summary

The core fix is correct: `accent-color: var(--blue)` remaps native control accents through the same per-theme token links already use, and scoping `color-scheme: dark` to dark-theme form controls (not `html`) is the right way to darken the unfilled range track without retinting scrollbars and other chrome. Product-facing range sliders (Settings + three calculators) will pick this up with no call-site changes. Remaining issues are gallery label drift, missing coverage for the `color-scheme` half of the fix, a weak negative accent test, and comments that slightly overstate slider count and native checkbox/radio usage.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 2

lrnselfreliance:
Triaged — 0 bugs, 4 suggestions, 2 nits. All acted on, and one of them turned up something larger than the suggestion itself.

## The `color-scheme` suggestion found a redundant rule, not a missing test

The ask was to cover the `color-scheme: dark` half, since dropping it would leave night's track wrong with nothing failing. I wrote those tests, they passed — then I mutated the rule to match nothing, and **they stayed green**.

Mantine's stylesheet puts `color-scheme: var(--mantine-color-scheme)` on `:root`, and `ThemeProvider` drives it with `forceColorScheme`. The property is inherited, so a range slider on a dark theme has had a dark track all along. My rule changed no pixel, and my "pale UA slab" claim was wrong — that was never the state. The rule is deleted and the claim is out of the comments and the gallery copy.

The tests stay, re-aimed at what actually supplies it: breaking `forceColorScheme` fails all three. Worth guarding, because it lives in another module and every accent assertion passes without it.

## Gallery label drift — fixed

Correct, and worse than drift: it invented an "Error" step, dropped Trace, and ran `0..4` where Settings maps `1..5` onto Critical / Warning / Info / Debug / Trace. Now keyed to `levelNameMap` exactly, starting at Info so both halves of the track show.

## Weak negative test — replaced

Agreed it was weaker than its comment, and it was also redundant: `toRgb` throws on anything unresolved and the computed value of `accent-color: auto` is the keyword, so the four positive tests already fail on an unstyled slider. I've noted that where they are, and replaced the negative with a constraint they genuinely don't cover — `accent-color` is inherited, so a text field beside the slider must still read `auto`. That's what says the accent comes from this rule rather than from a tinted document.

## checkbox/radio/progress — comment corrected, selectors kept

Right that these are defensive: the app's checkboxes and radios are Mantine's and `Progress` is a div. Kept for the day a native one appears, but the comment now says exactly that instead of implying they're on screen. The `progress` inconsistency you flagged resolved itself when the `color-scheme` block went away.

## Nits

Four product sliders, not five — the fifth is the sample. Corrected in both places. And the progressive-enhancement limit is now stated where the rule is, including what a complete fix would take (`::-webkit-slider-runnable-track` / `::-moz-range-track`). Accepting the tradeoff for now.

## Verification

jest 1157/1157, cypress component 329/329, `npm run build` clean. The three `color-scheme` tests were mutation-verified against `forceColorScheme` rather than against my own rule.

PR #645, opened 2026-08-07
fix: flush the duplicate's deletion before renaming onto its path

lrnselfreliance:
## Problem

A video download on the docker WROLPi died at commit with:

```
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: file_group.primary_path
[SQL: UPDATE file_group SET files=?, primary_path=?, a_text=? WHERE file_group.id = ?]
```

`VideoDownloader.finalize_download` calls `Video.delete_duplicate_videos`, which:

1. finds the copy occupying the destination path and deletes it — `FileGroup.delete` unlinks the files immediately, but only queues `session.delete` for the row;
2. calls `FileGroup.move()` on the copy it keeps. The `new_primary_path.exists()` guard passes, because the files are already gone from disk.

At flush time SQLAlchemy emits saves before deletes for a mapper, so the rename's `UPDATE ... primary_path` runs while the doomed row still holds that path.

The failure rolls back the entire download transaction: the freshly downloaded video never gets a DB record, and the previous copy's files are already unlinked. It triggers whenever a re-download's destination filename is held by a *smaller* duplicate — the equal-size and largest-at-destination cases take other branches, which is why existing tests didn't catch it. The earlier missing-file deletions in the same function are safe because the `get_by_path` query between them autoflushes.

## Fix

`session.flush()` after deleting the video at the destination, before the rename claims its path.

## Test

`test_delete_duplicate_video_smaller_copy_at_destination` seeds a smaller copy at the destination path and a larger one beside it. Without the flush it fails with the exact production `IntegrityError`; with it, the larger video is kept and renamed onto the path.

`modules/videos` and `wrolpi/files` suites pass (564 passed, 1 skipped).

### DIFF
--- a/modules/videos/models.py
+++ b/modules/videos/models.py
@@ -533,6 +533,11 @@ class Video(ModelHelper, Base):
         existing_video = Video.get_by_path(session, video_path)
         if existing_video and largest_video != existing_video:
             delete_video(existing_video)
+            # `FileGroup.delete` unlinks the files immediately, but its DELETE is only queued on the session.  The
+            # rename below claims `video_path` as its `primary_path`, and SQLAlchemy emits UPDATEs before DELETEs in a
+            # flush -- so without this flush the UPDATE hits the not-yet-deleted row and violates the unique index on
+            # `file_group.primary_path`, rolling back the entire download.
+            session.flush()
             for idx, video in enumerate(duplicate_videos):
                 if video == existing_video:
                     duplicate_videos.pop(idx)
--- a/modules/videos/models.py
+++ b/modules/videos/models.py
@@ -533,17 +533,17 @@ class Video(ModelHelper, Base):
         existing_video = Video.get_by_path(session, video_path)
         if existing_video and largest_video != existing_video:
             delete_video(existing_video)
-            # `FileGroup.delete` unlinks the files immediately, but its DELETE is only queued on the session.  The
-            # rename below claims `video_path` as its `primary_path`, and SQLAlchemy emits UPDATEs before DELETEs in a
-            # flush -- so without this flush the UPDATE hits the not-yet-deleted row and violates the unique index on
-            # `file_group.primary_path`, rolling back the entire download.
-            session.flush()
             for idx, video in enumerate(duplicate_videos):
                 if video == existing_video:
                     duplicate_videos.pop(idx)
 
         # Rename the video, add any missing tag names.
         if largest_video.video_path != video_path:
+            # Every deletion above unlinked its files immediately, but only queued its DELETE on the session.  The
+            # rename claims `video_path` as its `primary_path`, and SQLAlchemy emits UPDATEs before DELETEs within a
+            # flush -- so without this flush the UPDATE can hit a not-yet-deleted row and violate the unique index on
+            # `file_group.primary_path`, rolling back the entire download.
+            session.flush()
             largest_video.file_group.move(video_path)
         for tag_name in tag_names:
             if tag_name not in largest_video.file_group.tag_names:
--- a/modules/videos/test/test_video.py
+++ b/modules/videos/test/test_video.py
@@ -238,6 +238,49 @@ async def test_delete_duplicate_video_tagged_duplicate(async_client, test_sessio
 
 
 @pytest.mark.asyncio
+async def test_delete_duplicate_video_smaller_copy_at_destination(async_client, test_session, channel_factory,
+                                                                  video_factory):
+    """The kept Video can be renamed onto the path of the smaller duplicate that was just deleted."""
+    channel = channel_factory(name='Channel Name')
+    video_path = channel.directory / f'{channel.name}_20000101_ABC123456_The video title.mp4'
+
+    channel.info_json = {'entries': [{'id': 'ABC123456', 'title': 'The video title'}]}
+    entry = channel.info_json['entries'][0]
+
+    # The smaller copy already occupies the destination path.
+    small = video_factory(
+        channel_id=channel.id,
+        title=f'{channel.name}_20000101_ABC123456_The video title',
+        upload_date=datetime(2000, 1, 1, tzinfo=timezone.utc),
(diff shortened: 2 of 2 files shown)


modules/videos/models.py
```diff
@@ -533,6 +533,11 @@ def delete_video(video_: Video):
         existing_video = Video.get_by_path(session, video_path)
         if existing_video and largest_video != existing_video:
             delete_video(existing_video)
+            # `FileGroup.delete` unlinks the files immediately, but its DELETE is only queued on the session.  The
+            # rename below claims `video_path` as its `primary_path`, and SQLAlchemy emits UPDATEs before DELETEs in a
+            # flush -- so without this flush the UPDATE hits the not-yet-deleted row and violates the unique index on
+            # `file_group.primary_path`, rolling back the entire download.
+            session.flush()
```
lrnselfreliance: **[suggestion]** `session.flush()` is only called after deleting `existing_video` found via `Video.get_by_path`. Earlier in the same function, videos whose files are already gone are deleted and removed from `duplicate_videos` with no flush. SQLAlchemy filters session-deleted instances out of subsequent queries, so `get_by_path(session, video_path)` can return `None` even though the row still occupies `primary_path` until flush. With ≥2 remaining on-disk duplicates and the largest not at `video_path`, `move` still runs and the same UPDATE-before-DELETE IntegrityError can fire on commit. Scenario: stale DB row at destination (file unlinked), plus two real copies elsewhere, largest needs rename onto dest.

**Suggestion:** Flush immediately before the rename instead of (or in addition to) only after the `existing_video` delete, e.g.:

```python
if largest_video.video_path != video_path:
    # Pending DELETEs must hit the DB before this UPDATE reuses primary_path.
    session.flush()
    largest_video.file_group.move(video_path)
```

That covers both the smaller-at-destination path fixed here and missing-file-at-destination. Optionally add a small test that unlinks the small file before calling `delete_duplicate_videos` while keeping two other on-disk copies.

modules/videos/test/test_video.py
```diff
@@ -237,6 +237,49 @@ async def test_delete_duplicate_video_tagged_duplicate(async_client, test_sessio
...
+    large.file_group.size = small.file_group.size + 1
+    large_id = large.id
+
+    test_session.commit()
+    assert small.video_path == video_path
+
+    assert await Video.delete_duplicate_videos(test_session, 'https://example.com/video', entry['id'], video_path)
+    # The commit must not raise IntegrityError: the deleted duplicate still held `video_path` when the kept Video's
+    # `primary_path` was updated to it.
+    test_session.commit()
```
lrnselfreliance, line 276: **[nit]** The regression test is real (size skew forces keep-and-rename; final `commit()` is what would raise `IntegrityError` without the fix) and is not mock theater. Minor gaps only: it does not assert that the destination file on disk is the large video’s content (only that a file exists at the path after rename), and it does not cover equal-size-but-newer-at-destination (mtime sort keeps oldest; if oldest is not at dest and sizes equal, same code path as size skew). Not required for this PR’s claim.

**Suggestion:** Optional follow-ups only—assert `large`’s original path is gone after the call, and/or a third case with equal sizes where the newer copy sits at `video_path` so the older kept copy must rename onto it.

modules/videos/models.py
lrnselfreliance: **[nit]** Transaction/rollback safety is unchanged in kind: `FileGroup.delete` already unlinks files on disk before any flush/commit, so a later failure still leaves DB rolled back with files gone. The new `flush()` only makes the DELETE durable within the open transaction earlier; it does not introduce a new disk/DB split beyond the pre-existing pattern. Identity-map impact is fine: flush does not expire live instances the way commit does.

**Suggestion:** No code change required for this PR. If hardening later, consider documenting that callers must treat `delete_duplicate_videos` as non-rollback-safe for on-disk state (files already unlinked).

lrnselfreliance (commented):
## Summary

The fix correctly addresses a real SQLAlchemy unit-of-work ordering bug: after `delete_video` queues a DELETE for the smaller FileGroup at `video_path`, `FileGroup.move` dirty-sets `primary_path` on the kept row, and without an intervening flush the eventual UPDATE runs before the DELETE and violates the unique index on `file_group.primary_path`. The new regression test forces the larger-not-at-destination layout that older equal-size tests never hit, and exercises a real commit rather than mocks. Residual gap: missing-file deletes earlier in the same function can free the destination path without a flush before `move`, so a slightly broader flush placement would be more complete; otherwise this is a solid, minimal PR.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 2

lrnselfreliance:
Triaged all three.

**[suggestion] flush placement** — taken, but the second bug it describes doesn't exist as stated. The premise is that SQLAlchemy filters session-deleted instances out of subsequent queries; with the default `autoflush=True` (these sessions use a plain `sessionmaker(bind=_engine)`, so autoflush is on) the `get_by_path` query *flushes* those earlier deletes instead of hiding them, so the row is really gone from the DB before `move` runs. I built the described layout as a scratch test — stale row at the destination with its file unlinked, two real copies elsewhere, largest needing the rename — and it passes on the previous commit.

Still moved the flush to immediately before `FileGroup.move` anyway: the rename shouldn't depend on a query happening to sit between the deletes and itself. It now covers every deletion queued in the function. Re-verified the regression test fails with the exact `UNIQUE constraint failed: file_group.primary_path` when the flush is removed from its new position.

**[nit] test gaps** — added `assert not large_old_path.exists()` (reads a move, not a copy) plus the directory-contents check the sibling tests use. Skipped the equal-size-newer-at-destination case: as the comment notes, it lands on the same branch as the size skew, so it would assert the same code twice.

**[nit] rollback safety** — agreed, no change. `FileGroup.delete` unlinking before commit is pre-existing and out of scope here; worth documenting separately rather than growing this PR.

PR #646, opened 2026-08-08
db: make write intent take the write lock, and stop holding it for two minutes

lrnselfreliance:
Fixes the `database is locked` failures seen in production — including the one from creating a download during a refresh.

## What was failing

Three distinct sites in the API logs, all within a few minutes of each other:

| When | Site | Statement |
|---|---|---|
| within the window (x6) | `upsert_file` | `INSERT INTO file_group` (uploads) |
| same window | `post_download` → `create_downloads` | `INSERT INTO download` |
| a later run | `renew_recurring_downloads` | `UPDATE download SET status` |

## Cause 1 — deferred read-then-write

A transaction that begins as a reader and upgrades at its first write is refused **instantly** when another connection wrote in between: SQLite skips `busy_timeout` for lock upgrades, since waiting cannot refresh a stale snapshot. `get_immediate_db_session` existed to avoid this, but an opt-in helper only works if every author remembers it — and read-then-write is the ordinary shape of a write, so the three sites above forgot.

**The helper is gone.** `get_db_session(commit=True)` — the existing declaration of write intent — now begins `IMMEDIATE`. [Rails 7.1](https://fractaledmind.com/2023/12/11/sqlite-on-rails-improving-concurrency/) and Django 5.1 (`"transaction_mode": "IMMEDIATE"`) made the same call for the same reason: default the transaction, don't ask application code to opt in. Read sessions (`commit=False`) stay deferred, so WAL read-concurrency is untouched.

Reproduced first as a standalone script (a deferred read-then-write against a real `create_wrolpi_engine` DB, with another connection doing short `BEGIN IMMEDIATE` writes: **30/30 failures**, all instant), then as the regression test below.

## Cause 2 — a two-minute write transaction

Making IMMEDIATE the default is only safe if write transactions are short. `video_cleanup` ran its entire Channel-claiming join — `UPDATE video ... FROM channel ... LEFT JOIN file_group ON fg.directory LIKE col.directory || '/%'`, O(channels × file_groups) — inside one write transaction. It appears **eight times in the log at 110–130 seconds each**, on a library with ~1500 channels:

```
[WARNING] After refresh cleanup video_cleanup took 130.45688 seconds
```

That is well past the 30s `busy_timeout`, so while it ran *every* writer failed — including correctly-written IMMEDIATE ones. The failures above sit inside its 130-second window.

The join is now a read (WAL readers never block writers) and only the resulting UPDATEs take the lock, 500 rows per transaction. `claim_videos_for_channels` had the same join and the same fix — channel import calls it with every channel in `channels.yaml`, which is what the 08-08 failure was concurrent with.

## Also

- `renew_recurring_downloads` was a read session with an inner commit; now a write session. Its computed `next_download` values persist even when nothing was renewed — which is the point of computing them.
- `POST /api/download` no longer writes through `request.ctx.session`. That session is committed by the *response* middleware, so writing through it holds the lock for the whole request; a write session commits before the handler returns, ahead of the background dispatch `create_downloads` fires. The middleware session stays deferred deliberately, and says why.
- `_immediate_txn` is a `ContextVar` rather than a `threading.local`: the writers that set it include coroutines, and one asyncio thread runs many interleaved.
- `write_lock_held_briefly` moved into `conftest.py` — two suites need it now.

## Tests

Both new tests were verified by mutation — reverting either site alone fails its own test:

- `test_mutating_request_survives_concurrent_writer` — POST `/api/download` while another connection holds the write lock. Reverted, it fails with the production error verbatim: `(sqlite3.OperationalError) database is locked [SQL: INSERT INTO download (url, ...)]`.
- `test_video_cleanup_claims_outside_a_write_transaction` — asserts the claiming join does not run inside a `commit=True` cursor, and that the claims are still written.

`test_plain_write_session_stays_deferred` was deleted: it asserted the behavior this PR deliberately inverts.

Full backend suite: **1709 passed, 5 skipped**.

## Not addressed

Handlers other than `post_download` still write through the deferred request session. Arming the request session for every mutating method would fix them wholesale, but it holds the write lock across the entire request — including chunked uploads writing to disk — so it wants its own change, with the slow handlers moved off the request transaction first.

### DIFF
--- a/modules/videos/__init__.py
+++ b/modules/videos/__init__.py
@@ -76,38 +76,98 @@ async def video_modeler(progress_callback: Callable[[int], None] = None):
         await asyncio.sleep(0)
 
 
+# Rows written per transaction when claiming Videos for their Channels.  Small enough that the
+# write lock is held for milliseconds at a time; large enough that a big library does not pay a
+# transaction per row.
+CLAIM_CHUNK_SIZE = 500
+
+
+def _chunks(items: list, size: int = CLAIM_CHUNK_SIZE):
+    for start in range(0, len(items), size):
+        yield items[start:start + size]
+
+
+def _first_claim_per_video(rows) -> List[Tuple[int, int]]:
+    """One `(video_id, channel_id)` per Video.
+
+    A Video nested under two Channel directories matches both; the single `UPDATE ... FROM` this
+    replaced also picked one arbitrarily, so keep the first row and drop the rest rather than
+    writing the same Video twice."""
+    claims = dict()
+    for video_id, channel_id in rows:
+        claims.setdefault(video_id, channel_id)
+    return list(claims.items())
+
+
+def _claim_videos(claims: List[Tuple[int, int]]):
+    """Assign `(video_id, channel_id)` pairs, a chunk per write transaction.
+
+    The pairs are computed by a *read* — WAL readers never block writers — and only these short
+    UPDATEs take the write lock.  Doing the whole thing as one `UPDATE ... FROM channel` held the
+    lock for the entire join: 110-130 seconds per refresh on a library with ~1500 channels, which
+    is well past the 30s `busy_timeout`, so every other writer on the box failed with "database is
+    locked" for two solid minutes.
+
+    `channel_id IS NULL` is re-checked in the UPDATE because the read ran in an earlier
+    transaction: a download may have claimed the Video since, and it should win.
+    """
+    if not claims:
+        return
+    claimed = 0
+    for chunk in _chunks(claims):
+        with get_db_curs(commit=True) as curs:
+            curs.executemany('UPDATE video SET channel_id = ? WHERE id = ? AND channel_id IS NULL',
+                             [(channel_id, video_id) for video_id, channel_id in chunk])
+        claimed += len(chunk)
+    logger.debug(f'Claimed {claimed} Videos for their Channels')
+
+
+# Videos in a Channel's directory (or below it) that no Channel has claimed yet.
+_UNCLAIMED_VIDEOS_SQL = '''
+                        SELECT v.id, c.id
+                        FROM video v
+                                 INNER JOIN file_group fg ON fg.id = v.file_group_id
+                                 INNER JOIN collection col ON fg.directory = col.directory
+                            OR fg.directory LIKE col.directory || '/%'
+                                 INNER JOIN channel c ON c.collection_id = col.id
+                        WHERE v.channel_id IS NULL
+                        '''
+
+
 @register_refresh_cleanup
 @limit_concurrent(1)
 def video_cleanup():
     logger.info('Claiming Videos for their Channels')
-    with get_db_curs(commit=True) as curs:
-        # Delete all Videos if the FileModel no longer contains a video or audio file.
+    # Read the FileGroups that are no longer video/audio, then unmodel them in short transactions.
+    with get_db_curs() as curs:
         curs.execute('''
-                     UPDATE file_group
-                     SET model = NULL
+                     SELECT id
+                     FROM file_group
                      WHERE model = 'video'
                        AND mimetype NOT LIKE 'video/%'
                        AND mimetype NOT LIKE 'audio/%'
-                     RETURNING id
                      ''')
-        deleted_ids = [i['id'] for i in curs.fetchall()]
-        if deleted_ids:
+        stale_ids = [i['id'] for i in curs.fetchall()]
+    for chunk in _chunks(stale_ids):
+        ids = json.dumps(chunk)
+        with get_db_curs(commit=True) as curs:
+            curs.execute('''
+                         UPDATE file_group
+                         SET model = NULL
+                         WHERE id IN (SELECT value FROM json_each(:ids))
+                         ''', dict(ids=ids))
             curs.execute('''
                          DELETE
                          FROM video
                          WHERE file_group_id IN (SELECT value FROM json_each(:ids))
-                         ''', dict(ids=json.dumps(deleted_ids)))
-        # Claim all Videos in a Channel's directory for that Channel.  But, only if they have not yet been claimed.
-        curs.execute('''
-                     UPDATE video AS v
-                     SET channel_id = c.id
-                     FROM channel c
-                              INNER JOIN collection col ON col.id = c.collection_id
-                              LEFT JOIN file_group fg ON fg.directory = col.directory
-                                                      OR fg.directory LIKE col.directory || '/%'
-                     WHERE v.channel_id IS NULL
-                       AND fg.id = v.file_group_id
-                     ''')
+                         ''', dict(ids=ids))
+
+    # Claim all Videos in a Channel's directory for that Channel.  But, only if they have not yet
+    # been claimed.  The join is the slow part, so it runs as a read; only the writes take the lock.
+    with get_db_curs() as curs:
+        curs.execute(_UNCLAIMED_VIDEOS_SQL)
+        claims = _first_claim_per_video(curs.fetchall())
+    _claim_videos(claims)
 
 
 
@@ -126,15 +186,13 @@ def claim_videos_for_channels(channel_ids: List[int]):
         return
 
     logger.info(f'Claiming Videos for {len(channel_ids)} channel(s)')
-    with get_db_curs(commit=True) as curs:
-        curs.execute('''
-                     UPDATE video AS v
-                     SET channel_id = c.id
-                     FROM channel c
-                              INNER JOIN collection col ON col.id = c.collection_id
-                              LEFT JOIN file_group fg ON fg.directory = col.directory
-                                                      OR fg.directory LIKE col.directory || '/%'
-                     WHERE v.channel_id IS NULL
-                       AND fg.id = v.file_group_id
+    # Read the claims, then write them in chunks -- the join must not hold the write lock (see
+    # `_claim_videos`).  Channel import calls this with every channel in channels.yaml, so it is
+    # just as slow as the full `video_cleanup` sweep.
+    with get_db_curs() as curs:
+        curs.execute(f'''
+                     {_UNCLAIMED_VIDEOS_SQL}
                        AND c.id IN (SELECT value FROM json_each(:channel_ids))
                      ''', dict(channel_ids=json.dumps(list(channel_ids))))
+        claims = _first_claim_per_video(curs.fetchall())
+    _claim_videos(claims)
--- a/modules/videos/__init__.py
+++ b/modules/videos/__init__.py
@@ -1,10 +1,12 @@
 import asyncio
 import json
+import pathlib
 from typing import Callable, List, Tuple
 
 from sqlalchemy import or_
 from sqlalchemy.orm import Session
 
+from modules.videos.common import get_or_create_ffprobe_json
 from modules.videos.models import Video
 from wrolpi.common import logger, limit_concurrent, register_modeler, register_refresh_cleanup
 from wrolpi.db import get_db_curs, get_db_session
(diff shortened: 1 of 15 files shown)


wrolpi/db.py
```diff
@@ -184,8 +189,22 @@ def get_db_context() -> Tuple[sqlalchemy.engine.Engine, Session]:
 def get_db_session(commit: bool = False) -> Generator[Session, Any, None]:
     """
     Context manager that creates a DB session.  This will automatically rollback changes, unless `commit` is True.
+
+    `commit=True` declares write intent, so the transaction begins as a writer (`BEGIN IMMEDIATE`)
```
lrnselfreliance, line 193: **[bug]** Making every `get_db_session(commit=True)` take `BEGIN IMMEDIATE` is the correct *systemic* multi-process fix (SQLite skips `busy_timeout` on mid-transaction SHARED→RESERVED upgrades; Rails 7.1 / Django 5.1 / SQLite docs all converge on IMMEDIATE for write intent). But several existing long `commit=True` call sites were safe *only* while deferred (write lock taken at first dirty flush). The worst is `video_modeler` (`modules/videos/__init__.py`): it opens `get_db_session(commit=True)`, then `await video.get_ffprobe_json()` for up to 20 videos. After this PR the first `session.query` issues IMMEDIATE, so the write lock is held across all ffprobe subprocess work for the batch — often past the 30s `busy_timeout`, recreating the multi-minute writer starvation this PR is curing in `video_cleanup`. The PR docstring even says to do ffprobe outside a session; that call site was not updated.

**Suggestion:** Before merging global IMMEDIATE, split modelers (and any other `commit=True` + slow I/O paths) the same way as claim: read ids deferred, run ffprobe outside any session, short `commit=True` only for ORM writes. Search for `get_db_session(commit=True)` paired with `await` / disk I/O. Until those are fixed, do not ship global IMMEDIATE alone.

wrolpi/test/test_downloader.py
```diff
@@ -1713,7 +1713,7 @@ async def test_download_dispatched_outside_immediate_transaction(test_session, t
     """The download signal must be dispatched AFTER the claiming write transaction commits.
 
     Regression test for a production deadlock (found on 10.0.0.9, where downloads silently stopped):
-    `_dispatch_new_downloads` claimed downloads inside a `get_immediate_db_session()` (BEGIN
+    `_dispatch_new_downloads` claimed downloads inside a write (BEGIN
     IMMEDIATE) transaction and `await`ed `api_app.dispatch(...)` while that transaction was still
     open.  Under NullPool (production) the dispatched `signal_download_download` opens its OWN
     connection and — inheriting the `_immediate_txn` flag — also issues BEGIN IMMEDIATE, which
```
lrnselfreliance: **[bug]** `test_download_dispatched_outside_immediate_transaction` still probes the flag with `getattr(_immediate_txn, 'active', False)`. `_immediate_txn` is now a `contextvars.ContextVar` with no `.active` attribute, so `getattr` always returns `False` and the assertion always passes. This was the regression test for the production NullPool deadlock (dispatch nested under an open IMMEDIATE). After the ContextVar migration it no longer detects a real regression.

**Suggestion:** Record `_immediate_txn.get()` (and/or assert no write lock is held via the existing probe helper) at dispatch time. Optionally also assert the flag is false *after* the claim session exits in the dispatch task.

wrolpi/db.py
```diff
@@ -184,8 +189,22 @@ def get_db_context() -> Tuple[sqlalchemy.engine.Engine, Session]:
...
+
+    The cost is that writers serialize from BEGIN rather than from their first write, so a write
+    transaction must stay short — do the slow parts (globbing, ffprobe, indexing, big joins) in a
+    read session or outside a session, then write what you computed.  Readers are unaffected: WAL
+    readers never block, and `commit=False` sessions stay deferred.
     """
     _, session = get_db_context()
+    token = _immediate_txn.set(True) if commit else None
     try:
         yield session
```
lrnselfreliance, line 220: **[bug]** `_immediate_txn` is set `True` for the entire `with get_db_session(commit=True)` lifetime. `asyncio.create_task` (used by `background_task` in `create_downloads`) copies the current context, so a task spawned while a write session is open inherits `_immediate_txn=True` permanently for that task (parent `reset(token)` does not clear the child's copy). After the child's own `get_db_session(commit=True)` exits, `reset` restores the *inherited* `True`, so any later `get_db_session(commit=False)` in that task incorrectly issues `BEGIN IMMEDIATE`. `post_download` now wraps `create_downloads` in `commit=True`, so this path is hit on every user-created download. Interleaved Sanic tasks are fixed by ContextVar; task *spawn* inheritance reintroduces a different leak.

**Suggestion:** Prefer clearing the flag as soon as BEGIN has fired (e.g. reset in the `begin` listener after emitting `BEGIN IMMEDIATE`, or set the ContextVar only around the session's first connection use). Alternatively, start background tasks with a clean context (`contextvars.Context().run(...)` / explicit `_immediate_txn.set(False)` at the top of `background_task`). Add a unit test that `create_task` inside `commit=True` does not leave read sessions as IMMEDIATE.

wrolpi/downloader.py
```diff
@@ -1348,7 +1348,10 @@ def renew_recurring_downloads(self):
         """Mark any recurring downloads that are due for download as "new".  Start a download."""
         now_ = now()
 
-        with get_db_session() as session:
+        # `commit=True` (a writer from BEGIN): this reads the recurring Downloads and then renews
+        # the due ones.  As a read session its commit failed with "database is locked" whenever
+        # anything else wrote while it was reading -- a config import or refresh, constantly.
+        with get_db_session(commit=True) as session:
```
lrnselfreliance: **[suggestion]** `renew_recurring_downloads` correctly needs write intent (read then renew), but under IMMEDIATE it now holds the write lock for the full scan of *all* recurring downloads plus `calculate_next_download` (extra queries per row) even when nothing is due. Previously the deferred session only upgraded if something was dirtied/committed. Usually cheap, but on a large recurring list this is avoidable contention on every download-manager cycle.

**Suggestion:** Read due candidates in a deferred session (or raw read), then open a short `commit=True` session only for the rows that need `renew()` / `next_download` updates. Or limit the write transaction to "due" rows only.

wrolpi/files/worker.py
```diff
@@ -2045,7 +2045,7 @@ def on_planning_progress(processed: int, total: int):
             # it must begin as a writer; a deferred transaction's lock upgrade fails instantly when
             # anything else is writing (see `build_move_plan_bulk`).
             with flags.file_worker_discovery:
-                with get_immediate_db_session() as session:
+                with get_db_session(commit=True) as session:
```
lrnselfreliance, line 2048: **[suggestion]** `_execute_move_chunks` (and planning's DB phase) already used IMMEDIATE; this PR only renames the helper. It still holds the write lock across filesystem `mkdir` / file renames inside the transaction. Pre-existing, but more important now that IMMEDIATE is the default write shape and other writers wait up to 30s.

**Suggestion:** Separate "plan + DB path rewrites" from physical moves, or chunk commits so each IMMEDIATE txn only covers a small DB update after disk work for that chunk. Out of scope if moves are rare, but worth a follow-up.

modules/videos/__init__.py
```diff
@@ -76,38 +76,98 @@ async def video_modeler(progress_callback: Callable[[int], None] = None):
...
+
+
+# Videos in a Channel's directory (or below it) that no Channel has claimed yet.
+_UNCLAIMED_VIDEOS_SQL = '''
+                        SELECT v.id, c.id
+                        FROM video v
+                                 INNER JOIN file_group fg ON fg.id = v.file_group_id
+                                 INNER JOIN collection col ON fg.directory = col.directory
+                            OR fg.directory LIKE col.directory || '/%'
+                                 INNER JOIN channel c ON c.collection_id = col.id
```
lrnselfreliance, line 156: **[nit]** `_UNCLAIMED_VIDEOS_SQL` uses `INNER JOIN collection ... ON a OR b` without parentheses. SQL precedence is fine here, but the multi-line formatting makes the OR look like it might bind differently. Semantic match to the old join looks correct: still path-equality or nested `LIKE`, still channel-only via `channel.collection_id`, still one arbitrary channel per video via `_first_claim_per_video`, and `channel_id IS NULL` recheck on UPDATE is the right TOCTOU guard.

**Suggestion:** Parenthesize the ON clause for readability: `ON (fg.directory = col.directory OR fg.directory LIKE col.directory || '/%')`. Optional `ORDER BY` if you want deterministic channel wins for nested channel directories.

wrolpi/test/test_root_api.py
```diff
@@ -969,3 +970,34 @@ async def test_search_other_estimates(async_client, test_session, channel_factor
...
+@pytest.mark.asyncio
+async def test_mutating_request_survives_concurrent_writer(async_client, test_session, test_downloader,
+                                                           test_download_manager):
+    """Creating a Download must not fail because something else is writing.
+
+    Regression test for the production failure: a user created a video download while a refresh was
+    running and got `database is locked` on `INSERT INTO download`.  The request session begins
+    *deferred*, the handler reads (tag/duplicate checks) before it INSERTs, and SQLite refuses that
+    mid-transaction lock upgrade instantly -- busy_timeout is skipped for upgrades.  Mutating
+    requests must begin as writers so busy_timeout absorbs the contention instead.
```
lrnselfreliance, line 984: **[nit]** `test_mutating_request_survives_concurrent_writer` is a good multi-connection regression for POST `/api/download` + `write_lock_held_briefly` + `production_like_sessions`. It does not exercise multi-process or long-held locks (1s holder). Combined with the broken ContextVar deadlock test, production starvation coverage is thinner than the PR narrative.

**Suggestion:** Keep this test; add a modeler regression that asserts no write lock is held during ffprobe (mirror the claim-join test's "statement ran with commit=False" idea), and fix the ContextVar probe in the downloader dispatch test.

lrnselfreliance (commented):
## Summary

This is the correct *systemic* multi-process SQLite fix direction: `commit=True` must mean `BEGIN IMMEDIATE` so lock upgrades never bypass `busy_timeout`, and the long `video_cleanup` join must run as a WAL read with short chunked writes. That matches SQLite docs (BEGIN IMMEDIATE guarantees no later SQLITE_BUSY in the txn), Django 5.1 `transaction_mode: IMMEDIATE`, Rails SQLite concurrency work, and the classic "timeout ignored on upgrade" literature. **Do not merge as-is**: making every `get_db_session(commit=True)` take the write lock at BEGIN without first splitting existing long write sessions (especially `video_modeler` awaiting ffprobe inside the transaction) recreates multi-minute writer starvation. Also fix the broken ContextVar deadlock test and create_task ContextVar inheritance.

## Research notes (why this direction is right)

- SQLite DEFERRED: no lock at BEGIN; SHARED on first read; RESERVED on first write. Mid-txn SHARED→RESERVED when another writer has advanced the DB can return SQLITE_BUSY **without** invoking busy_timeout (stale snapshot / isolation).
- BEGIN IMMEDIATE: may wait (busy_timeout) at start; if it succeeds, no subsequent SQLITE_BUSY until COMMIT.
- Opt-in helpers (`get_immediate_db_session`) get forgotten; Rails/Django defaulted write transactions to IMMEDIATE for that reason.
- Cause 2 is equally real: a 110–130s write lock starves every IMMEDIATE waiter past 30s busy_timeout.

## Issue counts by severity

- bugs: 3
- suggestions: 3
- nits: 2

## Issues outside the diff

- **[suggestion]** wrolpi/api_utils.py:240 — **[suggestion]** Request middleware still injects a deferred session that response middleware commits. The PR correctly documents this and fixes `post_download`, but many mutating endpoints still write through `request.ctx.session` (e.g. `put_download`, channel/collection CRUD, tags). Those remain exposed to the classic mid-transaction SHARED→RESERVED upgrade failure under concurrent refresh/download writers. Not a defect in the claimed change set, but the multi-process problem is only partially closed for the HTTP surface.
  **Suggestion:** Track a follow-up to either (a) make mutating routes use short `get_db_session(commit=True)` sessions, or (b) begin request sessions IMMEDIATE only when the method is unsafe — carefully, since that would hold the lock for the whole handler. Prefer (a) for anything that does I/O.

lrnselfreliance:
All three bugs confirmed and fixed, both suggestions taken, nit applied. Pushed in cf32f34d. Each fix was written test-first, and each test was mutation-checked — reverting the production change alone fails it.

**[bug] `video_modeler` holding the lock across ffprobe** — correct, and the most important finding here. Test first (`test_video_modeler_does_not_hold_write_lock_during_ffprobe`, probes the real write lock from a competing connection while ffprobe runs); it failed exactly as you predicted. The batch is now read, ffprobed with no transaction open, then written. `get_or_create_ffprobe_json` in `modules/videos/common.py` is the new path-only entry point; `Video.get_ffprobe_json` delegates to it and still appends the cache file to its FileGroup.

One correction to the framing: this was not purely introduced by global IMMEDIATE. On master the batch's first `session.flush([video])` already took the write lock before the remaining ffprobes ran, so any batch containing a new Video held it across ffprobe already. Global IMMEDIATE widens it to every batch. The fix is warranted either way.

`validate()` — and the poster generation it can trigger when a Channel asks for it — is still inside the write transaction. It needs a session for the Channel lookup and mutates ORM state, so hoisting it is a bigger change; the code now says so at the call site rather than leaving it implied.

**[bug] vacuous ContextVar probe** — right, `getattr` on a ContextVar always returned the default, so the test passed unconditionally. Fixing the probe to `_immediate_txn.get()` would have left it vacuous in the other direction, since the flag no longer spans the session. It now probes the actual write lock from a competing connection under `production_like_sessions`, which is the invariant the test is named for. Mutation-checked: moving the dispatch back inside the claim transaction fails it.

**[bug] `create_task` context inheritance** — confirmed, with a failing test first (`test_background_task_does_not_inherit_write_intent`): a read session inside a task spawned from a write session took the write lock. Fixed by narrowing the flag to the session's opening `connection()` call instead of clearing it in the `begin` listener — the listener runs inside SQLAlchemy's own machinery, and mutating the ContextVar from there would leave the surrounding `reset()` restoring a value that was never correct. As a side effect the write lock is now taken when the session opens rather than at the caller's first statement, which is what "write intent" should mean anyway.

**[suggestion] `renew_recurring_downloads`** — taken. Reads first, then takes the lock only for rows that change. It also re-checks `status` in the write transaction, since the dispatcher may have claimed a download between the read and the write.

**[suggestion] `_execute_move_chunks` holding the lock across disk I/O** — agreed and out of scope here; noted with the request-session item below.

**[nit] join parenthesization** — done. No `ORDER BY` added: the previous `UPDATE ... FROM` picked arbitrarily too, so imposing an order now would be inventing behavior rather than preserving it.

**[nit] coverage thinness** — fair. Three lock tests now use the shared `probe_write_lock_is_held` (dispatch, modeler, renew scan) plus the claim-join statement test and the API concurrent-writer test. Multi-process and multi-minute holds still aren't covered; that wants a different harness than pytest's single test DB.

One qualification on the research note, since it shapes what we do next: BEGIN IMMEDIATE does **not** guarantee no later `SQLITE_BUSY`. [The SQLite docs](https://www.sqlite.org/lang_transaction.html) say COMMIT can still return `SQLITE_BUSY` when another connection has an open read. It removes the *upgrade* failure, which is the bug we hit, but a retry-on-busy at commit is still worth having eventually — worth its own issue rather than this PR.

Follow-ups I did not do here, per your out-of-diff note: mutating endpoints other than `post_download` still write through the deferred request session, and the move executor still holds the lock across `mkdir`/renames. Both want the slow work moved off the transaction first.

Suite: 1712 passed, 5 skipped.

PR #649, opened 2026-08-08
ui: the navbar shows which section you are in

lrnselfreliance:
Clicking **Videos** opened the videos page and left the bar looking untouched.

## Cause

react-router was doing its part the whole time — the tab carried `active` and `aria-current="page"` on every page under it. **Nothing in the stylesheet drew that class**, so all eleven tabs rendered identically and the mark was visible only to a screen reader.

So the fix is mostly CSS, plus one data correction.

## The mark

A deepened version of the existing hover wash, plus an underline. The underline is a pseudo-element in `currentColor`, not a border:

- `currentColor` is the foreground the bar already computed for whichever of the **fourteen** navbar colors the user picked — one rule covers every color in all four themes without naming one.
- A pseudo-element adds no height. A border would jog the whole row as the mark moved between tabs.

The overflow "More" menu and the mobile hamburger mark their copy of the tab too; on a phone that menu is where most tabs live.

## `end: true` on two links

`Calculators` and `Statistics` carried `end`, which restricts the match to the exact path. A tab is meant to stay current for the pages beneath it, so either would have dropped its mark the moment it grew a sub-page. Removed. The other nine never had it, which is why `/videos/1234` already kept Videos marked.

## Tests

`Nav.test.js` renders the real bar and asserts which tab is current across 17 routes, including sub-paths (`/videos/channel/5/video`, `/files/some/deep/directory`, `/more/calculators/electrical`), that the home link is marked only at `/`, and that exactly one tab is marked at a time.

Two pieces of browser had to be stubbed, and both are worth knowing about:

- **fresnel renders no navbar at all** unless a breakpoint query matches, and renders *two* bars if several match. The mock reports the `computer` breakpoint alone.
- **Every width is 0 in jsdom**, which convinces `useOverflowNav` that nothing fits and sweeps all eleven tabs into the closed "More" menu. Without a width stub the bar renders no tabs and every assertion here passes vacuously.

The styling itself is asserted against the stylesheet, since jsdom applies none — a rendering assertion cannot see the half of this bug the user actually reported.

## Verification

1179 tests across 66 suites pass; build clean (the two `Nav.js` lint warnings are pre-existing, verified against a stashed tree). Checked in the running dev stack: Videos marked on `/videos` and `/videos/1`, Calculators on `/more/calculators`, home only at `/`, and the mark legible on the red bar in both light and night.

### DIFF
--- a/app/src/App.css
+++ b/app/src/App.css
@@ -76,9 +76,10 @@ html .mantine-Notifications-root[data-position^='top'] {
  *
  * The underline is a pseudo-element in `currentColor`, not a border: `currentColor` follows
  * the foreground the bar computed for whatever navbar color the user picked, so this works
- * on all fourteen of them and in all four themes without naming a color; and a pseudo-element
- * adds no height, where a border would jog the whole row by 3px as the mark moved between
- * tabs.  The darker wash is the hover treatment deepened, so the two read as the same idea.
+ * on all twelve of them (see `navColorNames`) and in all four themes without naming a color;
+ * and a pseudo-element adds no height, where a border would jog the whole row as the mark
+ * moved between tabs.  The darker wash is the hover treatment deepened, so the two read as
+ * the same idea.
  */
 .wrolpi-navbar-link.active {
     position: relative;
--- a/app/src/App.css
+++ b/app/src/App.css
@@ -68,6 +68,45 @@ html .mantine-Notifications-root[data-position^='top'] {
 }
 
 /*
+ * The tab for the section the user is in.
+ *
+ * react-router marks it with `active` and `aria-current` on its own, and always did -- the
+ * mark reached the DOM and nothing here drew it, so every tab looked identical and clicking
+ * Videos appeared to leave the bar untouched.
+ *
+ * The underline is a pseudo-element in `currentColor`, not a border: `currentColor` follows
+ * the foreground the bar computed for whatever navbar color the user picked, so this works
+ * on all fourteen of them and in all four themes without naming a color; and a pseudo-element
+ * adds no height, where a border would jog the whole row by 3px as the mark moved between
+ * tabs.  The darker wash is the hover treatment deepened, so the two read as the same idea.
+ */
+.wrolpi-navbar-link.active {
+    position: relative;
+    background: rgba(0, 0, 0, 0.2);
+}
+
+.wrolpi-navbar-link.active::after {
+    content: '';
+    position: absolute;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    /* rem, not px, so the mark grows with the --ui-scale setting like the text above it. */
+    height: 0.1875rem;
+    background: currentColor;
+}
+
+/*
+ * The same tabs inside the overflow "More" menu and the mobile hamburger, which are the
+ * only place most of them appear on a phone.  Those render in a portal, outside the bar,
+ * so they take theme tokens rather than the bar's foreground.
+ */
+.mantine-Menu-item.active {
+    background: var(--head);
+    font-weight: 600;
+}
+
+/*
  * Everything in the bar takes the bar's foreground.
  *
  * `color: inherit` on the links alone was not enough, because two of the things in the
--- a/app/src/components/Nav.test.js
+++ b/app/src/components/Nav.test.js
@@ -1,0 +1,144 @@
+import fs from 'fs';
+import path from 'path';
+import React from 'react';
+import {screen} from '@testing-library/react';
+import {render} from '../test-utils';
+import {FileWorkerStatusContext} from '../contexts/FileWorkerStatusContext';
+import {NavBar} from './Nav';
+
+/*
+ * Which navbar tab is marked as the current page.
+ *
+ * `NavLink` marks itself with `aria-current="page"` and an `active` class, and both need to
+ * be right: the class is what the bar styles, and the attribute is what tells a screen
+ * reader which section the user is in.
+ *
+ * The rule these tests encode is "a tab stays current for everything underneath it" --
+ * /videos/1234 is still Videos.  That is `NavLink`'s default, so the tests exist to catch
+ * an `end` prop added to a link that has pages beneath it, which silently drops the mark on
+ * every one of those pages.
+ */
+
+beforeEach(() => {
+    /*
+     * Two bits of browser the navbar needs and jsdom does not provide.
+     *
+     * Fresnel renders only the breakpoints whose query matches, so a mock matching nothing
+     * renders no navbar at all, and one matching everything renders the mobile and desktop
+     * bars at once.  Match the `computer` breakpoint alone: one bar, tabs in a row.
+     */
+    window.matchMedia = jest.fn().mockImplementation(query => ({
+        matches: query.includes('1024'),
+        media: query,
+        onchange: null,
+        addListener: jest.fn(),
+        removeListener: jest.fn(),
+        addEventListener: jest.fn(),
+        removeEventListener: jest.fn(),
+        dispatchEvent: jest.fn(),
+    }));
+
+    /*
+     * Every width is 0 in jsdom, so useOverflowNav concludes that nothing fits and sweeps
+     * all eleven tabs into the "More" menu -- which is closed, so the bar renders no tabs
+     * and every assertion here would pass vacuously.  A wide bar with modest tabs runs the
+     * same measuring path a browser does.
+     */
+    Object.defineProperty(HTMLElement.prototype, 'clientWidth', {configurable: true, value: 2000});
+    jest.spyOn(HTMLElement.prototype, 'getBoundingClientRect').mockImplementation(() => ({
+        width: 60, height: 40, top: 0, left: 0, right: 60, bottom: 40, x: 0, y: 0,
+        toJSON: () => ({}),
+    }));
+});
+
+// NavBar reads the file-worker status for its indicator icons.  That context lives outside
+// contexts/contexts.js, so the spec supplies it -- see the note at the top of test-utils.
+const fileWorkerStatus = {status: null, error: null, refresh: jest.fn(), setFastPolling: jest.fn()};
+
+const renderNav = (route) => render(<NavBar/>, {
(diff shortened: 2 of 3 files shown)


app/src/components/Nav.test.js
```diff
@@ -0,0 +1,144 @@
...
+
+beforeEach(() => {
+    /*
+     * Two bits of browser the navbar needs and jsdom does not provide.
+     *
+     * Fresnel renders only the breakpoints whose query matches, so a mock matching nothing
+     * renders no navbar at all, and one matching everything renders the mobile and desktop
+     * bars at once.  Match the `computer` breakpoint alone: one bar, tabs in a row.
+     */
+    window.matchMedia = jest.fn().mockImplementation(query => ({
```
lrnselfreliance, line 33: **[suggestion]** The suite only forces the desktop bar (`matchMedia` matches queries containing `1024`, which selects Fresnel’s `computer` breakpoint and therefore `greaterThanOrEqual='tablet'`). It never renders the mobile hamburger (`Media between={['mobile', 'tablet']}`) or opens the overflow “More” menu. On a phone those menus are where almost every tab lives, and the new `.mantine-Menu-item.active` rule is the only mark those paths get — yet neither the DOM path (`DropdownMenuItem` / `Menu.Item component={NavLink}`) nor that CSS rule is exercised.

**Suggestion:** Add a mobile-oriented case (match a sub-`tablet` query, or render `MobileMenu` / open the menu) that asserts the current item is the one with `aria-current="page"` and class `active`, and assert the stylesheet also contains `.mantine-Menu-item.active` (or open the portal and check computed/class presence). Same for an overflow “More” path if you can force `useOverflowNav` to hide the current tab (narrow `clientWidth` / larger item widths).

app/src/components/Nav.test.js
```diff
@@ -0,0 +1,144 @@
...
+         * The reported bug in full: react-router was already marking the tab, and the mark
+         * was already reaching the DOM -- nothing styled it, so every tab looked identical
+         * and clicking Videos appeared to do nothing to the bar.
+         *
+         * jsdom applies no stylesheet, so no rendering assertion can see this.  The rule
+         * itself is the thing to guard.
+         */
+        const css = fs.readFileSync(path.join(__dirname, '..', 'App.css'), 'utf8');
+
+        expect(css).toMatch(/\.wrolpi-navbar-link\.active/);
```
lrnselfreliance: **[suggestion]** The “styles that class” guard only checks that `App.css` contains the substring `.wrolpi-navbar-link.active`. That passes if the rule body is emptied, if the `::after` underline is deleted (the half of the treatment that survives dark washes and is the main non-wash cue), or if `.mantine-Menu-item.active` never lands. Given the original bug was “mark in the DOM, nothing draws it,” this is a thin regression net for the visual half of the fix.

**Suggestion:** Tighten the CSS assertions to require the meaningful declarations (e.g. `.wrolpi-navbar-link.active::after` with `background: currentColor` / non-zero height, and `.mantine-Menu-item.active`), or share a small stylesheet-contract helper like the project already uses for `--ui-scale` / nav color tokens.

app/src/App.css
```diff
@@ -67,6 +67,45 @@ html .mantine-Notifications-root[data-position^='top'] {
...
 
+/*
+ * The tab for the section the user is in.
+ *
+ * react-router marks it with `active` and `aria-current` on its own, and always did -- the
+ * mark reached the DOM and nothing here drew it, so every tab looked identical and clicking
+ * Videos appeared to leave the bar untouched.
+ *
+ * The underline is a pseudo-element in `currentColor`, not a border: `currentColor` follows
+ * the foreground the bar computed for whatever navbar color the user picked, so this works
```
lrnselfreliance, line 78: **[nit]** The new comment says the underline works “on all fourteen of them.” `navColorNames` / `navColorHexMap` define twelve navbar colors, not fourteen. Harmless for runtime, but the comment is the kind of contract note other theme tests treat carefully.

**Suggestion:** Change “fourteen” to “twelve” (or whatever the single source of truth is if the palette grows).

app/src/components/Nav.test.js
```diff
@@ -0,0 +1,144 @@
...
+
+    it.each([
+        ['/videos/1234', 'Videos'],
+        ['/videos/channel/5/video', 'Videos'],
+        ['/archives/domain/example.com', 'Archive'],
+        ['/files/some/deep/directory', 'Files'],
+        ['/inventory/1/items', 'Inventory'],
+        ['/more/calculators/electrical', 'Calculators'],
+    ])('keeps the tab marked on a page underneath it: %s stays %s', (route, text) => {
+        // The reported bug: opening a video left no tab marked, because the tab only
```
lrnselfreliance: **[nit]** The nested-path cases comment that “opening a video left no tab marked, because the tab only matched its own exact path.” Videos in `allLinks` never had `end: true`; the actual reported bug (and the PR description) was missing CSS for an already-active `NavLink`. `end` was only on Calculators/Statistics. The tests are still valuable as a guard against adding `end` later, but the comment misstates the root cause.

**Suggestion:** Split the narrative: CSS for the Videos mark that was always in the DOM; `end` removal as a separate future-proof for Calculators/Statistics subpaths (calculators today use query params, so `end` did not hide those either).

app/src/components/Nav.test.js
```diff
@@ -0,0 +1,144 @@
...
+        ['/zim', 'Zim'],
+        ['/inventory', 'Inventory'],
+        ['/flasher', 'Flasher'],
+        ['/more/calculators', 'Calculators'],
+        ['/more/statistics', 'Statistics'],
+    ])('marks %s as %s', (route, text) => {
+        renderNav(route);
+
+        expect(currentTab()).toContain(text);
+    });
```
lrnselfreliance, line 119: **[nit]** Table cases use `expect(currentTab()).toContain(text)` without requiring uniqueness. A regression that left Home permanently active alongside Videos would still pass every table row; only the separate “exactly one tab” test (one route) would catch it. Home is actually safe without `end` under React Router 7’s root-path rule, but the exclusive check is thin coverage for “exactly one.”

**Suggestion:** Prefer `toEqual([text])` in the table cases, or always assert `currentTab()` length 1 next to `toContain`.

lrnselfreliance (commented):
## Summary

The PR correctly fixes the real bug: `NavLink` was already applying `active` / `aria-current="page"`, but nothing in CSS painted it. The bar treatment (deeper black wash + absolute `currentColor` underline) is a solid, theme-agnostic approach that avoids layout shift and rides the measured bar foreground, so it should track all 12 navbar colors × 4 themes without hardcoding hues. Removing `end` from Calculators/Statistics is the right default for nested sections and does not introduce prefix collisions with the current route tree. Tests thoughtfully stub Fresnel and overflow measurement so they do not pass vacuously, but they only cover the desktop in-bar path and the CSS guard is weaker than the comment implies for menu/mobile and for the underline itself.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 3

## Issues outside the diff

- **[suggestion]** app/src/components/Nav.js:135 — **[suggestion]** When the current section is only inside the overflow “More” dropdown, nothing in the bar itself is marked: `DropdownLinks` renders a plain `NavDropdownTrigger` button, not a `NavLink`, so the new active wash/underline never appear until the user opens the menu. The PR’s goal is “the navbar shows which section you are in,” and on narrower desktop widths that is still untrue for overflowed tabs (common once status icons grow the right cluster). Not a regression from `end`/CSS alone, but it is a hole in the feature as shipped.
  **Suggestion:** Consider marking the “More” trigger when any of its `overflowLinks` is active (class on the button, or `aria-current` on a suitable child), so the bar always shows a current-section cue without requiring the menu to be open.

lrnselfreliance:
Triaged all six — 0 bugs, and every one of them was fair. All addressed in 5ccfd12d.

## The substantive one

**Overflowed tabs (Nav.js:135).** You're right, and it's the worse half of the bug I set out to fix: narrow the window enough to fold the current tab into "More" and the bar goes blank again. The trigger is a plain button, not a `NavLink`, so nothing was marked until the user opened the menu to look — which is exactly when they least need telling.

"More" now takes the mark when it holds the current section. That needs the matching rule *outside* a `NavLink`, where the hooks that would answer it can't be called in a loop, so `isLinkActive` states it once: a link owns the pages beneath it unless it asks for an exact match.

I gave the trigger **the class only, not `aria-current`** — the real `NavLink` inside the menu already carries it, and a second one on the button that opens the menu announces the same page twice. Verified in the browser at 1100px: on `/zim` (folded into More) the trigger is marked; on `/videos` (still in the bar) it isn't.

## The rest

| | |
|---|---|
| **Mobile + overflow untested** | Both now covered. `.mantine-Menu-item.active` — the whole of the mark on a phone — was previously exercised by nothing. This meant splitting viewport and bar width out of the fixture so a test can ask for a narrow bar or a phone. The mobile case also needs the *real* `getBoundingClientRect` back: floating-ui positions the menu's portal and crashes the worker on a stub DOMRect. |
| **Thin CSS guard** | Agreed — it passed with the rule body emptied or the `::after` deleted, and the underline is the half that survives on a dark bar. It now reads the declarations: the wash, `content`, `background: currentColor`, a parsed non-zero height, and the menu-item rule. |
| **`toContain` in table cases** | `toEqual([text])`. As you say, a regression leaving Home permanently marked satisfied "contains" on every row. |
| **Comment misstates root cause** | Corrected. It claimed the tab "only matched its own exact path", which is the cause the PR description explicitly rules out. Now says what those cases actually guard: that nobody adds `end` to a link with pages beneath it. |
| **"fourteen" colors** | Twelve. I'd carried over Semantic's palette count, which includes black and white — neither is offered as a navbar color. |

## Confirming the new guards can fail

Both were checked against a deliberately broken tree rather than assumed: breaking the underline (`currentColor` → `red`) fails the CSS contract, and dropping the trigger's class fails the overflow test.

Worth noting one thing I got wrong in the process: my first tightened height assertion used `(?!0)`, which calls `0.1875rem` zero. It parses the number now.

1182 tests across 66 suites pass; build clean.

PR #650, opened 2026-08-08
ui: the Files filter's clear button rejoins its input

lrnselfreliance:
The clear button sat detached from the filter field and 11px below it, and it dragged the collapse-all button out of line with both.

## One stale rule did all three

```css
.search-clear { margin: 0.625rem; }
```

Left over from the Semantic-era filter, where the clear button was a *sibling* of the input and needed that gap. It is now `ActionInput`'s `action`, which joins the two by dropping the input's right border. So the margin:

- **reopened the seam** the component exists to close (11px gap → looked detached),
- **pushed the button down** by its own top margin (11px below the input), and
- **stretched the wrapper to 62px** around a 40px field, because a flex line is as tall as its tallest item *plus that item's margins*. The collapse-all button was then centred in that 62px — which is why it looked wrong too, despite carrying no bad styling of its own.

Measured before → after, in the running dev stack:

| | before | after |
|---|---|---|
| gap between input and clear | 11px | **0** |
| clear button vertical offset | +11px | **0** |
| wrapper height (field is 40px) | 62px | **40px** |
| button tops in the row | 95, 106 | **95, 95** |

The button now abuts the field and shares its outline, like the filters on Videos and Archives. The class had exactly one user left, so the rule goes with it.

## Test

Asserts the clear button is inside `ActionInput`'s wrapper (a sibling would be the detached arrangement again), and that no rule reintroduces a margin on it. jsdom applies no stylesheet, so the geometry itself can't be measured there — I confirmed the test fails when the rule is put back, rather than assuming it would.

## Verification

1158 tests across 65 suites pass; build clean (the three `FileBrowser.js` lint warnings are pre-existing, checked against a stashed tree).

### DIFF
--- a/app/src/components/FileBrowser.test.js
+++ b/app/src/components/FileBrowser.test.js
@@ -1,3 +1,5 @@
+import fs from 'fs';
+import path from 'path';
 import React from 'react';
 import {act, fireEvent, screen, waitFor} from '@testing-library/react';
 import {FileBrowser, filterBrowseFiles} from './FileBrowser';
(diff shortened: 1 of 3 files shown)


app/src/components/FileBrowser.test.js
```diff
@@ -492,6 +494,28 @@ describe('FileBrowser', () => {
...
+             * look the Videos and Archives filters have.
+             *
+             * It used to carry `search-clear`, a class left over from Semantic whose
+             * `margin: 0.625rem` did three things at once: opened a gap so the button read
+             * as detached, pushed it 11px below the input, and stretched the wrapper to
+             * 62px against a 40px field, which threw the collapse-all button out of line
+             * too.  jsdom applies no stylesheet, so the guard is that no rule reintroduces
+             * a margin on a control inside the wrapper.
+             */
+            renderFileBrowser(<FileBrowser browseFiles={mockBrowseFiles}/>);
```
lrnselfreliance, line 510: **[nit]** The new test renders `<FileBrowser browseFiles={mockBrowseFiles}/>`, but `FileBrowser` takes no props and always reads `browseFiles` from `useBrowseFiles()`. The prop is ignored; the fixture only appears because `beforeEach` already mocks that hook with `mockBrowseFiles`. Sibling Filter tests correctly use `renderFileBrowser(<FileBrowser/>)` (and override the mock when they need different data).

**Suggestion:** Drop the unused prop so the test matches the rest of the file: `renderFileBrowser(<FileBrowser/>);`

lrnselfreliance (commented):
## Summary

The PR correctly removes the sole consumer of the Semantic-era `search-clear` class and its `margin: 0.625rem` rule, so ActionInput can keep the filter clear button joined to the field without a detached gap or stretched flex row. The class is not used elsewhere (epub’s `search-clear` is an unrelated element id), and ActionInput’s existing stretch/flex rules are the right place for join/alignment. The new unit test is a reasonable dual guard (DOM membership + CSS-source absence); one small test nit remains.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 1

PR #654, opened 2026-08-11
ui: a PDF asks before it lights up a night-adapted eye

lrnselfreliance:
Night and amber pass media through an SVG color matrix so nothing a user opens costs them their
dark adaptation. It reaches everything drawn in our own document — images, video, canvases, and
same-origin iframes like the EPUB viewer.

**It does not reach a PDF**, and the miss is not subtle.

## Measured, on a device in night mode

| Surface | Non-red pixels |
| --- | --- |
| App chrome | 0% |
| **PDF on `/docs/:id`** | **55%** |
| **PDF in the preview modal** | **39.8%** |
| EPUB viewer | 0% |
| Image preview | 0% |

Chrome draws PDFs in its own out-of-process viewer, which the compositor paints outside the
parent's filter. The computed style on the iframe really does say `filter:
url(#wrolpi-night-red)` — and a full-brightness photograph of a blue sky arrives anyway, while
the chrome around it is correctly red. The EPUB result is the control that proves it isn't
iframes generally: same rule, same kind of element, our own document, filtered perfectly.

## Tinting from outside was tried, and does not work

A red `mix-blend-mode: multiply` overlay above the iframe cannot see the plugin's pixels, so it
stops blending and paints flat opaque red — **zero non-red pixels by measurement, document
entirely obliterated.** Worth stating plainly: the spectral metric passed on a result that was
useless, so it is not sufficient on its own.

## What this does

Withholds the PDF and lets the reader choose. Nothing renders until they do — hiding it with CSS
would still load and paint the viewer, which is the flash being avoided.

Gated on whether the theme is **currently filtering**, not on its name: a reader who turns
night's filter off has said they don't want this, and amber with its filter on has the same
unreachable PDF and does. Only PDFs are gated — text and the EPUB viewer tint correctly, so
gating them would hide something for no reason.

`MediaGate` takes `gated` as a prop rather than reading ThemeContext, so it stays a library
component like any other; the architectural test that enforces that still passes.

Also corrects the gallery paragraph claiming the matrix remaps PDFs — the one place we told
users something this measurement disproves. Added to `/theme-sample` per the standing rule.

## Checks

- 67 suites / 1195 tests pass (9 new); `npm run build` compiles
- Both new guards verified to fail when the behavior is removed
- Rendered in night mode from a production build

### DIFF
--- a/app/src/components/Docs.js
+++ b/app/src/components/Docs.js
@@ -1,7 +1,7 @@
-import React, {useEffect, useState} from 'react';
+import React, {useContext, useEffect, useState} from 'react';
 import {Link, Route, Routes, useLocation, useNavigate, useParams, useSearchParams} from "react-router";
 import {deleteFileGroups, getDocStatistics, tagFileGroup, untagFileGroup} from "../api";
-import {Media} from "../contexts/contexts";
+import {Media, ThemeContext} from "../contexts/contexts";
 import {
     APIButton,
     BackButton,
@@ -17,7 +17,18 @@ import {
     toLocaleString,
     useTitle
 } from "./Common";
-import {Button, ButtonGroup, Group, Header, Loading, Panel, Statistic, StatisticGroup, Tabs} from "./ui";
+import {
+    Button,
+    ButtonGroup,
+    Group,
+    Header,
+    Loading,
+    MediaGate,
+    Panel,
+    Statistic,
+    StatisticGroup,
+    Tabs
+} from "./ui";
 import {BulkTagModal} from "./BulkTagModal";
 import {TaggedDeleteConfirmModal} from "./TaggedDeleteConfirmModal";
 import {DeepSearchHint, docMimetypeFilterOptions, FilesView, SearchControlBar} from "./Files";
@@ -157,6 +168,8 @@ function DocsPage() {
 function DocPage() {
     const {fileGroupId} = useParams();
     const navigate = useNavigate();
+    // Whether the theme is currently filtering media; a PDF is the one embed it cannot reach.
+    const {mediaFilterEnabled} = useContext(ThemeContext);
     const {docFile, doc, fetchDoc} = useDoc(parseInt(fileGroupId));
     const [searchParams] = useSearchParams();
     const [taggedFileGroups, setTaggedFileGroups] = useState(null);
@@ -344,11 +357,19 @@ function DocPage() {
         {isCbz && activePath && <CbzViewer path={activePath}/>}
 
         {canEmbed && embedUrl && !isCbz && <div style={{marginBottom: '1em'}}>
-            <iframe
-                src={embedUrl}
-                title={docFile.title || docFile.name}
-                style={{width: '100%', height: '80vh', border: '1px solid var(--border)'}}
-            />
+            {/*
+              * Only the PDF is gated.  The EPUB viewer is our own same-origin document, so the
+              * theme's media filter reaches it; Chrome's PDF viewer is composited out of
+              * process and cannot be tinted.  Keyed on the path so selecting a different file
+              * re-gates rather than inheriting the last one's reveal.
+              */}
+            <MediaGate key={activePath} gated={isPdf && mediaFilterEnabled}>
+                <iframe
+                    src={embedUrl}
+                    title={docFile.title || docFile.name}
+                    style={{width: '100%', height: '80vh', border: '1px solid var(--border)'}}
+                />
+            </MediaGate>
         </div>}
 
         <Panel>
--- a/app/src/components/FilePreview.js
+++ b/app/src/components/FilePreview.js
@@ -103,10 +103,18 @@ function getEpubViewerURL(previewFile) {
 function IframePreview({url, gatePdf}) {
     const {mediaFilterEnabled} = React.useContext(ThemeContext);
 
-    // `gatePdf` only for PDFs.  Text and the EPUB viewer are documents of our own, which the
-    // theme's media filter tints; Chrome's PDF viewer is composited out of process and arrives
-    // in full color however the iframe is styled.
-    return <MediaGate gated={!!gatePdf && !!mediaFilterEnabled}>
+    /*
+     * `gatePdf` only for PDFs.  Text and the EPUB viewer are documents of our own, which the
+     * theme's media filter tints; Chrome's PDF viewer is composited out of process and arrives
+     * in full color however the iframe is styled.
+     *
+     * Keyed by `url`, which is load-bearing rather than tidy.  The effect that rebuilds a
+     * preview sets the modal to null and then to the new content in one pass, so React 18
+     * batches both and reconciles instead of unmounting: the MediaGate fiber survives, and a
+     * reader who revealed one PDF would get the next one in full color with no prompt -- the
+     * exact flash this exists to prevent.  The key forces a remount per document.
+     */
+    return <MediaGate key={url} gated={!!gatePdf && !!mediaFilterEnabled}>
         <div className='preview-fit'>
             <iframe title='textModal' src={url}
                     style={{
--- a/app/src/components/FilePreview.js
+++ b/app/src/components/FilePreview.js
@@ -6,13 +6,14 @@ import {getFile, tagFileGroup, untagFileGroup} from "../api";
 import {ArchivePreviewContent} from "./ArchivePreview";
 import {CbzViewer} from "./CbzViewer";
 import {StlViewer} from "react-stl-viewer";
-import {Button, ButtonGroup, Header, Icon, IconButton, Menu, Modal, toast} from "./ui";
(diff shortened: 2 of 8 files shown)


app/src/components/FilePreview.js
```diff
@@ -93,9 +94,19 @@ function getEpubViewerURL(previewFile) {
...
+ * A component rather than part of `getIframePreviewModal` because it reads the theme: the
+ * modal builders are plain functions called from an event handler, where hooks cannot run.
+ */
+function IframePreview({url, gatePdf}) {
+    const {mediaFilterEnabled} = React.useContext(ThemeContext);
+
+    // `gatePdf` only for PDFs.  Text and the EPUB viewer are documents of our own, which the
+    // theme's media filter tints; Chrome's PDF viewer is composited out of process and arrives
+    // in full color however the iframe is styled.
+    return <MediaGate gated={!!gatePdf && !!mediaFilterEnabled}>
```
lrnselfreliance: **[bug]** `MediaGate` is not keyed by the preview URL (Docs correctly uses `key={activePath}`). File preview rebuilds modal content in one effect that does `setPreviewModal(null)` then immediately `setPreviewModal(<Modal>…</Modal>)`. Under React 18 both updates batch, so React reconciles the old tree with the new one and reuses the `MediaGate` fiber. If the reader revealed PDF A, then opens PDF B without closing the modal (e.g. clicking another file in the browser), `revealed` stays `true` and B paints in full color with no prompt — the flash this feature exists to prevent.

**Suggestion:** Mirror Docs: `<MediaGate key={url} gated={!!gatePdf && !!mediaFilterEnabled}>` (or key `IframePreview` on `url`). That remounts on document change and re-gates without changing `MediaGate` itself.

app/src/components/ui/ui.test.js
```diff
@@ -1694,3 +1695,67 @@ describe('modal sizes', () => {
...
+    it('lets the reader put it away again', async () => {
+        renderUI(<MediaGate gated={true}><p>the document</p></MediaGate>);
+        await userEvent.click(screen.getByRole('button', {name: 'Show the full-color PDF'}));
+
+        await userEvent.click(screen.getByRole('button', {name: 'Hide the PDF again'}));
+
+        expect(screen.queryByText('the document')).not.toBeInTheDocument();
+    });
+
+    it('re-gates when filtering is turned back on', () => {
```
lrnselfreliance: **[suggestion]** The test named "re-gates when filtering is turned back on" only does `gated={false}` → `gated={true}` with initial `revealed === false`. That path never touches the effect at `MediaGate.tsx:44-46` (`if (!gated) setRevealed(false)`), and would still pass if the effect were deleted. The comment describes the real risk — reveal, turn filter off, turn filter on — but the steps never click "Show".

**Suggestion:** Drive the full cycle: mount `gated={true}`, click Show, rerender `gated={false}`, rerender `gated={true}`, assert children are absent again. That is the only assertion that proves the reset effect.

app/src/components/ui/MediaGate.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    // Re-gate when filtering comes back on, so revealing one document does not leave every
+    // later one unguarded.
+    React.useEffect(() => {
+        if (!gated) setRevealed(false);
+    }, [gated]);
+
+    if (!gated) return <>{children}</>
+
+    if (revealed) {
```
lrnselfreliance, line 50: **[suggestion]** When revealed, the hide control is rendered after `children`. In FilePreview, children are a `.preview-fit` block sized to `calc(100dvh - 18.75rem)` — calibrated without an extra button row. "Hide the PDF again" can land below the fold / require scrolling inside a fullscreen modal, so the only easy way out is the modal X (which does unload the iframe, but is not discoverable as "put the bright PDF away").

**Suggestion:** Prefer the hide control above the media, or reserve space inside the modal chrome (toolbar) when gated content is revealed. At minimum, verify on a real device that the button is reachable without hunting.

app/src/components/Docs.js
```diff
@@ -344,11 +357,19 @@ function DocPage() {
...
-                title={docFile.title || docFile.name}
-                style={{width: '100%', height: '80vh', border: '1px solid var(--border)'}}
-            />
+            {/*
+              * Only the PDF is gated.  The EPUB viewer is our own same-origin document, so the
+              * theme's media filter reaches it; Chrome's PDF viewer is composited out of
+              * process and cannot be tinted.  Keyed on the path so selecting a different file
+              * re-gates rather than inheriting the last one's reveal.
+              */}
+            <MediaGate key={activePath} gated={isPdf && mediaFilterEnabled}>
```
lrnselfreliance, line 366: **[suggestion]** FilePreview has a source-contract test that the PDF branch passes `gatePdf: true` and text/EPUB do not. Docs' embed — the surface measured at 55% non-red pixels — has no equivalent guard. A future refactor that embeds the iframe without `MediaGate` would only fail in night-mode manual QA.

**Suggestion:** Add a small source or render test that DocPage wires `MediaGate` with `gated={… mediaFilterEnabled}` around the embed (same spirit as FilePreview.test.js).

app/src/components/ThemeSamplePage.js
```diff
@@ -258,6 +259,23 @@ export function ThemeSamplePage() {
...
+                */}
+                <div style={{marginTop: 18}}>
+                    <MediaGate gated={!!mediaFilterEnabled}>
+                        <img src='/icon.svg' alt='Stand-in for a PDF' width={104} height={104}/>
+                    </MediaGate>
+                    <p style={{fontSize: '0.75rem', color: 'var(--muted)', margin: '8px 0 0'}}>
+                        <code>MediaGate</code>. A PDF is the one thing the filter cannot reach:
+                        the browser draws it in its own viewer, outside the page, so it arrives
+                        in full color however the frame is styled — and painting over it just
+                        hides the document. So night and amber withhold it and let you choose.
```
lrnselfreliance: **[nit]** Copy says "night and amber withhold it and let you choose." Amber's filter is off by default, so amber does not withhold unless the user turns filtering on. The following sentence ("unless a filter is on") is accurate; the earlier clause slightly overstates amber.

**Suggestion:** Phrase as "when a filter is on (night by default, amber if you ask), MediaGate withholds…" to match `resolveMediaFilter` defaults.

lrnselfreliance (commented):
## Summary

Solid, well-motivated fix: PDFs are gated on live `mediaFilterEnabled` (not theme name), `MediaGate` stays a dumb library component, and Docs/FilePreview cover the real embed surfaces. The existing `useContext` architecture guard correctly keeps ThemeContext out of MediaGate. One real correctness gap remains: FilePreview does not key `MediaGate` by document URL, so a revealed PDF can hand its reveal state to the next PDF under React 18 batching. Tests are good on withhold/show/hide but the re-gate case does not actually exercise a prior reveal.

## Issue counts by severity

- bugs: 1
- suggestions: 3
- nits: 2

## Issues outside the diff

- **[nit]** app/src/themes/tokens.css:375 — **[nit]** Gallery copy correctly stopped claiming the matrix remaps PDFs. Developer-facing comments still say PDFs are filtered (`tokens.css`, and similarly `names.ts` / `theme.ts` / `MediaFilterDefs.tsx`). Not user-visible, but the PR's measurement makes those comments false and they will mislead the next person who reaches for a CSS-only fix.
  **Suggestion:** Drop "PDFs" from those lists (or note that OOP PDF viewers are not reached and must be gated in UI). Optional follow-up; not a merge blocker.

lrnselfreliance:
All five accepted and pushed in 9c6f3407. I verified the two structural claims before acting on them, and both hold.

**[bug] MediaGate not keyed in FilePreview** — Confirmed at the mechanism level: the rebuild effect calls `setPreviewModal(null)` and then `setModalContent(...)` in the same pass, so React 18 batches them and reconciles rather than unmounting, and the fiber survives across documents. Keyed by `url`, mirroring what Docs already did with `activePath`. Added a guard for the key *and* one for the premise — if the rebuild ever stopped batching, the reasoning behind the key would be stale rather than merely belt-and-braces.

**[suggestion] the re-gate test proved nothing** — Correct, and this is a failure mode I should have caught: the test went `gated={false}` → `gated={true}` with `revealed` already false, which never reaches the reset effect. I verified the original passed with the effect deleted. It now clicks Show first, and I confirmed it fails without the effect and passes with it.

**[suggestion] hide control below the fold** — Confirmed `.preview-fit` is `calc(100dvh - 18.75rem)`, calibrated with no room for a trailing button row. Moved the control above the media and gave it an eye-slash icon. Putting the way out below a full-screen bright PDF was exactly backwards for the person who just lost their dark adaptation.

**[suggestion] no guard on the Docs embed** — Added `Docs.test.js` with the same contract, asserting the gate wraps the iframe, keys per document, and reads `mediaFilterEnabled` rather than the theme name. Verified it fails against a `theme === 'night'` condition and an unkeyed gate.

**[nit] gallery copy overstates amber** — Fixed; now reads "when a filter is on — night by default, amber if you ask".

68 suites / 1200 tests pass, build compiles.

PR #658, opened 2026-08-12
Map previews that never drew, and a Docker map with no world under it

lrnselfreliance:
Two independent bugs behind "the map works sometimes, and sometimes it does not". Both were reproduced on running systems and are fixed test-first.

## 1. Both preview modals were always blank

The per-region eye button and **View All Regions** each render a `div` into a Modal and hand it to MapLibre from an effect gated on `open`, reading the node out of a `useRef`:

```js
if (!open || !mapContainer.current) return;
```

Mantine's Modal mounts its body **one commit after** `opened` flips true, so on the render that opens the modal the ref is still null. The effect returns, nothing it depends on changes again, and it never runs a second time. There is no console error, no rejected promise, and no canvas — just an empty div, which is why this read as intermittent rather than broken.

An instrumented build prints it directly:

```
DBG allregions effect open=false ref=false catalog=0
DBG allregions effect open=false ref=false catalog=34
DBG allregions effect open=true  ref=false catalog=34   <-- bails, never runs again
```

It worked before the Mantine migration because Semantic UI's Portal put the modal body in the DOM in the same commit that opened it. Sampling a device still on the older build shows exactly that — modal present at t=0ms, canvas by t=50ms — while the effect bodies are byte-identical across the change. The modal library is the only variable.

Fixed with a callback ref, so the effect re-runs when the node arrives and this no longer depends on any modal's mounting schedule. The async source lookup also gets a `cancelled` flag: it could resolve after cleanup ran, building a map nothing would ever remove and leaking a WebGL context on every quick open-and-close.

## 2. Docker had no world underneath its map regions

Caddy serves `/blobs/map-overview.pmtiles` out of `/opt/wrolpi-blobs`. That blob is the world at zoom 0-6 — what a regional extract sits on top of. Without it, a user who subscribes to a few regions gets those regions floating on an empty background everywhere their data does not reach. Only a full planet file (134GB) makes it unnecessary.

Raspberry Pi and Debian installs get the blob from pi-gen, the live-build config, and `upgrade.sh`. Docker got nothing: `docker-compose.yml` bind-mounted a git-ignored host directory over `/opt/wrolpi-blobs`, and a bind mount *replaces* the image's directory rather than merging with it — so the path existed and was empty on every Docker WROLPi. Confirmed inside a running container, and the URL returned 404 there while returning 206 on hardware.

The web image now fetches the blob at build time as the other three builders do, and the mount that would hide it is gone. `curl -f` so an HTTP error fails the layer — an image that builds without the blob is the same bug, except silent at build time too.

## Testing

Test-first: each new test was written against the unfixed code and observed to fail.

- `app/src/components/MapPreview.test.js` — 8 tests driving the real page. **7 fail before the fix**, all 8 pass after. They assert the outcome (a map was built against a node that is really in the document) rather than the mechanism that quietly changed underneath it.
- `modules/map/test/test_map_overview_blob.py` — 7 tests over the deployment files. **4 fail before the fix**, all 7 pass after. Three are premise checks, so a mis-parse cannot read as a pass.

Verified end-to-end on the Docker stack: after rebuilding the web image, `/blobs/map-overview.pmtiles` went 404 to 206, the baked blob is 44,412,070 bytes with the `PMTiles` v3 magic, and the preview renders the full basemap with all region boxes.

- Backend: 1728 passed, 5 skipped
- Frontend Jest: 1210 passed, 69 suites
- `npm run build`: clean

**Cypress was not run.** Its e2e suite is broken on master, before and independent of this branch: the Cypress 15 esbuild preprocessor refuses JSX in a `.js` file and dies on `src/hooks/customHooks.js:283`, which has been unchanged since 2024. Identical failure on a clean master checkout. Worth its own fix; deliberately not bundled here.

### DIFF
--- a/app/src/components/Map.js
+++ b/app/src/components/Map.js
@@ -92,6 +92,9 @@ function RegionPreviewModal({bbox, name, open, onClose}) {
             });
 
             map.on('load', () => {
+                // The flag again: cleanup can land between construction and `load`, and this
+                // body draws into a map `remove()` has already torn down.
+                if (cancelled) return;
                 map.addSource('bbox', {
                     type: 'geojson',
                     data: {
@@ -122,7 +125,7 @@ function RegionPreviewModal({bbox, name, open, onClose}) {
                 });
                 map.fitBounds([[minLon, minLat], [maxLon, maxLat]], {padding: 40});
             });
-        });
+        }).catch(e => console.error('Could not build the region preview map:', e));
 
         return () => {
             cancelled = true;
@@ -171,6 +174,8 @@ function AllRegionsPreviewModal({catalog, open, onClose}) {
             });
 
             map.on('load', () => {
+                // See RegionPreviewModal: cleanup can land between construction and `load`.
+                if (cancelled) return;
                 const regions = catalog.filter(r => r.bbox);
                 regions.forEach((region, i) => {
                     const [minLon, minLat, maxLon, maxLat] = region.bbox.split(',').map(Number);
@@ -215,7 +220,7 @@ function AllRegionsPreviewModal({catalog, open, onClose}) {
                     });
                 });
             });
-        });
+        }).catch(e => console.error('Could not build the all-regions preview map:', e));
 
         return () => {
             cancelled = true;
--- a/app/src/components/Map.js
+++ b/app/src/components/Map.js
@@ -51,21 +51,34 @@ async function getPreviewMapSource() {
     return {type: 'vector', url: `pmtiles://${MAP_OVERVIEW_URL}`};
 }
 
+/*
+ * The container is a piece of state set by a callback ref, not a `useRef` read in the effect.
+ *
+ * A modal body is not in the DOM on the render that opens the modal -- Mantine mounts it a
+ * commit later -- so an effect gated on `open` that reads `ref.current` finds null, returns,
+ * and never runs again, because nothing it depends on changes after that.  Both previews were
+ * silently blank for exactly this reason.  A callback ref re-runs the effect when the node
+ * arrives, whenever that is, so this does not depend on any modal's mounting schedule.
+ */
 function RegionPreviewModal({bbox, name, open, onClose}) {
-    const mapContainer = React.useRef(null);
+    const [mapContainer, setMapContainer] = useState(null);
 
     React.useEffect(() => {
-        if (!open || !mapContainer.current || !bbox) return;
+        if (!open || !mapContainer || !bbox) return;
 
         const [minLon, minLat, maxLon, maxLat] = bbox.split(',').map(Number);
         const centerLon = (minLon + maxLon) / 2;
         const centerLat = (minLat + maxLat) / 2;
 
         let map;
+        // The source lookup is async, so the modal can close before it resolves.  Without this
+        // the cleanup below runs while `map` is still undefined and the map built afterwards
+        // is never removed -- a leaked WebGL context on every quick open-and-close.
+        let cancelled = false;
         getPreviewMapSource().then(source => {
-            if (!mapContainer.current) return;
+            if (cancelled) return;
             map = new maplibregl.Map({
-                container: mapContainer.current,
+                container: mapContainer,
                 style: {
                     version: 8,
                     sources: {'basemap': source},
@@ -111,14 +124,17 @@ function RegionPreviewModal({bbox, name, open, onClose}) {
             });
         });
 
-        return () => { if (map) map.remove(); };
-    }, [open, bbox]);
+        return () => {
+            cancelled = true;
+            if (map) map.remove();
+        };
+    }, [open, bbox, mapContainer]);
 
     return <Modal open={open} onClose={onClose} size='fullscreen' closeIcon title={name}>
         {/* No `.media` here: the bbox highlight is drawn as a map layer on the canvas itself
             (see the `bbox` source/layers above), so the automatic `canvas` element-type filter
             already covers it. There is no separate DOM overlay to wrap. */}
-        <div ref={mapContainer} style={{width: '100%', height: '50vh'}}/>
+        <div ref={setMapContainer} style={{width: '100%', height: '50vh'}}/>
     </Modal>;
 }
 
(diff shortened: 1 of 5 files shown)


app/src/components/Map.js
```diff
@@ -51,21 +51,34 @@ async function getPreviewMapSource() {
...
+        let cancelled = false;
         getPreviewMapSource().then(source => {
-            if (!mapContainer.current) return;
+            if (cancelled) return;
             map = new maplibregl.Map({
-                container: mapContainer.current,
+                container: mapContainer,
                 style: {
                     version: 8,
                     sources: {'basemap': source},
```
lrnselfreliance, line 84: **[suggestion]** The new `cancelled` flag is checked only before `new maplibregl.Map(...)`. The `map.on('load', ...)` callbacks in both `RegionPreviewModal` and `AllRegionsPreviewModal` still run `addSource` / `addLayer` / `fitBounds` even if cleanup already called `map.remove()`. Quick open-and-close (the exact race the cancelled flag was added for) can leave `load` firing against a removed map and throw in the event handler. Same gap at the All Regions `load` handler.

**Suggestion:** Guard the load body with `if (cancelled) return;` (or rely on a local `let map` check after `remove` by clearing `map = null` in cleanup and testing that). Mirror the `destroyed` pattern already used in `MapViewer.js`.

docker/web/Dockerfile
```diff
@@ -1,7 +1,26 @@
...
+#
+# Raspberry Pi and Debian installs get this from pi-gen, the live-build config, and
+# upgrade.sh.  Docker had no equivalent and shipped without it -- see
+# modules/map/test/test_map_overview_blob.py.  42 MB.
+#
+# `curl -f` so an HTTP error fails this layer rather than writing the error body to the file
+# and exiting 0; an image that builds without the blob is the bug all over again, except now
+# it is silent at build time too.
+RUN mkdir -p /opt/wrolpi-blobs && \
+    curl -fsSL https://wrolpi.nyc3.cdn.digitaloceanspaces.com/maps/map-overview.pmtiles \
```
lrnselfreliance, line 21: **[suggestion]** The overview blob is fetched by URL with `curl -fsSL` and only validated as non-empty (`[ -s ...]`). There is no size floor, checksum, or content-type check. A CDN outage that returns a small error HTML body would fail `-f` (good); a wrong or truncated-but-non-empty object would bake a broken basemap into the image and only fail at runtime. Project pi-gen/live-build use the same URL-without-hash pattern, so this is consistent, but Docker is now a third download site for the same artifact.

**Suggestion:** Optionally pin a known size range (e.g. roughly 40–50 MB) or a SHA-256 beside the curl, shared with `upgrade.sh` / pi-gen if you want one source of truth. Not required for merge if matching existing blob-fetch style is intentional.

docker-compose.yml
```diff
@@ -121,7 +121,10 @@ services:
       - './icon.png:/etc/caddy/icon.png'
       - './test:/opt/media'
       - './modules/map/static:/opt/wrolpi/modules/map/static:ro'
-      - './blobs:/opt/wrolpi-blobs:ro'
+      # No mount over /opt/wrolpi-blobs.  A bind mount REPLACES the image's directory rather
```
lrnselfreliance, line 124: **[suggestion]** Removing `./blobs:/opt/wrolpi-blobs:ro` is correct and required so the image-baked `map-overview.pmtiles` is visible. Local-dev implication: there is no longer a compose-level way to drop extra/replacement blobs into the web container without rebuilding the image or adding an override file. That is almost certainly fine for this single overview file (and zim has its own blob copy), but it is a behavior change for anyone who had started using `./blobs` as an override hook.

**Suggestion:** Keep the comment as-is (it already explains why). If override remains useful, document `docker-compose.override.yml` with a named volume or a non-shadowing path rather than reintroducing a bind mount over `/opt/wrolpi-blobs`.

app/src/components/Map.js
```diff
@@ -51,21 +51,34 @@ async function getPreviewMapSource() {
...
         const [minLon, minLat, maxLon, maxLat] = bbox.split(',').map(Number);
         const centerLon = (minLon + maxLon) / 2;
         const centerLat = (minLat + maxLat) / 2;
 
         let map;
+        // The source lookup is async, so the modal can close before it resolves.  Without this
+        // the cleanup below runs while `map` is still undefined and the map built afterwards
+        // is never removed -- a leaked WebGL context on every quick open-and-close.
+        let cancelled = false;
         getPreviewMapSource().then(source => {
```
lrnselfreliance, line 78: **[nit]** `getPreviewMapSource().then(...)` has no `.catch()`. Today `getPreviewMapSource` always resolves (internal try/catch), so this is safe. If the helper ever starts rejecting, or if `new maplibregl.Map` throws synchronously inside the `then`, that becomes an unhandled rejection. Pre-existing shape, still true after the fix.

**Suggestion:** `.then(...).catch(() => {})` or try/catch inside the `then` after the cancelled check; low priority.

lrnselfreliance (commented):
## Summary

Solid, well-documented PR that correctly fixes both independent bugs: blank Mantine map preview modals (Mantine `Portal` returns null on first mount, so a `useRef` + `open`-gated effect never re-runs) and Docker’s missing overview blob (empty gitignored `./blobs` bind mount shadowed `/opt/wrolpi-blobs`). The callback-ref approach, cancelled flag on the async source path, Dockerfile `curl -f` + non-empty check, compose mount removal, and outcome-oriented tests are the right shape. Remaining items are incomplete cancellation in `load` handlers, no blob integrity pin, and a few test nits—not blockers.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 3

## Issues outside the diff

- **[nit]** app/src/components/MapPreview.test.js:295 — **[nit]** `draws a layer for every region that has a bbox` installs `jest.spyOn(map, 'addSource')` after construction, then manually re-invokes `map.handlers.load()`. The FakeMap already fires `load` via `setTimeout(..., 0)` when `on('load')` is registered, so the auto-fire may have already run before the spy. The manual second fire makes the assertion pass, but it does not prove the production path only draws once and is slightly mock-theater.
  **Suggestion:** Spy on `FakeMap.prototype.addSource` before the click, or record `addSource` calls inside FakeMap (like `constructed`), and assert against that without a second manual `load()`.

- **[nit]** app/src/components/MapPreview.test.js:338 — **[nit]** “closing a preview / removes the map” waits until the map exists, then Escape. It does not cover the race the PR comment highlights: close before `getPreviewMapSource()` resolves, where `cancelled` must prevent a leaked Map. That path is the reason `cancelled` was added and is currently untested.
  **Suggestion:** Mock `getMapFiles` with a deferred promise, open the modal, Escape before resolve, then resolve and assert `constructed` stays empty (or any late map has `removed === true`).

PR #659, opened 2026-08-12
ui: an icon button that was given a color, and never used it

lrnselfreliance:
`IconButton` resolved its variant as:

```ts
variant={props.variant ?? fromRole?.variant ?? 'default'}
```

Any call site naming a **color** but no role and no variant therefore landed on Mantine's `default` variant — which draws its background, hover, text and border from `--mantine-color-default*` alone and **never reads `color`**. The button rendered byte-identical to one with no props at all:

```
<IconButton color='violet'/>   variant=default   --ai-bg: var(--mantine-color-default)
<IconButton/>                  variant=default   --ai-bg: var(--mantine-color-default)
<IconButton color='violet' variant='filled'/>    --ai-bg: var(--mantine-color-violet-filled)
```

`Button` was never affected — it leaves `variant` undefined and Mantine fills in `filled`, which is why the same `color='violet'` has always worked there. This makes `IconButton` match: `filled` when a color is named, `default` when one is not (most of them).

### The eleven call sites

Each was passing a color that could not reach a pixel:

| | |
|---|---|
| `DonatePage:19`, `Files:421` | the color **the user picked**, discarded |
| `Downloads:241,404,427,460` | `danger` and `blue` on the download controls |
| `ControllerPage:326,408,1591` | the service **Open** buttons |
| `FilePreview:71,541` | |

Both `violet` and `danger` are valid keys in our palette (`mantine.ts:28,39`) — the names were fine, the variant discarded them.

### Precedence

The color fallback is **last**, behind both `variant` and `role`, so `role='danger'` keeps its dashed night outline even when a color is also named. Tested.

### Verification

Checked in the browser across all four themes. Each icon button now matches the labelled button of the same color exactly; an icon button with no color keeps the neutral default:

```
light   violet  rgb(92,79,168)    both     no-color  rgb(255,255,255) + border
dark    violet  rgb(154,143,224)  both     no-color  rgb(38,41,43)    + border
night   violet  rgb(122,41,41)    both     no-color  rgb(22,4,4)      + border
amber   violet  rgb(122,76,20)    both     no-color  rgb(26,15,0)     + border
```

In night and amber the hues collapse to brightnesses, which is what those themes intend — the same thing already happens to labelled buttons.

The gallery grows one labelled/icon pair per hue, so a regression here is visible rather than silent.

- 68 suites / 1204 tests pass; the new test fails on the old fallback
- `npm run build` is clean

### DIFF
--- a/app/src/components/ui/ui.test.js
+++ b/app/src/components/ui/ui.test.js
@@ -309,6 +309,48 @@ describe('Button', () => {
 
         expect(screen.getByRole('button', {name: 'Delete channel'})).toBeInTheDocument();
     });
+
+    it('paints an icon button the color it was given', () => {
+        /*
+         * `IconButton` fell back to Mantine's `default` variant whenever no role was set, and
+         * that variant draws from `--mantine-color-default*` alone -- it never reads `color`.
+         * So a named color rendered byte-identical to no color at all, silently, at eleven
+         * call sites.  `Button` was never affected: it leaves `variant` undefined and Mantine
+         * fills in `filled`, which is why the same `color='violet'` worked there.
+         */
+        renderUI(<>
+            <IconButton icon='external' label='Open it' color='violet'/>
+            <IconButton icon='external' label='Plain'/>
+            <Button color='violet'>Labelled</Button>
+        </>);
+
+        const bg = (name, role = 'button') =>
+            screen.getByRole(role, {name}).getAttribute('style') || '';
+
+        // The color reaches the background, and matches what the labelled Button already did.
+        expect(bg('Open it')).toContain('var(--mantine-color-violet-filled)');
+        expect(bg('Labelled')).toContain('var(--mantine-color-violet-filled)');
+        // An icon button that names no color keeps the neutral default -- most of them.
+        expect(bg('Plain')).toContain('var(--mantine-color-default)');
+        expect(bg('Plain')).not.toContain('violet');
+    });
+
+    it('lets a role and an explicit variant still win over a color', () => {
+        // The color fallback is the LAST resort: it must not outrank either of the two ways a
+        // call site already had to state a variant, or `role='danger'` would lose its dashed
+        // night outline the moment someone also named a color.
+        renderUI(<>
+            <IconButton icon='trash' label='Roled' role='danger'/>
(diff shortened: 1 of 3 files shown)


app/src/components/ui/ui.test.js
```diff
@@ -309,6 +309,48 @@ describe('Button', () => {
...
+        // The color reaches the background, and matches what the labelled Button already did.
+        expect(bg('Open it')).toContain('var(--mantine-color-violet-filled)');
+        expect(bg('Labelled')).toContain('var(--mantine-color-violet-filled)');
+        // An icon button that names no color keeps the neutral default -- most of them.
+        expect(bg('Plain')).toContain('var(--mantine-color-default)');
+        expect(bg('Plain')).not.toContain('violet');
+    });
+
+    it('lets a role and an explicit variant still win over a color', () => {
+        // The color fallback is the LAST resort: it must not outrank either of the two ways a
```
lrnselfreliance, line 339: **[nit]** The second test's comment says the color fallback must not outrank a role "or `role='danger'` would lose its dashed night outline the moment someone also named a color," but the case actually rendered is `role='danger'` alone plus a separate `color`+`variant='outline'` control. The role→variant win is exercised; the combined `role='danger' color='…'` path described in the comment is not.

**Suggestion:** Optionally add `<IconButton role='danger' color='violet' label='…'/>` and assert `data-variant='filled'` plus `wrolpi-button-danger`, so the comment's precedence claim is locked by the same case it names. Not blocking — role already supplies `variant: 'filled'` before the color branch runs.

lrnselfreliance (commented):
## Summary

Solid, tightly scoped fix for a real silent failure: `IconButton` always fell back to Mantine's `default` variant, which ignores `color`, so named colors never reached pixels. The new chain (`props.variant` → `fromRole?.variant` → color-implies-`filled` → `default`) matches `Button`'s effective filled default when a color is named, keeps neutral chrome for the common no-color case, and leaves role/`variant` call sites alone. Gallery and unit tests assert real `--mantine-color-*-filled` / `default` CSS vars (not mocks) and cover the main inverse and precedence paths; no correctness bugs found.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 1

PR #660, opened 2026-08-12
ui: a downloads table wider than the phone showing it

lrnselfreliance:
The two downloads tables came out **676px and 727px in a 540px window**. Completed At and Control sat off the right edge, and their buttons could not be reached without scrolling sideways.

### Cause

`.table-ellipsis` was trying to truncate under `table-layout: auto`, where a table is as wide as its columns want to be and the container simply overflows:

```css
.table-ellipsis td.column-ellipsis {
    max-width: 0;        /* the ellipsis trick for auto layout */
    min-width: 25em;     /* "About iPhone SE width" */
}
```

`25em` is 385px. A floor wider than the screen makes the overflow certain.

Fixed layout inverts it — the columns divide the table's width, and the column marked `column-ellipsis` gives up what it cannot show. Neither the max-width trick nor the floor is needed then. Columns that must fit something exact (a checkbox, a group of buttons) state a width; the remainder is the truncating column's.

**`table.table-ellipsis`, not the bare class.** Mantine sets `table-layout` on its own `.mantine-Table-table` — same specificity, later in the cascade — so the first version of this lost silently. Every other property in the rule applied, so the table *looked* fixed while it stayed `auto`, and the recurring table came out **wider** than before (871px). Worth knowing for the next Mantine override.

### Three things the rework had to answer for

Each measured in the browser, not reasoned about:

**A header overran its column.** "Completed At" printed on top of "Control", because `.wrolpi-th-sort` is `nowrap` so a label never splits from its arrow. It wraps here — and needs `min-width: 0` as well: the label is an anonymous flex item whose automatic minimum is its longest word, so without it the label still refuses to go below "Completed" and still overruns.

**A grouped button was squeezed.** `ButtonGroup` is a nowrap flex row whose children shrink, and the Stop button on a pending download came out **23px** against its neighbour's 40 — clipped, and too small to hit. Columns are now sized for the widest group a row can hold (a failed download shows delete, edit and retry), so the buttons no longer give.

**Status's Top Processes table had the same bug.** It already asked for `.table-ellipsis` and already marked its Command cell, so it overflowed a phone too (548px into 503px). The same rule fixes it, plus widths for the three short readings.

### Headers stay on one line

A wrapped header is the fallback, not the layout, so every column is sized to hold its own label unwrapped — including on a phone. "Download Frequency" is the exception at 11.7em, which a phone cannot spare beside four other columns, so it takes a class that widens at the tablet breakpoint and wraps below it.

### Measured

Nothing overflowing, no header overrun, no clipped cell, no shrunken button:

```
540px    once       40 / 208 / 146 / 146
         recurring  123 / 108 / 131 / 62 / 116
1400px   once       40 / 1033 / 146 / 146
         recurring  871 / 185 / 131 / 62 / 116
```

Desktop gains too: the checkbox column drops from 90px to 40 and Completed At from 291 to 146, so the URL column picks up about 255px.

### Testing

Five new tests in the Cypress layout suite. None of this is checkable in jsdom — it is all resolved widths, and the central question is whether our `table-layout` beats the one Mantine sets, which is exactly the part that failed silently.

- `npm run build` clean

### DIFF
--- a/app/src/App.css
+++ b/app/src/App.css
@@ -336,12 +336,94 @@ textarea.otp {
     cursor: pointer;
 }
 
+/*
+ * A table that truncates to the width it has, rather than growing past it.
+ *
+ * `table-layout: fixed` is what makes the truncation possible at all.  Under the default
+ * `auto` layout a table is as wide as its columns want to be and the container simply
+ * overflows -- on a 540px phone the downloads tables came out 676px and 727px, so Completed
+ * At and Control sat off the right edge and the buttons could not be reached without
+ * scrolling sideways.  Fixed layout inverts that: the columns get the table's width divided
+ * between them, and the one marked `column-ellipsis` gives up what it cannot show.
+ *
+ * The old rule tried to do this with `max-width: 0` and `min-width: 25em`.  The max-width was
+ * the usual trick for forcing an ellipsis under `auto` layout; the min-width was there so the
+ * text column stayed readable -- but 25em is 385px, and a floor wider than the screen is what
+ * made the overflow certain.  Under fixed layout neither is needed: the column width comes
+ * from the header, and every other column is sized to its content so the ellipsis column
+ * keeps the remainder.
+ *
+ * Columns that must fit something exact -- a checkbox, a group of buttons -- state a width on
+ * their header cell.  Any column that does not is sized by what is left over.
+ */
+/*
+ * `table.` rather than the class alone: Mantine sets `table-layout` on its own
+ * `.mantine-Table-table`, which is the same specificity and comes later in the cascade, so
+ * the bare class silently lost and the table stayed `auto` -- with every other property here
+ * applying, which made it look like the rule had worked.
+ */
+table.table-ellipsis {
+    width: 100%;
+    table-layout: fixed;
+}
+
 .table-ellipsis td.column-ellipsis {
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
-    max-width: 0; /* Ensures the content doesn't exceed the cell's width */
-    min-width: 25em; /* About iPhone SE width */
+}
+
+/*
+ * The one column width that has to change with the screen.
+ *
+ * Most headers here are sized to hold their own label on one line at any width -- a wrapped
+ * label is a fallback, not a layout.  "Download Frequency" is the exception: 11.7em on one
+ * line, which a phone cannot spare when four other columns need room, and it is one of five
+ * columns rather than one of four.  So it wraps below the tablet breakpoint and gets its line
+ * back above it, where the width costs nothing.
+ *
+ * 700px is the app's `tablet` breakpoint (see `AppMedia` in contexts.js).
+ */
+.table-ellipsis th.col-frequency {
+    width: 7em;
+}
+
+@media (min-width: 700px) {
+    .table-ellipsis th.col-frequency {
+        width: 12em;
+    }
+}
+
+/*
+ * A header wraps instead of running into the next column.
+ *
+ * `.wrolpi-th-sort` is `white-space: nowrap` so a sortable label never splits from its arrow.
+ * That is right under `auto` layout, where the column widens to fit; under fixed layout the
+ * column does not widen, so "Completed At" overran its cell and printed on top of "Control".
+ * Wrapping to two lines is the only thing a fixed column can do with a label too long for it.
+ *
+ * `min-width: 0` as well as the wrap: the button is a flex container, and its text is an
+ * anonymous flex item whose automatic minimum is its longest word.  Without this the label
+ * still refuses to go below "Completed" and still overruns.  A column narrower than its
+ * longest word is a width to fix at the call site, not something CSS can hide.
+ */
+.table-ellipsis th .wrolpi-th-sort {
+    display: flex;
+    white-space: normal;
+    min-width: 0;
+    text-align: left;
+}
+
+/*
+ * Grouped buttons keep their size in a fixed column.
+ *
+ * `ButtonGroup` is a nowrap flex row whose children may shrink, and in a column too narrow
+ * for them they did: the Stop button on a pending download was squeezed to 23px against its
+ * neighbour's 40px, a target too small to hit and visibly clipped.  The column is sized for
+ * the widest group it has to hold (three buttons), so the buttons themselves should not give.
+ */
+.table-ellipsis td .mantine-ButtonGroup-group > * {
+    flex-shrink: 0;
 }
 
 .card-title-ellipsis {
--- a/app/src/components/admin/Downloads.js
+++ b/app/src/components/admin/Downloads.js
@@ -584,6 +584,16 @@ export function OnceDownloadsTable({downloads, fetchDownloads}) {
         }
     };
 
+    /*
+     * Widths for everything except the URL, which takes what is left and truncates.  See
+     * `.table-ellipsis` in App.css: under fixed layout a column with no width stated gets an
+     * equal share, which would give a checkbox as much room as a URL.
+     *
+     * The numbers are what the content needs, measured: a checkbox is 2.5em; "Completed At" is
+     * 9.5em on one line with its sort arrow, which is narrow enough to keep on a phone rather
+     * than wrapping it there; and Control is 9.5em, which holds the widest button group a row
+     * can have (a failed download shows delete, edit and retry).
+     */
     const tableHeaders = [
         {
             key: 'select',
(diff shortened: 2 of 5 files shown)


app/src/components/admin/Downloads.js
```diff
@@ -654,12 +666,22 @@ export function RecurringDownloadsTable({downloads, fetchDownloads, onDelete}) {
...
+                    any width. */}
                 <Table.Row>
-                    <Table.HeaderCell style={{width: '50%'}}>URL</Table.HeaderCell>
-                    <Table.HeaderCell style={{width: '12.5%'}}>Download Frequency</Table.HeaderCell>
-                    <Table.HeaderCell style={{width: '12.5%'}}>Completed At</Table.HeaderCell>
-                    <Table.HeaderCell style={{width: '6.25%'}}>Next</Table.HeaderCell>
-                    <Table.HeaderCell style={{width: '18.75%', textAlign: 'right'}}>Control</Table.HeaderCell>
+                    <Table.HeaderCell>URL</Table.HeaderCell>
+                    <Table.HeaderCell className='col-frequency'>Download Frequency</Table.HeaderCell>
+                    <Table.HeaderCell style={{width: '8.5em'}}>Completed At</Table.HeaderCell>
```
lrnselfreliance, line 682: **[bug]** Recurring's non-URL columns are `7em` (`.col-frequency`) + `8.5em` + `4em` + `7.5em` = `27em`. These tables use Mantine `Table` (`font-size: var(--mantine-font-size-sm)` = `0.875rem`). With the app's `--ui-scale: 1.1` root, that is ~15.4px per em, so the four columns already want ~416px plus the table border. Downloads sits in `PageContainer` with `padding: 0` below 700px, so the table is the viewport. On a 375px iPhone SE / 390–430px phone the specified widths meet or exceed the table; under `table-layout: fixed` the unspecified URL column then gets `max(0, table − sum(specified))` — i.e. nothing — and/or the table overflows again. The original report was a 540px window (where ~124px remains for the URL). The new 400px Cypress fixture never mounts this five-column table, so the regression is untested.

**Suggestion:** Shrink Recurring's phone widths so URL still has a remainder at 375px (e.g. drop Frequency / Completed At / Control a few em, or media-query them like `.col-frequency` already does). A Cypress fixture that is the real five-column Recurring header inside `<Table className='table-ellipsis'>` at 375–400px should assert both `tableLayout === 'fixed'` and a non-zero URL column.

app/src/components/admin/Downloads.js
```diff
@@ -592,11 +602,13 @@ export function OnceDownloadsTable({downloads, fetchDownloads}) {
...
             />,
-            sortBy: null
+            sortBy: null,
+            width: '2.5em',
         },
         {key: 'url', text: 'URL', sortBy: i => i.url.toLowerCase()},
-        {key: 'completed_at', text: 'Completed At', sortBy: i => i.last_successful_download || ''},
-        {key: 'control', text: 'Control', sortBy: null},
+        {key: 'completed_at', text: 'Completed At', sortBy: i => i.last_successful_download || '',
+            width: '9.5em'},
```
lrnselfreliance, line 610: **[bug]** `completed_at` is locked to `9.5em` (~124px of content after Mantine `horizontalSpacing={10}` → `rem(10)` padding). That fits the "Completed At" label + sort arrow, but `OnceDownloadRow` also renders `DownloadProgressModal` in this cell — a `Button` labelled `` `${humanBandwidth(speed)} / ${eta}` `` (backend examples: `'3m44s'`, `'00:19'`). A label like "5 MBps / 3m44s" plus button padding is well over 124px. Recurring uses the same modal in an even tighter `8.5em` column. Under the old `auto` layout the column grew; under `fixed` the button overflows the cell and is clipped by the Control cell's background — exactly when the Stop control is showing. The Cypress "Completed At" column is empty text `"3 days"`.

**Suggestion:** Give an in-progress row a compact control (icon-only, or a short percent) that fits the column, or stop assigning a header-sized width to a cell that can hold a labelled button. Add a fixture with `DownloadProgressModal`'s real label in the Completed At cell and assert it stays inside the cell box.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -1448,6 +1448,159 @@ describe('table rules', () => {
...
+     * be reached without scrolling sideways.  `.table-ellipsis` was trying to truncate under
+     * `table-layout: auto`, where a table is as wide as its columns want to be, and the
+     * `min-width: 25em` floor on the truncating column made overflowing certain.
+     *
+     * None of this is visible in jsdom -- it is all resolved widths, and the fix turns on
+     * whether our `table-layout` beats the one Mantine sets on its own class. It did not, at
+     * first: every other property in the rule applied, so the table looked fixed and stayed
+     * auto.  These measure the rendered table instead.
+     */
+    const wideTable = <table className='table-ellipsis'>
```
lrnselfreliance, line 1464: **[suggestion]** The suite comment says the tests exist to prove `table-layout: fixed` beats Mantine's `.mantine-Table-table` / `table-layout: var(--table-layout, auto)` rule. The fixture is a raw `<table className='table-ellipsis'>` with native `<th>`/`<td>` and a native checkbox — no `Table` wrapper, no `.m_b23fa0ef`, no `horizontalSpacing` padding, no `.wrolpi-th-sort`. `getComputedStyle(table).tableLayout` would be `fixed` even if the selector were the bare class that lost the cascade. The wrap test then does `const inner = th.firstElementChild; if (!inner) return;` — "Completed At" and "Control" are text nodes, so the only child checked is the checkbox. The wrap / Mantine-specificity / 2.5em-checkbox claims are therefore untested, which is why Issues 1 and 2 can ship green.

**Suggestion:** Mount `<Table className='table-ellipsis'>` with `Table.HeaderCell onSort` (so `.wrolpi-th-sort` exists) and a Mantine `Checkbox`. Keep the `tableLayout === 'fixed'` assertion — that is the one that actually fails if `table.` is dropped. Assert header button `.right` against the `th` for the narrow "Completed At" column.

app/src/components/admin/Downloads.js
```diff
@@ -592,11 +602,13 @@ export function OnceDownloadsTable({downloads, fetchDownloads}) {
                 indeterminate={selectedIds.length > 0 && selectedIds.length < (downloads?.length || 0)}
                 onChange={toggleSelectAll}
             />,
-            sortBy: null
+            sortBy: null,
+            width: '2.5em',
```
lrnselfreliance, line 606: **[suggestion]** `width: '2.5em'` is described as "a checkbox is 2.5em". `em` here is the table's sm font (~15.4px), so the column is ~38.5px border-box. Mantine cell padding is `rem(10)` each side (~11px × 2 with `--ui-scale: 1.1`). Content box ≈ 16.5px; `Checkbox` is a fixed `--checkbox-size-sm: 20px`. The box overflows the select column by a few pixels into URL. The Cypress fixture uses a native checkbox in a raw `th` with browser padding, so it does not see this.

**Suggestion:** Size the column in `rem`/`px` that include padding + 20px (e.g. `calc(20px + 1.25rem)` or `3.25em`), and measure a real `Checkbox` in a `Table.HeaderCell`.

app/src/App.css
```diff
@@ -336,12 +336,94 @@ textarea.otp {
     cursor: pointer;
 }
 
+/*
```
lrnselfreliance, line 339: **[suggestion]** The new CSS/JS comments narrate the old `auto`/`min-width: 25em` bug, the Mantine specificity fight, 676px/727px measurements, and why each property exists — dozens of lines per rule. Comments in this change restate the code and embed design history; they should be a short WHY. Same pattern at `SortableTable.js:47`, `Downloads.js:587` / `669`, `Status.js:351`, and the Cypress describe block.

**Suggestion:** Collapse each block to the one fact a later editor cannot recover from the code (e.g. `table.` beats `.mantine-Table-table`; `.col-frequency` is the 700px tablet breakpoint; `min-width: 0` lets the sort button shrink). Move the incident write-up to the PR body.

lrnselfreliance (commented):
## Summary

Switching `.table-ellipsis` to `table-layout: fixed` (with `table.` specificity to beat Mantine) is the right fix for the 540px overflow, and wiring explicit `em` widths through `SortableTable` / the Recurring and Status headers is the correct companion change. The Once-downloads and Status tables should stay inside a phone at typical widths. Two production holes remain: Recurring's four fixed columns already consume ~416px before the URL gets anything, so a 375–430px phone still loses the URL column; and `Completed At` is sized for its header, not for the in-progress `DownloadProgressModal` button that actually lives in that cell. The new Cypress tests measure a raw `<table>`, so they do not catch either issue or the Mantine cascade they document.

## Issue counts by severity

- bugs: 2
- suggestions: 3
- nits: 0

PR #661, opened 2026-08-13
test: the component suite has not run since cypress 15, on any machine

lrnselfreliance:
Six component spec files — **334 tests**, including the 4,300-line layout suite — have been unable to start since 2026-08-08. Cypress exited before printing anything, and under its default Electron browser it did not even do that: it hung at 0% CPU until killed. Chrome prints the error in seconds, which is the only reason this was findable at all.

Three things broke, all from the `13.17.0 → 15.20.0` bump (#601).

### 1. `baseUrl: "src"` — and why CI stayed green

Cypress 15 loads its config with **tsx**, and tsx honors tsconfig path resolution. Cypress requires Node's core `events` module; with `baseUrl: "src"` that bare specifier resolves against `src/`, and **on a case-insensitive filesystem it matches `src/Events.js`**. That file imports the app, so the entire component tree got required *in Node* — JSX in `.js` files, which esbuild will not parse.

That is the whole CI paradox: **Linux is case-sensitive**, `events` never matches `Events.js`, so it resolves to the built-in and e2e passes. Every macOS checkout was dead.

The baseUrl buys nothing — there is not one absolute import in the codebase — so it goes.

### 2. `framework: "create-react-app"` was removed in Cypress 14

The handler is present in the 13.17.0 binary and absent from 15:

```
13.17.0  @cypress/webpack-dev-server/dist/helpers/createReactAppHandler.js  ✓
15.20.0  — gone —
```

The webpack config is handed over explicitly now, taken from react-scripts so components compile exactly as `npm start` compiles them. Two consequences:

- **`webpack-dev-server` 5** as a devDependency — Cypress 15 requires major 5, react-scripts pins 4.15.2. npm nests react-scripts' own copy, so the production build is untouched (verified: `npm run build` clean, react-scripts still resolves 4.15.2).
- **babel-loader widened to `cypress/`** — react-scripts compiles `src` and nothing else, so the support file (which is JSX) reached babel with no preset.

### 3. Nothing ever ran the suite

`cypress run` means e2e — its `specPattern` is `cypress/e2e/**` — and these specs live beside the components in `src/`. No job has ever executed them. A `cypress-component-tests` job runs them now; without it this breaks again in silence.

### The test that was hiding

One test fails the moment the suite can run, and it is mine, from #660: it asserted the URL column is the widest in its row. It is not, and never was — at that viewport the fixed columns take ~307px of 400, leaving the URL ~93 against Control's 146. It shipped green because nothing could run it. It now asserts a usable remainder.

### How this was established

| experiment | result |
|---|---|
| `release` + cypress 13 | starts, compiles, runs a spec |
| `release` + cypress 15, same code and node | esbuild crash |
| rename 121 JSX `.js` → `.jsx` | crash advances to `window is not defined` — app being loaded in Node |
| remove `baseUrl` only, revert all else | crash gone; clean `Unexpected framework` error |

### Verification

- **334 component tests pass** across all 6 spec files (`Tags`, `Channels`, `Download`, `modal-stack`, `ui-layout` 309, `Inventory`)
- `npm run build` clean, react-scripts still on its own webpack-dev-server 4.15.2

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -168,6 +168,34 @@ jobs:
           name: Stop React App
           command: pkill -f npm
   # Jest tests moved to GitHub Actions (.github/workflows/jest-tests.yml)
+  # The component suite, which measures rendered layout and color in a real browser.
+  #
+  # It needs a job of its own because `cypress run` means e2e: its `specPattern` is
+  # `cypress/e2e/**`, and these specs live beside the components in `src/`.  So this suite ran
+  # nowhere -- not here, not in the jest job -- and stayed green in every PR while it was in
+  # fact unable to start.  Nothing here is checkable in jsdom, so "it runs in CI" is the only
+  # thing keeping it honest.
+  #
+  # No app server: component tests mount components against a dev server Cypress starts itself.
+  cypress-component-tests:
+    docker:
+      - image: cimg/node:20.19-browsers
+    resource_class: large
+    steps:
+      - checkout
+      - restore_cache:
+          key: app-browsers-{{ checksum "app/package-lock.json" }}
+      - run:
+          name: Install Dependencies
+          command: cd app && npm install
+      - run:
+          name: Install Cypress Binary
+          command: cd app && npx cypress install
+      - run:
+          name: Run Cypress Component Tests
+          command: cd app && CI=true npx cypress run --component
+      - store_artifacts:
+          path: app/cypress/screenshots
   cypress-e2e-tests:
     docker:
       - image: cimg/node:20.19-browsers
@@ -229,3 +257,4 @@ workflows:
       - react-app-start
       # jest-tests moved to GitHub Actions
       - cypress-e2e-tests
+      - cypress-component-tests
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -168,15 +168,8 @@ jobs:
           name: Stop React App
           command: pkill -f npm
   # Jest tests moved to GitHub Actions (.github/workflows/jest-tests.yml)
-  # The component suite, which measures rendered layout and color in a real browser.
-  #
-  # It needs a job of its own because `cypress run` means e2e: its `specPattern` is
-  # `cypress/e2e/**`, and these specs live beside the components in `src/`.  So this suite ran
-  # nowhere -- not here, not in the jest job -- and stayed green in every PR while it was in
-  # fact unable to start.  Nothing here is checkable in jsdom, so "it runs in CI" is the only
-  # thing keeping it honest.
-  #
-  # No app server: component tests mount components against a dev server Cypress starts itself.
+  # Component specs live under `src/`, so they need `cypress run --component`; the e2e job's
+  # `cypress run` will not match them.  Cypress starts its own dev server, so no `npm start`.
   cypress-component-tests:
     docker:
       - image: cimg/node:20.19-browsers
--- a/app/cypress.config.js
+++ b/app/cypress.config.js
@@ -1,3 +1,4 @@
+const path = require("path");
 const { defineConfig } = require("cypress");
 
 module.exports = defineConfig({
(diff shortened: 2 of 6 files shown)


app/src/components/ui/ui-layout.cy.js
```diff
@@ -1510,13 +1510,15 @@ describe('a truncating table fits the width it is given', () => {
...
-            });
+            /*
+             * ...and it still gets the leftover.  This asserted that the URL was the widest
+             * column in the row, which is not true and never was: at this width the fixed
+             * columns take about 307px of 400, so the URL gets ~93 against Control's 146.
+             * The test shipped green because the suite could not run at all.  What matters is
+             * that the column keeps a usable remainder rather than collapsing.
+             */
+            expect(cell.getBoundingClientRect().width, 'the URL keeps a usable remainder')
+                .to.be.greaterThan(30);
```
lrnselfreliance: **[suggestion]** The replacement assertion does not test what the example claims. `--ui-scale` is 1.1 (`tokens.css`), so 20em of fixed columns is ~352px of a 400px viewport and the URL remainder is ~48px — already smaller than Control. `greaterThan(30)` therefore accepts a column that kept a sliver of padding after leftover leaked elsewhere, and it would still pass if `table-layout: fixed` were lost (the URL would be huge). The preceding overflow check covers "table does not spill"; this one is supposed to cover "the ellipsis column received the leftover."

**Suggestion:** Measure leftover against the other cells in the row (table width minus the non-ellipsis columns, with a small px tolerance), or at least compare the ellipsis cell to that computed remainder. Drop the 30px magic number.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -1510,13 +1510,15 @@ describe('a truncating table fits the width it is given', () => {
...
             expect(cell.scrollWidth, 'the long URL is truncated').to.be.greaterThan(cell.clientWidth);
-            // ...and it is still the widest column, which is the point of truncating it rather
-            // than any of the others.  A 25em floor made it the widest by overflowing instead.
-            const others = [...cell.closest('tr').children].filter(td => td !== cell);
-            others.forEach(td => {
-                expect(cell.getBoundingClientRect().width, 'the URL keeps the remainder')
-                    .to.be.greaterThan(td.getBoundingClientRect().width);
-            });
+            /*
+             * ...and it still gets the leftover.  This asserted that the URL was the widest
```
lrnselfreliance: **[suggestion]** The new block comment narrates the old assertion, the pixel arithmetic that disproved it, and that the suite could not run. That is change history, not an invariant the next reader needs.

**Suggestion:** Keep one sentence of why (`table-layout: fixed` gives leftover to the unsized column; assert that remainder, not "widest"). Move the 307/93/146 post-mortem to the PR body.

app/cypress.config.js
```diff
@@ -14,8 +15,52 @@ module.exports = defineConfig({
   },
   component: {
     devServer: {
-      framework: "create-react-app",
+      /*
+       * Cypress 14 removed the built-in create-react-app dev server, so the webpack config
```
lrnselfreliance: **[suggestion]** The `devServer` comment is a Cypress 14 migration story (`Unexpected framework create-react-app`, what Cypress owns vs. what CRA owns). It restates the code below it.

**Suggestion:** One line: CRA's built-in framework was removed; this hands Cypress the same webpack config `npm start` uses, minus entry/output/html plugins Cypress replaces.

app/cypress.config.js
```diff
@@ -14,8 +15,52 @@ module.exports = defineConfig({
...
+        const cypressOwnsThese = [
+          'HtmlWebpackPlugin', 'InterpolateHtmlPlugin', 'MiniCssExtractPlugin',
+          'WebpackManifestPlugin', 'ManifestPlugin', 'ModuleNotFoundPlugin',
+          'ForkTsCheckerWebpackPlugin',
+        ];
+        config.plugins = (config.plugins || [])
+          .filter(plugin => !cypressOwnsThese.includes(plugin.constructor.name));
+
+        /*
+         * react-scripts compiles `src` and nothing else, so the support file -- which is
```
lrnselfreliance: **[suggestion]** Same pattern for the babel-loader widen: it retells that the support file "failed on its first tag" and that specs are "as much our code as `src`." The `include` mutation is already the documentation.

**Suggestion:** `react-scripts` only babel-loads `src`; the JSX support file lives in `cypress/`.

.circleci/config.yml
```diff
@@ -168,6 +168,34 @@ jobs:
           name: Stop React App
           command: pkill -f npm
   # Jest tests moved to GitHub Actions (.github/workflows/jest-tests.yml)
+  # The component suite, which measures rendered layout and color in a real browser.
```
lrnselfreliance: **[suggestion]** Nine lines of history about `cypress run` being e2e-only and the suite staying green while unable to start. Useful in the PR description; it will rot in the workflow file.

**Suggestion:** Two lines: component specs live under `src/`, so they need `cypress run --component`; Cypress starts its own dev server (no `npm start`).

lrnselfreliance (commented):
## Summary

The three fixes are the right ones and look correct: Cypress 15's tsx config load no longer resolves Node's `events` onto `src/Events.js`, the component bundler now uses CRA's webpack 5 config instead of the removed `create-react-app` framework, and CI finally invokes `cypress run --component`. No remaining `src/`-absolute imports depend on `baseUrl`; the lockfile hoists webpack-dev-server 5.2.6 for Cypress and correctly nests 4.15.2 under react-scripts for `npm start`; babel-loader is widened only on CRA's app-JS rule (`loader` + `include`), which is the rule that compiles the JSX support file. Dominant leftover risk is the new URL-column assertion being too loose to catch a leftover-width regression, plus several new comments that narrate the outage rather than the invariant.

## Issue counts by severity

- bugs: 0
- suggestions: 5
- nits: 0

PR #662, opened 2026-08-13
review: the 375px phone this fix had not been measured on

lrnselfreliance:
Answers the five review findings on #660, rebased onto #661 so its Cypress tests actually execute.

Master currently carries a regression: at a real 375px viewport the recurring table is **417px wide with the URL column at exactly 0**. #660 fixed a 540px overflow and created a 375px one.

### The five findings

| | |
|---|---|
| **Recurring starved the URL** | Its four fixed columns wanted 416px between them. Widths are phone-first now, in `col-*` classes, widening at the tablet breakpoint. |
| **Progress button overflowed** | `"5 MBps / 3m44s"` measured 169–188px against 124px of column. It shows the percent; the modal already lists speed, ETA and bytes, and the full text is on the button's `title`. |
| **Checkbox overflowed its column** | 2.5em gave 16.5px of content for a 22px checkbox. 3em holds the checkbox and its padding. |
| **Fixtures proved nothing** | A raw `<table>` reports `tableLayout: fixed` even with the selector that loses to Mantine — the test could not fail for the reason it claimed. They mount `SortableTable`, `Table.HeaderCell onSort` and a real `Checkbox` now. |
| **Comments** | Trimmed to what cannot be recovered from the code. |

### Two defects the restored suite caught

Neither was findable while the component suite could not start:

**A sized column that did not fit its own label.** Recurring's Control column was 4.5em — 70px of content for a 78px "Control". Now 5.2em, sized for the word rather than for the button it holds.

**A header test that measured the wrong thing.** The elastic URL column ends up narrower than its own header ("URL" plus its sort arrow needs 72px against 58), and one word cannot wrap, so the column clips. My first assertion measured the *text* width, missed the arrow and the gap, and passed while the button overran by 14px. The test now states the invariant per kind of column: a sized column must fit its label; only the elastic column may clip.

### Measured at 375px

```
once       [x]46  URL 58  Completed 123  Control 146
recurring  URL 33  Frequency 100  Completed 104  Next 57  Control 80
```

Nothing overflows, no sized header is clipped. The recurring URL column is 33px there — that is the accepted trade for keeping every column on a phone rather than hiding any.

### Verification

- **336 component tests pass** (the two new ones included) — for the first time these run at all
- `npm run build` clean

### DIFF
--- a/app/src/components/SortableTable.js
+++ b/app/src/components/SortableTable.js
@@ -44,24 +44,22 @@ export class SortableTable extends React.Component {
         const data = this.props['data'] ? this.sortData(this.props['data']) : null;
 
         const tableHeader = (spec) => {
-            // `width` is how a column that must fit something exact -- a checkbox, a group of
-            // buttons -- states its size.  Under `table-layout: fixed` (see `.table-ellipsis`)
-            // the header cell is where a column's width is declared; columns that name none
-            // divide what is left, which is what the truncating column wants.
-            const {key, text, sortBy, width} = spec;
-            const style = width ? {width} : undefined;
+            // Under `table-layout: fixed` the header cell is where a column's width is
+            // declared; a column that names none divides what is left.  `className` rather
+            // than a width so the value can change at a breakpoint.
+            const {key, text, sortBy, className} = spec;
             if (sortBy) {
                 const active = sortColumn === key;
                 return <Table.HeaderCell
                     key={key}
-                    style={style}
+                    className={className}
                     sorted={active ? direction : null}
                     onSort={() => this.changeSort(key)}
                 >
                     {text}
                 </Table.HeaderCell>;
             } else {
-                return <Table.HeaderCell key={key} style={style}>{text}</Table.HeaderCell>
+                return <Table.HeaderCell key={key} className={className}>{text}</Table.HeaderCell>
             }
         }
 
--- a/app/src/components/admin/Downloads.js
+++ b/app/src/components/admin/Downloads.js
@@ -58,12 +58,8 @@ function DownloadProgressModal({progress, url}) {
 
     const speedText = humanBandwidth(progress.speed);
     const etaText = progress.eta || '...';
-    /*
-     * A percent, not "5 MBps / 3m44s".  This button lives in the Completed At column, which is
-     * sized for its header; the speed-and-ETA label measured 169-188px against 124px of column
-     * and was clipped by the cell beside it.  Nothing is lost by shortening it -- the modal it
-     * opens already lists speed, ETA and bytes, which is where a reader goes for the detail.
-     */
+    // A percent: this button sits in a column sized for its header, and speed-and-ETA does not
+    // fit there.  The modal it opens carries the detail, and `title` the full text.
     const buttonLabel = Number.isFinite(progress.percent) ? `${Math.round(progress.percent)}%` : '...';
 
     return <>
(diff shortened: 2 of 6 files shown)


app/src/components/admin/Downloads.js
```diff
@@ -58,10 +58,17 @@ function DownloadProgressModal({progress, url}) {
 
     const speedText = humanBandwidth(progress.speed);
     const etaText = progress.eta || '...';
-    const buttonLabel = `${speedText} / ${etaText}`;
+    /*
+     * A percent, not "5 MBps / 3m44s".  This button lives in the Completed At column, which is
```
lrnselfreliance: **[suggestion]** The percent-button comment still narrates the old measurement (“169-188px against 124px”) and the design history of the modal. Same class of leftover in App.css (the URL-clip story at the `overflow: hidden` rule, the 23px ButtonGroup story) and in the Cypress header-fit block (hard-coded “50px against 36px”).

**Suggestion:** Keep one clause of why (percent in a header-sized column; clip when a word cannot wrap; do not shrink ButtonGroup children). Drop the measured-px archaeology.

app/src/components/ui/ui-layout.cy.js
```diff
@@ -1450,158 +1452,198 @@ describe('table rules', () => {
...
+                         * own label -- "URL" plus its sort arrow is 50px against 36px of
+                         * column here, and one word cannot wrap -- so what holds it in is the
+                         * clip, not the fit.
+                         *
+                         * `scrollWidth` against `clientWidth` measures the content either way,
+                         * whether the label is a text node or the flex button a sortable header
+                         * puts it in, and it still reports the overflow once the cell clips it.
+                         */
+                        if (th.className.includes('col-')) {
+                            expect(th.scrollWidth,
```
lrnselfreliance: **[suggestion]** Sized columns are checked with `th.scrollWidth <= th.clientWidth + 1`. That compares against the padding box. The desktop Frequency test correctly uses `oneLineWidth` vs `contentWidth` (padding subtracted). `col-control-row` is 5.2em ≈ 80px; after 11px padding each side the text box is ~58px. A label that is 60–78px can pass this assertion and still sit in the padding or be clipped by `.table-ellipsis th { overflow: hidden }`. That is the same class of miss the 4.5em Control failure was meant to close.

**Suggestion:** For `col-*` headers, assert `oneLineWidth(th)` (or the sort button’s scroll width) against `contentWidth(th)`, matching the desktop Frequency check.

app/src/components/SortableTable.js
```diff
@@ -44,24 +44,22 @@ export class SortableTable extends React.Component {
...
-            // `width` is how a column that must fit something exact -- a checkbox, a group of
-            // buttons -- states its size.  Under `table-layout: fixed` (see `.table-ellipsis`)
-            // the header cell is where a column's width is declared; columns that name none
-            // divide what is left, which is what the truncating column wants.
-            const {key, text, sortBy, width} = spec;
-            const style = width ? {width} : undefined;
+            // Under `table-layout: fixed` the header cell is where a column's width is
+            // declared; a column that names none divides what is left.  `className` rather
+            // than a width so the value can change at a breakpoint.
+            const {key, text, sortBy, className} = spec;
```
lrnselfreliance, line 50: **[suggestion]** Header specs no longer apply `width` as a style. Downloads migrated to `className`, but Tags.js, Zim.js, and CollectionTable.js still pass numeric `width` (1, 2, 4, 14). Those values were already treated as CSS px (Semantic leftover), and none of those tables use `table-ellipsis`, so this is not a visible regression. The prop is now silently ignored.

**Suggestion:** Keep applying `style={{width}}` when `width` is set, or delete the dead keys at the call sites so the API has one way to size a column.

PR #663, opened 2026-08-13
ui: an upload target smaller than the dialog around it

lrnselfreliance:
The upload dialog was **682×367 on a 1440×900 desktop** — 47% of the width, 41% of the height — and the drop target inside it was a strip about 90px tall. A small thing to drag a file onto, with the file browser it came from covering most of the screen behind it.

### What changed

**Upload** and **Move** now pass `fullScreen`: edge to edge on a phone, **90% above the tablet breakpoint**. Filling a desktop entirely reads as the app having been replaced rather than as something opened over it, so the page still frames the dialog.

Move earns it on content — a `DirectorySearch` above a row per path being moved, which grows with the selection. 620px showed a handful of rows and scrolled the rest.

Two things had to follow:

- **The footer floated.** Mantine sizes a fullscreen dialog's content to the viewport but leaves the body at its content height, so the actions landed two thirds up the screen with a quarter of the viewport empty beneath. The body takes the leftover now and the actions ride to the bottom of it. Keyed off Mantine's own `data-full-screen`, so it follows the prop rather than a class every call site has to remember.
- **The drop target grows into the room**, with a `min-height` floor so it stays large once the progress bars claim their own space. It went from ~90px to 518px tall on a desktop.

### What deliberately did not change

Rename and the delete/ignore confirmations keep their sizes — a fullscreen dialog for one text field or one sentence reads as broken.

**New Folder also keeps `large`.** When I offered the options I described it as a directory picker; it is not. It is a `<pre>` path preview and a `TextInput` — the same shape as Rename. Say the word and I will make it fullscreen too, but it looked like an error in my own description rather than a decision worth acting on.

### A naming trap, documented

`size='fullscreen'` and the `fullScreen` prop are **not** the same thing, and both exist:

- `size='fullscreen'` → full **width**, height still from the content. Right for the 12 dialogs already using it (logs, previews, download progress) — several are short, and stretching them to 100vh would leave them mostly empty.
- `fullScreen` → fills the viewport, subject to the 90% cap above.

Both are now explained where the sizes are declared.

### Verification

Measured in the browser at both sizes:

```
1440x900   modal 1296x810  (90% x 90%, centred: 72px / 45px margins)
           drop target 1224x518, actions flush to the foot
390x844    modal 390x844   (100% x 100%, at 0,0)
```

Three new Cypress tests cover it, and each is load-bearing — removing the CSS fails two of them, and the phone test fails if the wrapper stops forwarding `fullScreen`:

```
90% of the width: expected 1440 to be close to 1296 +/- 2
the actions sit at the foot of the dialog: expected 702.375 to be at most 2
```

The gallery grows a fullscreen example, checked in all four themes (1296×810, correct surface token, actions at the foot in each).

- 339 component tests pass
- `npm run build` clean

### DIFF
--- a/app/src/components/FileBrowser.js
+++ b/app/src/components/FileBrowser.js
@@ -710,11 +710,11 @@ export function MoveModal({open, paths, onClose, onSubmit}) {
     </Table.Row>);
 
     /*
-     * Audited: `large` (620px).  This dialog holds a DirectorySearch and a two-column
-     * Source/Destination table of every path being moved, which is the densest content
-     * of the five FileBrowser modals.
+     * Fullscreen: a DirectorySearch above a two-column Source/Destination table of every
+     * path being moved, which grows with the selection.  620px showed a handful of rows and
+     * scrolled the rest.
      */
-    return <Modal size='large'
+    return <Modal fullScreen
                   onClose={onClose}
                   open={open}
     >
@@ -925,7 +925,12 @@ export function FileBrowserUploadModal({open, onClose, destination, onComplete})
 
     const displayPath = destination ? `${mediaDirectory}/${destination}` : mediaDirectory;
 
-    return <Modal open={open} onClose={handleClose} size='large'>
+    /*
+     * Fullscreen, for the drop target.  At 620px the target was a strip about 90px tall in a
+     * dialog covering a fifth of the screen -- a small thing to drag a file onto, and the
+     * page behind it was the file browser it came from.  Given the room, the target takes it.
+     */
+    return <Modal open={open} onClose={handleClose} fullScreen>
         <Modal.Header>Upload to: {displayPath}</Modal.Header>
         <Modal.Content>
             <form onSubmit={e => e.preventDefault()}>
@@ -940,8 +945,26 @@ export function FileBrowserUploadModal({open, onClose, destination, onComplete})
                 </div>
             </form>
 
-            <Panel style={{marginTop: '1em', padding: '1em'}}>
-                <div {...getRootProps()} style={{cursor: 'pointer', textAlign: 'center'}}>
+            {/* The target takes the room the dialog gives it rather than sitting as a strip
+                at the top of it.  `min-height` as well as `flex`, so it stays a large target
+                once the progress bars below it start claiming their own space. */}
+            <Panel style={{
+                marginTop: '1em',
+                padding: '1em',
+                flex: '1 1 auto',
+                display: 'flex',
+                flexDirection: 'column',
+            }}>
+                <div {...getRootProps()}
+                     style={{
+                         cursor: 'pointer',
+                         textAlign: 'center',
+                         display: 'flex',
+                         alignItems: 'center',
+                         justifyContent: 'center',
+                         flex: '1 1 auto',
+                         minHeight: '30vh',
+                     }}>
                     <input {...getInputProps()}/>
                     <Header as='h4' icon='file text'>Click here, or drop files here to upload</Header>
                 </div>
--- a/app/src/components/ui/ui.css
+++ b/app/src/components/ui/ui.css
@@ -494,6 +494,48 @@ html .wrolpi-searchbox-clear {
     border-top: 1px solid var(--border);
 }
 
+/*
+ * A fullscreen dialog fills its height as well as its width.
+ *
+ * Mantine gives the content 100vh but leaves the body at its content height, so the footer
+ * landed wherever the content ended -- on the upload dialog, two thirds up the screen with a
+ * quarter of the viewport empty beneath it.  The body takes the leftover instead, and the
+ * actions ride to the bottom of it.
+ *
+ * `data-full-screen` is Mantine's own marker, so this follows the prop rather than a class a
+ * call site has to remember.
+ */
+.mantine-Modal-content[data-full-screen] {
+    display: flex;
+    flex-direction: column;
+}
+
+/*
+ * Edge to edge is for a phone, where the alternative is a dialog with a few pixels of page
+ * showing around it.  On a desktop it reads as the app having been replaced rather than as
+ * something opened on top of it, so the dialog stops at 90% and leaves the page framing it.
+ *
+ * 700px is the app's `tablet` breakpoint (`AppMedia` in contexts.js).
+ */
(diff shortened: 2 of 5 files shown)


app/src/components/ui/ui.css
```diff
@@ -494,6 +494,48 @@ html .wrolpi-searchbox-clear {
...
+@media (min-width: 700px) {
+    .mantine-Modal-content[data-full-screen] {
+        width: 90vw;
+        max-width: 90vw;
+        height: 90vh;
+        max-height: 90vh;
+    }
+}
+
+.mantine-Modal-content[data-full-screen] .mantine-Modal-body {
```
lrnselfreliance, line 529: **[suggestion]** The fullscreen body is now a column flex (`flex: 1; display: flex; flex-direction: column`) with no `min-height: 0` / `overflow` and no growing, scrolling slot for Move’s Source/Destination table. Short content pins `.wrolpi-modal-actions` to the foot, which the new Cypress test covers. Once `paths` is long enough that the table’s min-content height exceeds the leftover, the body’s `min-height: auto` expands past the 90vh/100dvh content box and the whole dialog — including Reset/Move — scrolls. That is the case Move was promoted to `fullScreen` for.

**Suggestion:** On `.mantine-Modal-content[data-full-screen] .mantine-Modal-body` set `min-height: 0; overflow: auto` (or `min-height: 0` and put `flex: 1; min-height: 0; overflow: auto` on a wrapper around Move’s `<Table>`). Then the path list uses leftover height and the actions stay on screen. `min-width: 0` on the same body rule would also keep `.wrolpi-table-scroll`’s `overflow-x: auto` bounded now that the table wrapper is a flex item.

app/src/components/FileBrowser.js
```diff
@@ -940,8 +945,26 @@ export function FileBrowserUploadModal({open, onClose, destination, onComplete})
                 </div>
             </form>
 
-            <Panel style={{marginTop: '1em', padding: '1em'}}>
-                <div {...getRootProps()} style={{cursor: 'pointer', textAlign: 'center'}}>
+            {/* The target takes the room the dialog gives it rather than sitting as a strip
+                at the top of it.  `min-height` as well as `flex`, so it stays a large target
+                once the progress bars below it start claiming their own space. */}
+            <Panel style={{
```
lrnselfreliance, line 951: **[suggestion]** The drop target grows only because `Modal.Content` is a fragment (`<>{children}</>` in Overlays.tsx), so the `Panel` with `flex: '1 1 auto'` is a direct child of `.mantine-Modal-body`. The new tests mount a `<p>drop files here</p>`, not this flex chain, so they stay green if `Modal.Content` later becomes a wrapper (the unused `scrolling` prop is the obvious reason someone would add one) or if the inline `flex`/`minHeight` is removed. The 90px → 518px win is the reason this dialog went fullscreen, and nothing asserts it.

**Suggestion:** Keep the flex on a class under `.mantine-Modal-content[data-full-screen]` rather than on this Panel, or unwrap `Modal.Content`’s children the way Header/Actions already are, so growth does not depend on the fragment. Add a Cypress mount of `FileBrowserUploadModal` (or a fixture with the same Panel/flex markup) that checks the drop target’s height is a large fraction of the dialog, and still above `30vh` after a couple of progress rows.

app/src/components/ui/ui.css
```diff
@@ -494,6 +494,48 @@ html .wrolpi-searchbox-clear {
     border-top: 1px solid var(--border);
 }
 
+/*
+ * A fullscreen dialog fills its height as well as its width.
```
lrnselfreliance, line 498: **[suggestion]** Several new comments narrate the before/after of this PR (90px strip, footer two thirds up, 620px showed a handful of rows) instead of stating the standing constraint. The same story appears in `ui.css`, both FileBrowser call sites, and `modal-stack.cy.js`. The surrounding files already run long, but these blocks will be stale the moment the measurements change and they restate what the selectors do.

**Suggestion:** Cut each to the non-obvious constraint. Here: body fills leftover height so actions sit at the bottom; keyed off Mantine’s `data-full-screen`. In FileBrowser, “fullscreen so the drop target / path table can use leftover height” is enough. Leave the `size='fullscreen'` vs `fullScreen` note on `modalSizes` — that naming trap is the one comment a later caller needs.

app/src/components/ui/ui.css
```diff
@@ -494,6 +494,48 @@ html .wrolpi-searchbox-clear {
...
+ * showing around it.  On a desktop it reads as the app having been replaced rather than as
+ * something opened on top of it, so the dialog stops at 90% and leaves the page framing it.
+ *
+ * 700px is the app's `tablet` breakpoint (`AppMedia` in contexts.js).
+ */
+@media (min-width: 700px) {
+    .mantine-Modal-content[data-full-screen] {
+        width: 90vw;
+        max-width: 90vw;
+        height: 90vh;
```
lrnselfreliance, line 524: **[nit]** The 90% cap uses `90vh` / `90vw`. Mantine’s own fullscreen height is `100dvh`, and `App.css` already prefers `dvh` so mobile/tablet chrome is counted. The cap applies from 700px up, so iPad-class Safari is in scope: `90vh` can be taller than the `position: fixed; top: 0; bottom: 0` inner, and the dialog will not sit in a true 5% frame.

**Suggestion:** Use `90dvh` / `90dvw` (or `90%` of the inner) to match Mantine and the rest of the app.

lrnselfreliance (commented):
## Summary

The change is correct. `fullScreen` is not among the props `ModalBase` pulls out, so it reaches Mantine through `{...props}`; `size='fullscreen'` still maps to width `100%` and never sets `data-full-screen`, so the twelve existing full-width dialogs do not pick up the 90% cap or footer pinning. Upload and Move are the only FileBrowser callers flipped; Rename (`small`), New Folder (`large`), Ignore (`small`), and delete confirmations stay as they were. Dominant risk is the new flex column on the fullscreen body: the drop target’s growth depends on `Modal.Content` remaining a fragment, and a long Move table is not given its own scroll region, so the newly pinned footer still scrolls away once the path list exceeds the leftover height.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 1

PR #664, opened 2026-08-13
ui: two upload dialogs that had drifted apart

lrnselfreliance:
The dashboard's Upload dialog was still `size='large'` with a ~90px drop strip, because it was a second copy of the file browser's upload form rather than the same one. Two dropzones, two overwrite toggles, two copies of the progress-bar rendering, both driving the same `useUploadFile` state — so last week's fix to the file browser's dialog could not reach the dashboard's.

There is now one `UploadModal`, and both entry points open it.

## The one difference that is real

The file browser knows where the files go, because the user selected the folder before pressing Upload. The dashboard has to ask. That is the `destination` prop:

| `destination` | behavior |
| --- | --- |
| `undefined` | ask, with a `DirectorySearch`; the drop target is replaced by the instruction to pick one |
| a string | upload there, and name it in the title |
| `''` | the media directory — a real destination, not a missing one |

That last row is the trap worth naming: the file browser passes `''` when nothing is selected, so a falsy check would send that user to a directory search instead of the drop target. A test pins it.

## Measured

Dashboard, 1440×900:

```
before   modal ~700×~350   drop target a strip
after    modal 1296×810    drop target 1261×481
```

File browser, same viewport: modal 1296×810, title `Upload to: /media/wrolpi/archive`, no destination search, actions flush to the foot. The 90%-desktop / edge-to-edge-phone sizing is the CSS that already shipped; nothing here changes it.

## Layout note

`UploadForm` returns a fragment rather than a wrapping element, and the `<form>` is around the fields that have one. The drop target grows into the dialog's leftover height, which it can only do while its parent is the modal body's flex column — a `<div>` or a `<form>` around the whole form absorbs the growth and the target silently goes back to its content height. That is how the first attempt here failed, and the comment says so.

## Tests

- `Upload.test.js`, new: three cases covering the ask/fixed/media-root modes.
- Full suites: 70 jest suites / 1217 tests, 339 Cypress component tests, `npm run build` clean.

### DIFF
--- a/app/src/components/Upload.js
+++ b/app/src/components/Upload.js
@@ -6,19 +6,11 @@ import {useMediaDirectory, useUploadFile} from "../hooks/customHooks";
 import _ from "lodash";
 
 /*
- * One upload form, two entry points.
+ * `destination` says where the files go:
  *
- * The dashboard asks the user where the files should go; the file browser already knows,
- * because the user selected the directory before pressing Upload.  That is the ONLY difference
- * between them, and it is `destination`:
- *
- *   undefined -- ask, with a DirectorySearch.  Until one is chosen there is nothing to drop
- *                files onto, so the target is replaced by the instruction to pick one.
- *   a string  -- fixed, and shown in the title.  '' is the media directory itself, which is a
- *                real destination; only `undefined` means "ask".
- *
- * Everything else -- the dropzone, the tag selector, the overwrite toggle, the per-file and
- * overall progress -- is the same work against the same `useUploadFile` state.
+ *   undefined -- ask, with a DirectorySearch.
+ *   a string  -- upload there.  '' is the media directory, a real destination; a falsy check
+ *                against it reads the media root as "nowhere".
  */
 
 export function UploadForm({
@@ -57,10 +49,9 @@ export function UploadForm({
     const canDrop = !askForDestination || !!destination;
 
     /*
-     * A fragment, not a wrapper element.  The drop target below grows into whatever height the
-     * dialog has left, which it can only do while its parent is the modal body's flex column --
-     * a `<div>` or a `<form>` around all of this would absorb the growth and leave the target
-     * its content height again.  The form is around the fields that have one.
+     * A fragment, and the `<form>` around only the fields.  The drop target below grows into
+     * the dialog's leftover height, which requires its parent to be the modal body's flex
+     * column: any element wrapping all of this absorbs the growth instead.
      */
     return <>
         <form onSubmit={e => e.preventDefault()}>
@@ -86,9 +77,8 @@ export function UploadForm({
         </Stack>
         </form>
 
-        {/* The target takes the room the dialog gives it rather than sitting as a strip at the
-            top of it.  `min-height` as well as `flex`, so it stays a large target once the
-            progress bars below it start claiming their own space. */}
+        {/* The target grows into the dialog's leftover height.  `min-height` as well as
+            `flex`, so it stays a large target once progress bars claim their own space. */}
         {canDrop ?
             <Panel style={{
                 marginTop: '1em',
@@ -111,8 +101,8 @@ export function UploadForm({
                     <Header as='h4' icon='file text'>Click here, or drop files here to upload</Header>
                 </div>
             </Panel>
-            // The instruction takes the same room the target would, so it sits where the user
-            // is about to look rather than as a line above two thirds of an empty dialog.
+            // The instruction takes the room the target would, rather than sitting as a line
+            // above an empty dialog.
             : <Header as='h4' style={{
                 flex: '1 1 auto', display: 'flex', alignItems: 'center', justifyContent: 'center',
             }}>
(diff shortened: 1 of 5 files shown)


app/src/components/Upload.js
```diff
@@ -73,39 +74,123 @@ export function Upload({disabled}) {
...
+    // A fixed destination is the caller's to declare; only the searching form sets its own.
+    React.useEffect(() => {
+        if (open && !asksForDestination) {
+            setDestination(destination);
+        }
+    }, [open, destination, asksForDestination, setDestination]);
+
+    // Destination was cleared, so the queued files have nowhere to go.
+    React.useEffect(() => {
+        if (!upload.destination && open) {
```
lrnselfreliance: **[bug]** `if (!upload.destination && open) doClear()` treats `''` (the media directory, which this PR documents as a real destination) the same as “no destination”. `doClear` in `useUploadFile` is not wrapped in `useCallback` and always does `setProgresses({})` / `setFiles([])` (new references), so `doClear` changes every render and this effect re-fires. Opening the dashboard modal never calls `setDestination` (the first effect is skipped when `asksForDestination`), so hook state stays `''` and React hits “Maximum update depth exceeded”. The same loop happens in the file browser when nothing is selected (`destination=''`), which is the enabled default Upload path. The tests hide this because they mock `doClear` as a stable `jest.fn()`. Separately, even if the loop is fixed, `useUploadFile`’s `doUpload` still bails on `!destination`, so a media-root drop would not start an upload.

**Suggestion:** Only clear queued files when the user is searching and the destination is actually cleared, and do not depend on an unstable `doClear` (omit it from the deps, or memoize it). Something like `if (asksForDestination && !upload.destination && open)` plus a stable `doClear` — or clear only on a `''`/`undefined` transition, not on every render while it is empty. Do not use a falsy check against hook `destination` when `''` is a valid fixed destination. If `''` is meant to be a working upload target, `doUpload`’s `if (!destination) return` in `useUploadFile` also needs a matching change.

app/src/components/Upload.test.js
```diff
@@ -0,0 +1,70 @@
...
+        expect(mockUseUploadFile.setDestination).toHaveBeenCalledWith('videos');
+    });
+
+    /*
+     * '' is the media directory, which is where the file browser uploads to when no folder is
+     * selected.  Only `undefined` means "ask" -- a falsy check here would send that user to a
+     * DirectorySearch instead of the drop target.
+     */
+    it('treats the media directory as a destination, not a missing one', () => {
+        render(<UploadModal open={true} onClose={() => {}} destination=''/>);
```
lrnselfreliance, line 63: **[suggestion]** The three destination cases pin DirectorySearch vs dropzone visibility, which is the `asksForDestination` fork, but they do not pin the rest of the `''` contract. The media-root test never asserts the title (`Upload to: …`) or `setDestination('')`; a falsy check in `displayPath` or in the `setDestination` effect would still pass. The mock `doClear` is identity-stable, so the infinite-loop in Issue 1 cannot fail this file. Close/reset (`handleClose` / `onComplete`), overwrite, and progress are untested.

**Suggestion:** Assert `setDestination` was called with `''` and that the header is `Upload to: /media/wrolpi` (not the dashboard title and not a trailing slash). Add a case that `doClear` is not invoked in a loop when `destination=''`. Close/reset can stay thin; the destination title/`setDestination('')` assertions are the ones that actually lock the trap this file claims to pin.

app/src/components/Upload.js
```diff
@@ -73,39 +74,123 @@ export function Upload({disabled}) {
...
+        if (inProgress) return;
+        doClear();
+        onClose();
+        if (onComplete) onComplete();
+    };
+
+    const displayPath = asksForDestination ? null : `${mediaDirectory}/${destination}`;
+
+    /*
+     * Fullscreen, for the drop target.  At `large` the target was a strip about 90px tall in a
```
lrnselfreliance: **[suggestion]** Several new comments narrate the old two-dialog split and the `#663` `large`/90px-strip layout rather than a current invariant. The worst is the fullscreen block (“At `large` the target was a strip about 90px tall…”). The module banner (lines 8–22), the drop-target strip comments (89–91, 114–115), the dashboard “same dialog the file browser opens” note, and the test file’s describe blurb do the same. The fragment comment (59–64) is the one that is actually load-bearing (wrapping `UploadForm` would break `.mantine-Modal-body` flex growth) and can stay if shortened.

**Suggestion:** Keep a one-liner where the invariant is non-obvious (`destination === undefined` means ask; `''` is the media root; `UploadForm` must stay a fragment). Drop the history of the other dialog, the 90px strip, and “why they drifted”.

app/src/components/Upload.js
```diff
@@ -73,39 +74,123 @@ export function Upload({disabled}) {
...
+    }, [upload.destination, open, doClear]);
+
+    const handleClose = () => {
+        if (inProgress) return;
+        doClear();
+        onClose();
+        if (onComplete) onComplete();
+    };
+
+    const displayPath = asksForDestination ? null : `${mediaDirectory}/${destination}`;
```
lrnselfreliance: **[nit]** `displayPath` is always `` `${mediaDirectory}/${destination}` `` when a destination is provided. For media-root `''` that becomes `/media/wrolpi/` with a trailing slash. The previous file-browser modal used `destination ? \`${mediaDirectory}/${destination}\` : mediaDirectory`, and the theme sample header is `Upload to: /media/wrolpi`.

**Suggestion:** Use the old join (skip the slash when `destination` is empty) so the title stays `Upload to: ${mediaDirectory}` for the media root.

PR #665, opened 2026-08-13
ui: a download form that emptied itself after every submit

lrnselfreliance:
Submitting a download reset the entire form, so a user adding several downloads re-chose the tags, destination and video settings every time — even though the next download almost always wants the same ones.

Now a successful submit empties **only the URLs**. Everything else survives, and a **Clear** button beside Download resets the form when the next download is unrelated.

Measured on the dashboard, submitting twice in a row:

| | URLs | Videos tag | Audio only |
| --- | --- | --- | --- |
| after 1st submit | `''` | Radio | on |
| after 2nd submit | `''` | Radio | on |
| after Clear | `''` | — | off |

Both downloads reached the API. The second was submittable as soon as a URL was typed — no reopening the form.

## How

`useForm` gains `clearPathsOnSuccess`, which puts the named paths back to their default and leaves the rest alone. `clearOnSuccess` is unchanged, so nothing that wants a full reset moves.

The six create-mode download forms (video, channel, archive, RSS, files, scrape) pass `['url', 'urls']` — both spellings, because a single-URL form has `url` and a textarea form has `urls`, and a path the form does not declare is skipped rather than invented. Edit forms already passed `clearOnSuccess={false}` and are untouched.

One thing worth flagging: `VideosDownloadForm` had no `emptyFormData`, so its "empty" was `defaultFormData` — which is merged with the seed URL the browser extension supplies. Clearing against that would have put the just-submitted URL back in the field. It now passes the blank form explicitly.

## Tests

- `useForm.test.js`: clears only the named paths; does not invent a path the form lacks.
- `Download.cy.js`: URLs emptied and settings kept; ready for the next download as soon as one is typed; Clear resets.

The Cypress tests watch **Audio only**, deliberately — Compress and Video format are remembered in localStorage and would survive a full reset too, so they would have proved nothing. Negative control: reverting the video form to `clearOnSuccess` fails the preservation test (`expected input to be 'checked'`).

Suites: 1221 jest, 342 Cypress component, `npm run build` clean.

### DIFF
--- a/app/src/components/Download.js
+++ b/app/src/components/Download.js
@@ -48,12 +48,8 @@ import {
  * leaves out of scope (hooks/useForm.js itself is unmigrated).
  */
 
-/*
- * What a successful submit empties.  Everything else -- tags, destination, frequency, video
- * settings -- survives, so a user adding downloads one after another types only the URL.
- * Both spellings: the forms use `url` for a single download and `urls` for a textarea, and
- * `clearPaths` ignores a path the form does not have.
- */
+// Both spellings: a single-URL form has `url`, a textarea form has `urls`, and `clearPaths`
+// skips a path the form does not declare.
 const URL_PATHS = ['url', 'urls'];
 
 function FieldLabel({htmlFor, label, required, helpContent, helpPosition = 'top'}) {
@@ -312,11 +308,6 @@ export function TitleExclusionInput({form, path = 'settings.title_exclude'}) {
 }
 
 export function DownloadFormButtons({onCancel, form}) {
-    /*
-     * Clear is the user's, not the form's.  A successful submit empties only the URLs, so the
-     * tags, destination and settings behind them survive for the next download; this is the
-     * button that puts those back to their defaults when the next download is unrelated.
-     */
     return <Group justify='space-between' mt='sm'>
         <Button role='cancel' onClick={onCancel} type='button'>Cancel</Button>
         <Group gap='sm'>
@@ -1470,7 +1461,6 @@ export function SuffixFormInput({form, name = 'suffix', path = 'settings.suffix'
 export function ScrapeFilesDownloadForm({
                                             download,
                                             submitter,
-                                            clearOnSuccess,
                                             onDelete,
                                             onCancel,
                                             onSuccess,
@@ -1513,7 +1503,8 @@ export function ScrapeFilesDownloadForm({
         submitter,
         defaultFormData: mergeDeep(emptyFormData, download),
         emptyFormData,
-        clearPathsOnSuccess: clearOnSuccess ? URL_PATHS : null,
+        // `onSuccess` is only given by the edit form, which keeps what it was given.
+        clearPathsOnSuccess: onSuccess ? null : URL_PATHS,
         onSuccess: localOnSuccess,
     });
 
--- a/app/src/components/Download.js
+++ b/app/src/components/Download.js
@@ -48,6 +48,14 @@ import {
  * leaves out of scope (hooks/useForm.js itself is unmigrated).
  */
 
+/*
+ * What a successful submit empties.  Everything else -- tags, destination, frequency, video
+ * settings -- survives, so a user adding downloads one after another types only the URL.
+ * Both spellings: the forms use `url` for a single download and `urls` for a textarea, and
+ * `clearPaths` ignores a path the form does not have.
+ */
+const URL_PATHS = ['url', 'urls'];
+
 function FieldLabel({htmlFor, label, required, helpContent, helpPosition = 'top'}) {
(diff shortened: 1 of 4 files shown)


app/src/components/Download.js
```diff
@@ -1486,7 +1513,7 @@ export function ScrapeFilesDownloadForm({
         submitter,
         defaultFormData: mergeDeep(emptyFormData, download),
         emptyFormData,
-        clearOnSuccess,
+        clearPathsOnSuccess: clearOnSuccess ? URL_PATHS : null,
```
lrnselfreliance: **[bug]** `ScrapeFilesDownloadForm` maps `clearOnSuccess ? URL_PATHS : null`, but `clearOnSuccess` has no default and neither `DownloadMenu` nor `EditScrapeFilesDownloadForm` pass it. `useForm` used to default that flag to `false`, so create-mode scrape still does not empty URLs after a successful submit. The other five create forms do; a user adding several scrapes keeps the last URLs and can resubmit them. Edit scrape happens to stay correct only because the flag is still falsy — there is no explicit `clearOnSuccess={false}` the way Channel/RSS do.

**Suggestion:** Give create-mode scrape the same behavior as the other forms. Either default `clearOnSuccess` to `true` and pass `clearOnSuccess={false}` from `EditScrapeFilesDownloadForm`, or follow the Videos/Archive pattern (`clearPathsOnSuccess: onSuccess ? null : URL_PATHS`). `emptyFormData` only declares `urls`; if edit ever clears, add `url: ''` so the single-URL field restores from defaults instead of the `''` fallback.

app/src/hooks/useForm.js
```diff
@@ -111,6 +112,25 @@ export function useForm({
...
+     * tags, destination and settings, and empties only what identifies the last submission.
+     * A path the defaults do not mention clears to '', because the field it belongs to is
+     * controlled and `undefined` would hand it back to the DOM.
+     */
+    const clearPaths = (paths) => {
+        const defaults = emptyFormData || defaultFormData || {};
+        setFormData(prev => {
+            const next = _.cloneDeep(prev);
+            paths.forEach(path => _.set(
+                next, path, _.has(defaults, path) ? _.cloneDeep(_.get(defaults, path)) : '',
```
lrnselfreliance: **[suggestion]** `clearPaths` always `_.set`s every requested path. A path missing from defaults is written as `''`, not skipped. That contradicts `Download.js:55` (`clearPaths` “ignores a path the form does not have”) and the test named “does not invent a path the form does not have”, whose assertion (`formData.url === ''`) actually locks in invention. After a urls-only submit the form gains a leftover `url: ''` (and the reverse on url-only forms), so `dirty` stays true even when every declared field is back to its default. Nothing user-facing reads `dirty` on these forms today, and submitters pick named fields, so this is not a functional break.

**Suggestion:** Skip a path unless it exists on `prev` or on the defaults (and assert `'url' in formData` is false). If keeping the `''` fallback, rename the test, drop the “ignores / does not invent” comments, and assert the extra key on purpose.

app/src/components/Download.js
```diff
@@ -304,15 +312,30 @@ export function TitleExclusionInput({form, path = 'settings.title_exclude'}) {
 }
 
 export function DownloadFormButtons({onCancel, form}) {
+    /*
+     * Clear is the user's, not the form's.  A successful submit empties only the URLs, so the
```
lrnselfreliance: **[suggestion]** The Clear-button block and the `clearPaths` header in `useForm.js` narrate the feature (submit empties URLs, Clear is for an unrelated next download, download-form tags/destination story). That is WHAT plus design history, not a non-obvious constraint. The useful WHY is already one line: restore named paths so a repeated submit does not wipe settings; missing defaults become `''` so a controlled input is not handed back to the DOM.

**Suggestion:** Cut these down to that constraint. Leave the Videos `emptyFormData` comment — that one is the non-obvious seed-URL reason.

lrnselfreliance (commented):
## Summary

The hook change is sound: `clearPathsOnSuccess` runs after a successful submit, restores only the named paths from `emptyFormData`/`defaultFormData`, and leaves `clearOnSuccess` as a full reset behind it (`fetchOnSuccess` still wins if set). Five of the six create download forms are wired correctly, edit forms still skip the clear (via `clearOnSuccess={false}` or a truthy `onSuccess`), and `VideosDownloadForm` now passes a blank `emptyFormData` so a browser-extension seed URL cannot come back after submit. The scrape create form never receives a truthy clear flag, so it keeps the submitted URLs; comments and the new unit test also claim unknown paths are skipped, but `clearPaths` invents them as `''`.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 0

PR #666, opened 2026-08-16
downloads: a deferred once-download that nothing ever retried

lrnselfreliance:
## The bug

Deferring a download only *schedules* it — `defer()` sets the status, `calculate_next_download` sets `next_download`. The only thing that acted on that schedule was `renew_recurring_downloads`, which filters on `frequency != None`.

A once-download has no frequency. So a single transient error deferred it with a `next_download` that nothing ever read, and it stayed `deferred` until someone retried it by hand. `delete_old_once_downloads` didn't clear it either — that requires a `last_successful_download`, which a download that never succeeded doesn't have.

Net effect: every transient failure added one more permanently-stuck row to the downloads list, so a failure rate that should have self-healed in three hours instead grew without bound.

## How it turned up

Two deployed WROLPis were showing a steadily growing pile of download errors. Every stuck row on both was the same shape:

```
status='deferred'  attempts=1  frequency=NULL  last_successful_download=NULL
next_download='2026-08-12 18:09:33'   <-- four days in the past
```

All of them carried the same transient upstream error (an HTTP 403 on the media stream). Re-running the exact same download by hand succeeded immediately, which is what made the missing retry the interesting half of the problem rather than the error itself.

## The fix

Add `renew_deferred_once_downloads`, called from `do_downloads` alongside the recurring renewal. It uses the same read-then-write split as `renew_recurring_downloads` so the common case (nothing due) never takes the write lock, and re-checks each row's status under the write lock since the dispatcher claims downloads constantly.

`attempts` is deliberately preserved, so `calculate_next_download`'s `3^attempts` backoff keeps growing across retries — 3h, 9h, 27h, and onward. A URL that never succeeds backs off toward never, rather than retrying forever at a fixed interval. That's why there's no explicit attempt cap: the existing backoff already self-limits, and it matches how recurring downloads behave.

## Tests

Written first, confirmed red against `master` (the deferral assertions passed; only the missing renewal failed):

- `test_deferred_once_download_is_retried` — a deferred once-download is left alone before its backoff elapses, renewed once it's due, and completes on the retry.
- `test_deferred_once_download_backoff_grows` — a second failure defers for 3^2 hours, not 3^1 again.
- `test_failed_once_download_is_not_renewed` — an unrecoverably-failed download is never renewed, however overdue.

Full backend suite: 1731 passed, 5 skipped. Frontend: 1221 passed.

### DIFF
--- a/wrolpi/downloader.py
+++ b/wrolpi/downloader.py
@@ -1254,6 +1254,11 @@ class DownloadManager:
             self.log_error(f'Unable to renew downloads!', exc_info=e)
 
         try:
+            self.renew_deferred_once_downloads()
+        except Exception as e:
+            self.log_error(f'Unable to renew deferred once downloads!', exc_info=e)
+
+        try:
             self.delete_old_once_downloads()
         except Exception as e:
             self.log_error(f'Unable to delete old downloads!', exc_info=e)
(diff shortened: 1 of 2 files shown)


wrolpi/downloader.py
```diff
@@ -1382,6 +1387,43 @@ def renew_recurring_downloads(self):
             if renewed_count:
                 self.log_debug(f'Renewed {renewed_count} recurring downloads')
 
+    def renew_deferred_once_downloads(self):
+        """Mark any deferred once-downloads whose backoff has elapsed as "new" so they are retried.
+
+        Deferring only schedules -- it sets `next_download` and nothing more.  `renew_recurring_downloads`
```
lrnselfreliance: **[suggestion]** The new method's docstring narrates the pre-fix bug and embeds design history (`renew_recurring_downloads` only handles `frequency`, a 403 stranded the row, attempts are preserved so backoff grows). `renew_recurring_downloads` itself is one sentence; the extra paragraphs will go stale and are already in the PR description.

**Suggestion:** Keep the first sentence. If the attempts contract needs to live in-code, one short clause is enough (`renew()` is called without `reset_attempts` so backoff keeps growing).

wrolpi/test/test_downloader.py
```diff
@@ -617,6 +617,92 @@ async def test_recurring_downloads(test_session, test_download_manager, fake_now
     assert download.next_download == datetime(2020, 1, 1, 5, 0, 0, 997200, tzinfo=pytz.UTC)
 
 
+@pytest.mark.asyncio
+async def test_deferred_once_download_is_retried(test_session, test_download_manager, fake_now, test_downloader,
+                                                 await_switches):
+    """A once-download that was deferred is retried once its backoff has elapsed.
+
+    A deferred download is only ever scheduled -- `next_download` is set, but nothing marks it "new"
```
lrnselfreliance: **[suggestion]** Same design-history leak as the production docstring: the test restates why the bug existed instead of stating the behavior under test. The first sentence already covers it.

**Suggestion:** Drop the second paragraph. The assertions (`is_new`, `attempts == 1`, then `is_complete`) document the contract.

lrnselfreliance (commented):
## Summary

The fix is correct: `renew_deferred_once_downloads` is the missing scheduler for deferred once-downloads, mirrors the read-then-write split of `renew_recurring_downloads`, and is invoked from `do_downloads` on the same perpetual-worker cadence (every 5s). Filters (`status=deferred`, `frequency IS NULL`, `next_download < now`) exclude pending/new/failed/complete and cannot double-renew recurring rows; the write-side `is_deferred` re-check is sufficient because the dispatcher only claims `new`. Attempts are preserved so `3^attempts` backoff keeps growing, and the three tests actually prove renewal, backoff, and the failed exclusion. Residual risks — a one-shot backlog of overdue deferred once-downloads after upgrade, and never-succeeded rows still escaping `delete_old_once_downloads` — are pre-existing or bounded by the dispatcher, not defects in this change.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

wrolpi/downloader.py
lrnselfreliance: Agreed — applied in 883658a2. Trimmed to the invariant a reader needs; the history lives in the PR description.

wrolpi/test/test_downloader.py
lrnselfreliance: Agreed — applied in 883658a2. Trimmed to the invariant a reader needs; the history lives in the PR description.

PR #667, opened 2026-08-18
downloads: a batch of failed downloads that retried in lockstep

lrnselfreliance:
Follow-up to #666. That PR made deferred downloads retry at all; this one stops the retries from arriving as a synchronized burst.

## The problem

`calculate_next_download` gives every deferred download the same `3^attempts` backoff, measured from the moment it failed. Downloads that fail together therefore retry together — and a synchronized burst against one host is what gets the whole batch rejected in the first place. The batch is then deferred again, still in lockstep. The backoff grows, but the group never breaks up.

On a deployed device this was visible in the data. Fifteen downloads renewed by #666 all failed inside a ten-minute window, and all fifteen came back with `next_download` values spanning those same ten minutes:

```
attempts=2  next_download=2026-08-18 09:44:02
attempts=2  next_download=2026-08-18 09:45:09
...
attempts=2  next_download=2026-08-18 09:54:09
```

With `download_wait` at 10s, that's the whole batch pushed out again in one clump, each one carrying a full metadata + subtitle + comment fetch. The next round trips the same rejection and reschedules the same way.

Worth being clear about what this is not: the upstream errors are transient and time-boxed. The same URLs succeed on retry once the window passes, and a newer downloader binary makes no difference — I tested stable against the current nightly across six of the failing URLs and both succeeded on all six. The bug here is the scheduling shape, not the fetch.

## The fix

Jitter each deferred retry forward by up to half its backoff (`DEFERRED_RETRY_JITTER = 0.5`).

- The backoff stays a **floor** — jitter is added, never subtracted, so nothing retries earlier than it does today.
- A recurring download is still capped at its own frequency, so nothing retries *later* than it does today either. Where the frequency already clamped the backoff, behavior is unchanged.

For the batch above, a 9-hour backoff now spreads across a 4.5-hour window: roughly 17 minutes apart instead of 40 seconds.

## Tests

`test_deferred_downloads_do_not_retry_in_lockstep` is the new one, written first. It defers eight downloads against a frozen clock — so jitter is the only thing that can separate them — and asserts they don't all land on the same moment, while staying inside `[backoff, backoff * 1.5]`. Verified red before the fix (`AssertionError: All deferred downloads were scheduled for the same moment`) by pinning the constant to `0.0`.

`test_deferred_recurring_download_is_capped_at_its_frequency` covers the clamp, so jitter can't push a recurring retry past its frequency.

Three existing tests pinned deferred scheduling to an exact timestamp. Those now assert a floor and a ceiling, which is the real contract. The jitter-sensitive tests were run 8x to check for flakiness — no failures.

Full backend suite: 1733 passed, 5 skipped. Frontend: 1221 passed.

## Not addressed here

The transient rejections themselves. Reducing how often they happen is a separate piece of work — a PO token provider, or feeding the existing encrypted-cookies path, which is currently unused on both deployed devices.

### DIFF
--- a/wrolpi/downloader.py
+++ b/wrolpi/downloader.py
@@ -1656,9 +1656,7 @@ class DownloadManager:
             seconds = hours * Seconds.hour
             if download.frequency:
                 seconds = min(seconds, download.frequency)
-            # Spread the retry across a window so downloads that failed together do not retry together; a
-            # synchronized burst against one host is what gets a whole batch rejected.  Jitter is added, never
-            # subtracted, so the backoff above is still a floor.
+            # Spread retries that failed together.
             seconds += random.uniform(0, seconds * DEFERRED_RETRY_JITTER)
             if download.frequency:
                 # A recurring download is never retried later than its own frequency.
--- a/wrolpi/downloader.py
+++ b/wrolpi/downloader.py
@@ -47,6 +47,10 @@ logger = logger.getChild(__name__)
 
 ARIA2C_PATH = which('aria2c', '/usr/bin/aria2c')
 
+# A deferred Download's retry is pushed back by up to this fraction of its backoff, so a batch that failed
+# together does not retry together.
+DEFERRED_RETRY_JITTER = 0.5
+
 ARIA2C_PROGRESS_RE = re.compile(
     r'\[#\w+\s+'
     r'([\d.]+\w+)/([\d.]+\w+)'  # downloaded/total
@@ -1651,6 +1655,13 @@ class DownloadManager:
             hours = 3 ** (download.attempts or 1)
             seconds = hours * Seconds.hour
             if download.frequency:
+                seconds = min(seconds, download.frequency)
+            # Spread the retry across a window so downloads that failed together do not retry together; a
+            # synchronized burst against one host is what gets a whole batch rejected.  Jitter is added, never
+            # subtracted, so the backoff above is still a floor.
+            seconds += random.uniform(0, seconds * DEFERRED_RETRY_JITTER)
+            if download.frequency:
+                # A recurring download is never retried later than its own frequency.
                 seconds = min(seconds, download.frequency)
             delta = timedelta(seconds=seconds)
             return now() + delta
--- a/wrolpi/test/test_downloader.py
+++ b/wrolpi/test/test_downloader.py
@@ -20,7 +20,7 @@ from wrolpi.conftest import probe_write_lock_is_held, production_like_sessions
 from wrolpi.db import get_db_context
 from wrolpi.downloader import Downloader, Download, DownloadFrequency, import_downloads_config, \
     get_download_manager_config, RSSDownloader, parse_aria2c_progress, _parse_size, \
-    set_download_progress, clear_download_progress, make_progress_callback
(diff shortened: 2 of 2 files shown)


wrolpi/downloader.py
```diff
@@ -1652,6 +1656,13 @@ def calculate_next_download(session: Session, download: Download) -> Optional[da
             seconds = hours * Seconds.hour
             if download.frequency:
                 seconds = min(seconds, download.frequency)
+            # Spread the retry across a window so downloads that failed together do not retry together; a
```
lrnselfreliance: **[suggestion]** The new comment restates `+= random.uniform(0, …)` (“jitter is added, never subtracted”) and embeds the lockstep / host-burst design history. That rationale belongs in the PR description, not beside the arithmetic.

**Suggestion:** Drop it, or keep a single short WHY (e.g. “spread retries that failed together”). The second clamp’s one-liner at 1664 is enough to explain why `min(frequency)` runs twice.

wrolpi/test/test_downloader.py
```diff
@@ -665,18 +670,65 @@ async def test_deferred_once_download_backoff_grows(test_session, test_download_
...
+    # Second failure: 3^2 hours, a larger gap than the first.
+    assert datetime(2020, 1, 1, 14, 0, 0, tzinfo=pytz.UTC) <= download.next_download \
+           <= datetime(2020, 1, 1, 18, 30, 0, tzinfo=pytz.UTC)
+
+
+def test_deferred_downloads_do_not_retry_in_lockstep(test_session, test_download_manager, fake_now,
+                                                    test_downloader):
+    """Downloads that failed together are spread out, rather than all retrying at the same moment.
+
+    Every deferred download gets the same `3^attempts` backoff, so without jitter a batch that failed
```
lrnselfreliance: **[suggestion]** The new test’s docstring repeats the same design story (identical `3^attempts` backoff, synchronized burst, batch rejected and deferred again). The first sentence already states what the test protects.

**Suggestion:** Keep the one-line summary; delete the rest.

lrnselfreliance (commented):
## Summary

This correctly breaks lockstep deferred retries: jitter is applied only on the `is_deferred` path, added after the `3^attempts` backoff (never subtracted), and recurring downloads are re-clamped to `frequency` so a retry cannot land later than it did before. `random.uniform` is the process-global unseeded PRNG, so each download gets an independent draw; the range assertions are inclusive of both endpoints and will not flake, and first-scheduled recurring `next_download` stays on the deterministic zig_zag path. The only residual scheduling gap is intentional: when frequency already equals the backoff (hourly recurring, or weekly after ~5 failures), the second clamp removes all jitter. Dominant issues are comment noise, not correctness.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

wrolpi/downloader.py
lrnselfreliance: Agreed — applied in 02af7beb. Trimmed to a single WHY; the reasoning lives in the PR description.

wrolpi/test/test_downloader.py
lrnselfreliance: Agreed — applied in 02af7beb. Trimmed to a single WHY; the reasoning lives in the PR description.

PR #668, opened 2026-08-19
files: a failed directory rename that hung for five minutes

lrnselfreliance:
Discovered on the docker dev stack while running FileBrowser footer-rename e2e tests, which rename real files through the live API.

## The hang

A directory rename (`POST /api/files/rename`) queues a move and blocks the request on `wait_for_job`. On failure, `handle_move` reverted the files and sent a failure event but never updated the job status, so the job stayed `pending`. `wait_for_job` only understands `complete`, with a 300s timeout.

The HTTP request — and the UI's Rename modal — hung for five minutes. `post_rename` then converted the `TimeoutError` to a misleading `FileConflict`. A "Directory has been renamed" toast could already have fired from the move sub-job.

The move we hit failed with `sqlite3.IntegrityError: UNIQUE constraint failed: file_group.primary_path`: the DB held stale FileGroup rows whose `primary_path` was already the destination (from an earlier successful rename of the same directory; the files had been recreated on disk at the old location without a refresh).

## The wedge

After that sequence the perpetual file-worker loop stopped processing jobs. Queued refreshes sat untouched for 15+ minutes while `/api/files/worker_status` reported `idle`. No "Perpetual worker … had error" lines. Restarting the API was the only recovery.

Two mechanisms, both real:

1. **`CancelledError` killed the loop silently.** `perpetual_signal` caught it, skipped re-dispatch, and logged nothing. Any cancel escaping `process_queue` (there are explicit `asyncio.sleep(0)` yield points throughout the worker) ended file processing for the life of the process.

2. **The shared `perpetual_tasks_started` Event outlives the worker that set it.** Docker runs Sanic with `auto_reload` and many workers. Worker A claims the Event and is the only process running the pumps. Worker A exits — recycle, crash, or its signal task dying — without a full server shutdown. The Event stays set. Every later `after_server_start` is a no-op. That matches the incident: last file-worker activity from one PID, then silence from every PID.

## The fix

- Failed move/reorganize jobs are marked `failed` with the recorded error. `wait_for_job` raises `FileWorkerJobFailed` immediately instead of spinning for 300s.
- Unexpected `CancelledError` is logged and the perpetual worker is rescheduled, unless `before_server_stop` has fired. Sanic's `is_running` is **not** used as a shutdown signal — it stays false in these docker workers and treating it as shutdown killed every loop after one tick.
- Track `perpetual_tasks_owner_pid`. If that PID is gone, a replacement worker reclaims and restarts the loops. Every Sanic worker still starts the file-worker pump even when it does not own the other loops.
- `transfer_queue` drains with `get_nowait()` instead of check-then-act on `qsize()`, so a cross-process `queue.Empty` no longer escapes.
- Production `wait_for_job` only polls. Request processes were stealing from the shared public queue (LIFO), which is how a move jumped ahead of a queued refresh.
- A stall log fires if jobs sit queued while status is idle and a tick processed nothing.

Stale destination FileGroup rows are still an abnormal state. The non-negotiable part is that a failed move fails the request promptly; stealing those rows is not in this PR.

## Tests

Written first for the wedge (`test_claim_perpetual_tasks_when_owner_process_is_dead` and friends failed with `ImportError` before the claim API existed).

- `test_wait_for_job_raises_immediately_when_job_failed`
- `test_handle_move_marks_job_failed_on_exception`
- `test_wait_for_job_fails_promptly_on_stale_destination_filegroup`
- `test_rename_directory_stale_destination_fails_promptly`
- `test_handle_reorganize_marks_job_failed_on_exception`
- `test_perpetual_iteration_reschedules_after_unexpected_cancel`
- `test_claim_perpetual_tasks_when_owner_process_is_dead`

Verified end-to-end against the live docker stack with the footer-rename Cypress spec (not in this PR): 4/4, several runs in a row, ~15–20s each.

## Not addressed here

`single_tasks_started` uses the same Event-only pattern (config import, startup downloads). Same class of outage if that owner worker dies.

### DIFF
--- a/main.py
+++ b/main.py
@@ -318,7 +318,8 @@ async def start_single_tasks(app: Sanic):
 @perpetual_signal(sleep=1)
 async def perpetual_file_worker_queue():
     file_worker.transfer_queue()
-    await file_worker.process_queue()
+    processed = await file_worker.process_queue()
+    file_worker.check_queue_stall(processed=processed)
 
 
 @perpetual_signal(sleep=1)
--- a/wrolpi/api_utils.py
+++ b/wrolpi/api_utils.py
@@ -165,12 +165,32 @@ async def cleanup_session(request: Request, response_: HTTPResponse):
 
 PERPETUAL_WORKERS = list()
 
-# Always started in every Sanic worker so file jobs survive the owner process dying.
 FILE_WORKER_PERPETUAL_EVENT = 'wrolpi.perpetual.perpetual_file_worker_queue'
+OWNER_WATCHDOG_EVENT = 'wrolpi.perpetual.perpetual_owner_watchdog'
+OWNER_HEARTBEAT_STALE_SECONDS = 30
+
+
+def _touch_heartbeat(app) -> None:
+    try:
+        app.shared_ctx.perpetual_tasks_heartbeat.value = time()
+    except Exception:
+        logger.error('Failed to update perpetual-tasks heartbeat', exc_info=True)
+
+
+def _i_am_owner(app) -> bool:
+    try:
+        return app.shared_ctx.perpetual_tasks_owner_pid.value == os.getpid()
+    except Exception:
+        return False
 
 
 def _perpetual_owner_is_alive(app) -> bool:
-    """True if the process that claimed the single-process perpetual loops is still running."""
+    """True if the claiming worker is still running *and* keeping a heartbeat.
+
+    pid_is_running alone is not enough: after a crash the kernel can reuse the
+    owner PID for an unrelated process.  A zero heartbeat means the owner just
+    claimed and has not ticked yet, so a live PID still counts.
+    """
     try:
         pid = app.shared_ctx.perpetual_tasks_owner_pid.value
     except Exception:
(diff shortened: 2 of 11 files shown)


lrnselfreliance:
Agreed on the every-worker pump. That was the wrong trade.

Pushed `3115ecd6` so only the claiming worker runs `perpetual_file_worker_queue`. Every other worker runs `perpetual_owner_watchdog`: if we own the loops, refresh the heartbeat; if the owner PID is gone or the heartbeat is stale (30s), claim and start the pumps. File jobs stay single-consumer. After restart, pid 171 started all 11 events; the rest started only the watchdog.

Tests were written first (`test_non_owner_starts_watchdog_not_file_worker`, `test_claim_perpetual_tasks_when_heartbeat_is_stale`, `test_watchdog_reclaims_and_starts_pumps_when_owner_is_dead`) and failed before the change.

Also in this commit:
- Production `wait_for_job` polls at 0.1s, not 100 Hz.
- `post_rename` puts the recorded job error (first line) on the 409.
- Failed owner-PID writes are logged, not swallowed.
- Heartbeat covers PID reuse.

Left as follow-ups, as you said:
- `_jobs` never evicts.
- `single_tasks_started` has the same Event-only outage.
- `check_queue_stall` still cannot report the pump's own death.

`cancel_background_tasks` / `cancel_refresh_tasks` do not cancel perpetual signals. The only intentional stop is shutdown (`before_server_stop`).

PR #669, opened 2026-08-19
upgrade: restart services when the upgrade fails before code changes

lrnselfreliance:
## Problem

`upgrade.sh` stops `wrolpi-api` and `wrolpi-app` before doing any git work. If the upgrade then fails — a failed `git fetch`, a GPG signature verification failure (exit 5/6), or a failed checkout — the script exited with the services still stopped, leaving the WROLPi in a failed state even though the code on disk was unchanged and perfectly runnable.

## Fix

- After stopping the services, set `RESTART_SERVICES_ON_FAILURE=true`. The EXIT trap now starts `wrolpi-api` and `wrolpi-app` back up on any non-zero exit while that flag is set, so every pre-upgrade failure (fetch, signature verification, checkout) recovers automatically.
- The post-checkout verification failure (exit 6) happens after the new commit is checked out, so it now rolls the branch back to the previously-running commit before exiting. The trap therefore restarts services on known-good code, never on an unverified commit — the signature guard still fully holds.
- The flag is cleared right before handing off to `scripts/upgrade.sh`, which starts modifying venvs/packages/configs. A blind restart there could run half-upgraded code, so recovery for late failures remains `repair.sh`'s job, as before.

## Testing

- `bash -n` syntax check.
- Simulated the trap + flag flow with stubbed `systemctl`: signature failure restarts both services and preserves the exit code; late failure and success paths do not trigger the restart.
- Frontend Jest suite: 70 suites / 1221 tests passed (no JS changes; baseline confirmation).

### DIFF
--- a/upgrade.sh
+++ b/upgrade.sh
@@ -89,10 +89,6 @@ trap '
     echo "WROLPi upgrade has completed"
   else
     echo "WROLPi upgrade has FAILED (exit code: $EXIT_STATUS)"
-    # If we failed before the real upgrade work started (fetch, signature
-    # verification, or checkout), no code was changed on disk (the exit-6
-    # path rolls the checkout back), so restart the services we stopped
-    # instead of leaving the WROLPi down.
     if [ "$RESTART_SERVICES_ON_FAILURE" = true ]; then
       echo "Restarting WROLPi services after failed upgrade..."
       systemctl start wrolpi-api || :
@@ -110,9 +106,18 @@ set -o pipefail
 # Units may not be installed yet if a previous install/repair failed; repair.sh installs them later.
 systemctl stop wrolpi-api || :
 systemctl stop wrolpi-app || :
-# Services are now stopped; until the upgrade actually modifies the code the
-# EXIT trap can safely bring them back up on failure.
 RESTART_SERVICES_ON_FAILURE=true
+
+# Restore the previously-running commit after a failed checkout or verification.
+# Services must only restart on that known-good commit, so if the rollback does
+# not land there, leave them stopped.
+rollback_or_no_restart() {
+  git -C /opt/wrolpi checkout -f -B "${BRANCH}" "${PREVIOUS_HEAD}" || :
+  if [ "$(git -C /opt/wrolpi rev-parse HEAD 2>/dev/null)" != "${PREVIOUS_HEAD}" ]; then
+    echo "ERROR: rollback to ${PREVIOUS_HEAD} failed; not restarting services."
+    RESTART_SERVICES_ON_FAILURE=false
+  fi
+}
 
 # Fetch the latest commits without modifying the working tree.
 git -C /opt/wrolpi fetch || exit 4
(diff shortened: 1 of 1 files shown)


upgrade.sh
```diff
@@ -134,13 +148,21 @@ fi
 # Verify HEAD again after checkout in case the branch was swapped between fetch and checkout.
 if ! git -C /opt/wrolpi verify-commit HEAD 2>&1; then
   echo "ERROR: HEAD commit signature verification failed after checkout. Aborting upgrade."
+  # Roll back to the previously-running (verified) commit so the EXIT trap
+  # restarts the services on known-good code, never on the unverified commit.
+  git -C /opt/wrolpi checkout -f -B "${BRANCH}" "${PREVIOUS_HEAD}" || :
```
lrnselfreliance: **[bug]** On post-checkout `verify-commit` failure, rollback is `git checkout -f -B "${BRANCH}" "${PREVIOUS_HEAD}" || :`. If that checkout fails (disk full, permissions, unexpected git state), the error is swallowed, `exit 6` still runs with `RESTART_SERVICES_ON_FAILURE=true`, and the EXIT trap starts api/app against the unverified commit still checked out. That breaks the invariant stated in the comment immediately above: the trap is supposed to restart on known-good code, never on the unverified commit. `|| :` is needed so `set -e` still reaches `exit 6`, but it must not keep the restart flag set after a failed rollback.

**Suggestion:** If rollback fails, clear the flag before exiting so services stay down rather than running unverified code, e.g. `if ! git -C /opt/wrolpi checkout -f -B "${BRANCH}" "${PREVIOUS_HEAD}"; then echo "ERROR: rollback to ${PREVIOUS_HEAD} failed; not restarting services."; RESTART_SERVICES_ON_FAILURE=false; fi`. Optionally confirm `$(git -C /opt/wrolpi rev-parse HEAD) = "$PREVIOUS_HEAD"` before leaving the flag true.

upgrade.sh
```diff
@@ -125,6 +137,8 @@ if ! git -C /opt/wrolpi verify-commit origin/"${BRANCH}" 2>&1; then
 fi
 
 # Signature verified, safe to apply.
+# Remember the running code's commit so a failed verification below can roll back to it.
+PREVIOUS_HEAD=$(git -C /opt/wrolpi rev-parse HEAD)
```
lrnselfreliance, line 145: **[suggestion]** `PREVIOUS_HEAD` is captured only to roll back the exit-6 path. The following `git checkout -f -B "${BRANCH}" "origin/${BRANCH}" || exit 4` can fail after git has already moved the branch ref (`checkout -B` updates the ref, then checks out the tree). This PR now restarts api/app on that `exit 4`, so a partial checkout that used to leave the box down can now start services on a mixed/new tree that was never post-checkout verified. Fetch/exit-5 paths are still safe (worktree unchanged); this is the new restart combined with a non-atomic checkout.

**Suggestion:** On checkout failure, attempt the same rollback to `PREVIOUS_HEAD` (or skip restart unless `HEAD` still equals `PREVIOUS_HEAD`) before `exit 4`, so the trap only brings services up on the previously-running commit.

upgrade.sh
```diff
@@ -89,6 +89,15 @@ trap '
     echo "WROLPi upgrade has completed"
   else
     echo "WROLPi upgrade has FAILED (exit code: $EXIT_STATUS)"
+    # If we failed before the real upgrade work started (fetch, signature
```
lrnselfreliance: **[suggestion]** The four-line comment restates what the following `if` does and dumps the fetch/verify/checkout/exit-6 design into the trap body. Same class of narrative comments sit next to the flag set (lines 113–114) and the rollback (lines 151–152). They explain WHAT and retell the change rather than a short WHY.

**Suggestion:** Drop the trap-body block (the flag name plus `systemctl start` are enough). Keep one short WHY at the rollback and at the flag-clear (`RESTART_SERVICES_ON_FAILURE=false` is the non-obvious constraint).

lrnselfreliance (commented):
## Summary

The change correctly fixes a real availability hole: `upgrade.sh` used to stop `wrolpi-api`/`wrolpi-app` and then leave them down on a failed fetch, signature check, or checkout even though the tree was still runnable. The single-quoted EXIT trap plus `RESTART_SERVICES_ON_FAILURE` is the right mechanism (`$RESTART_SERVICES_ON_FAILURE` is expanded at fire time, not definition time), EXIT runs on the claimed `exit 4/5/6` and `set -e` paths, success does not restart, and the flag is cleared after git work and before `scripts/upgrade.sh` mutates venvs/packages. The main gap is the exit-6 rollback: `|| :` still lets the trap start services if that checkout fails, which is exactly the unverified-commit case the new code is trying to avoid. There is no separate docker/rpi/`wrolpi-help` variant of this script; help is never stopped here, and late failures still belong to `repair.sh`.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 0

upgrade.sh
lrnselfreliance: Fixed in 751d9bfe: rollback now goes through `rollback_or_no_restart()`, which verifies `rev-parse HEAD` equals `PREVIOUS_HEAD` after the checkout and clears `RESTART_SERVICES_ON_FAILURE` if it does not. A failed rollback now leaves services stopped instead of restarting on the unverified commit.

upgrade.sh
lrnselfreliance: Fixed in 751d9bfe: the checkout failure path now also calls `rollback_or_no_restart()` before `exit 4`, so a partial `checkout -B` either rolls back to the previously-running commit or leaves services stopped.

upgrade.sh
lrnselfreliance: Trimmed in 751d9bfe: dropped the trap-body block and the flag-set narration; kept short WHY comments at the rollback helper, the non-atomic checkout, and the flag clear.

PR #670, opened 2026-08-20
e2e: exercise the FileBrowser footer buttons against real files

lrnselfreliance:
## What

Sixteen Cypress e2e tests for the FileBrowser's sticky-footer buttons, run against the live docker stack with **no API mocking** — real files are created, renamed, moved, deleted, tagged, ignored, and uploaded, and every assertion checks the disk as well as the table.

| Spec | Covers |
|---|---|
| `footer-rename.cy.js` | selection rules; modal prefill/Reset/disabled-until-changed; file rename; directory rename with children |
| `footer-move.cy.js` | single move via DirectorySearch with destination preview; multi-select move; Reset/Escape moves nothing |
| `footer-delete.cy.js` | confirm-cancel; file delete; recursive open-directory delete; tagged file → force-delete flow |
| `footer-ignore-upload-refresh.cy.js` | Upload/Ignore single-directory enable rules; ignore/un-ignore config round-trip; real upload; scoped Refresh verified via its completion event |

All specs are skipped in CI, which runs only the React dev server.

## Harness

* `fb:*` tasks in `cypress.config.js` seed a `cypress-fb/` sandbox inside the media directory; every task is jailed to that sandbox (plus a `CypressFB*`-only tag-directory cleaner), so a buggy test cannot touch shared fixtures.
* `cy.serverNow` / `cy.waitForEvent` / `cy.refreshFiles` in `commands.js`: background jobs (move, refresh) only queue work and `worker_status` reads `idle` while jobs wait in the queue, so completion is detected by polling `/api/events/feed` for events newer than the server's clock — with a hard timeout that fails the test loudly.
* `support/file-browser.js` holds shared setup/teardown and a WROL-mode guard. Teardown is unconditional: sandbox files, DB rows via a final refresh, test tags and their `tags/` directories, and any leaked ignored-directory config.

## Invariants the specs encode

* The DB must agree with the seeded disk before a directory rename, or the move job dies on a `primary_path` UNIQUE constraint — every setup refreshes the sandbox and waits for the event.
* Event toasts stack over the sticky footer and block clicks; the specs hide the notification layer since they assert on disk and table, not toasts.
* Open folders live in the `?folders=` query and survive reloads.

## Testing

* All 4 specs: two consecutive full runs green (16/16), zero residue afterward (no sandbox, tags, or config changes).
* `pytest -nauto`: green apart from one pre-existing unrelated flake.
* Jest: 70 suites / 1221 tests passing.

### DIFF
--- a/app/cypress/e2e/files/footer-ignore-upload-refresh.cy.js
+++ b/app/cypress/e2e/files/footer-ignore-upload-refresh.cy.js
@@ -1,0 +1,143 @@
+/*
+ * The Ignore, Upload, and Refresh buttons in the FileBrowser's sticky footer, against the
+ * live stack.  Ignore/Upload only make sense with a single directory selected, so this spec
+ * also carries their enable-state rules.
+ */
+import {
+    assertWROLModeOff,
+    deselectRow,
+    dialog,
+    footerButton,
+    row,
+    selectRow,
+    setupFileBrowser,
+    teardownFileBrowser,
+} from '../../support/file-browser';
+
+// Requires the full stack (API + real media directory); CI runs only the React dev server.
+(Cypress.env('CI') ? describe.skip : describe)('FileBrowser footer: Ignore, Upload, Refresh', () => {
+
+    const seedFiles = {
+        'fb-notes.txt': 'a file',
+        'fb-beta/fb-inside.txt': 'inside the directory',
+    };
+
+    const IGNORED_DIR = 'cypress-fb/fb-beta';
+
+    before(assertWROLModeOff);
+
+    beforeEach(() => setupFileBrowser(seedFiles, 'fb-notes.txt'));
+
+    after(() => {
+        // Un-ignore unconditionally: a failed test must not leave the sandbox in the
+        // ignored_directories config (it would confuse later runs and the user's stack).
+        cy.request({
+            method: 'POST',
+            url: '/api/files/unignore_directory',
+            body: {path: `/media/wrolpi/${IGNORED_DIR}`},
+            failOnStatusCode: false,
+        });
+        teardownFileBrowser();
+    });
+
+    // cy.task has no built-in retry; background work (uploads) lands on disk asynchronously.
+    const waitForDisk = (relativePath, timeout = 30000) => {
+        const deadline = Date.now() + timeout;
+        const poll = () => {
+            cy.task('fb:exists', relativePath).then((exists) => {
+                if (exists) return;
+                expect(Date.now(), `${relativePath} did not appear on disk within ${timeout}ms`)
+                    .to.be.lessThan(deadline);
+                cy.wait(500);
+                poll();
+            });
+        };
+        poll();
+    };
+
+    it('Upload and Ignore require a single directory selection', () => {
+        // Nothing selected: Upload targets the media directory, Ignore has no directory.
+        footerButton('Upload').should('be.enabled');
+        footerButton('Ignore').should('be.disabled');
+
+        selectRow('fb-notes.txt');
+        footerButton('Upload').should('be.disabled');
+        footerButton('Ignore').should('be.disabled');
+        deselectRow('fb-notes.txt');
+
+        selectRow('fb-beta');
+        footerButton('Upload').should('be.enabled');
+        footerButton('Ignore').should('be.enabled');
+
+        selectRow('fb-notes.txt');
+        footerButton('Upload').should('be.disabled');
+        footerButton('Ignore').should('be.disabled');
+    });
+
+    it('ignores and un-ignores a directory through the config', () => {
+        selectRow('fb-beta');
+        footerButton('Ignore').click();
+
+        dialog().should('contain.text', 'Ignore Directory')
+            .should('contain.text', `/media/wrolpi/${IGNORED_DIR}`);
+        dialog().contains('button', 'Ignore').click();
+
+        // The directory lands in the settings config.
+        cy.request('/api/settings').its('body.ignored_directories')
+            .should('include', IGNORED_DIR);
+
+        // With the settings reloaded the same button now offers the reverse action.
+        footerButton('Unignore').should('be.enabled').click();
+        dialog().should('contain.text', 'Remove Ignore Directory');
+        dialog().contains('button', 'Un-ignore').click();
+
+        cy.request('/api/settings').its('body.ignored_directories')
+            .should('not.include', IGNORED_DIR);
+        footerButton('Ignore').should('be.enabled');
+    });
+
+    it('uploads a file into the selected directory', () => {
+        selectRow('fb-beta');
+        footerButton('Upload').click();
+
+        dialog().should('contain.text', `Upload to: /media/wrolpi/${IGNORED_DIR}`);
+        dialog().find('input[type="file"]').selectFile({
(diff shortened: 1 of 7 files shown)


app/cypress/support/commands.js
```diff
@@ -2,3 +2,80 @@ import {mount} from 'cypress/react'
...
+        return true;
+    };
+
+    const deadline = Date.now() + timeout;
+    const poll = () => {
+        cy.request(`/api/events/feed?after=${encodeURIComponent(since)}`).then(({body}) => {
+            const events = body.events || [];
+            const failed = events.find((e) => failures.includes(e.event));
+            expect(failed, failed && `${failed.event}: ${failed.message}`).to.be.undefined;
+            if (events.some(matches)) return;
```
lrnselfreliance, line 54: **[bug]** `waitForEvent` succeeds on the first matching event with `dt > since`. There is no job id, and when `message` is omitted any event of that name counts. `refreshFiles` then waits for exact `Refreshed: cypress-fb`, which cannot tell two refreshes of the same directory apart. That matters because `POST /api/files/rename` queues a parent-directory refresh and returns without waiting (`wrolpi/files/api.py` after the rename). In `footer-rename.cy.js` the file-rename test does that, then the next test's `beforeEach` runs `fb:reset` (new seed on disk) and `refreshFiles([SANDBOX])`. If the previous refresh completes after `serverNow`, `waitForEvent` returns while the newly POSTed refresh is still queued. The DB still reflects the pre-reset tree (`fb-renamed.txt`, leftover `fb-alpha-renamed/…`) while disk has the seed — the stated directory-rename `primary_path` UNIQUE failure. `moveAndWait` has the same unscoped `file_move_completed` match. Clock/timezone are fine (`body.now` and event `dt` are UTC ISO); the poll also correctly sees events that finish before the first poll. The bug is correlating the wrong completion.

**Suggestion:** Do not treat an in-flight previous job as success. Snapshot matching events (or their `dt`s) before the POST and wait for one that was not in the snapshot; or wait for the rename's follow-up `directory_refresh` in that test before yielding; or put a job id in the event. `moveAndWait` should at least match the destination in the `Moved N items to …` message.

app/cypress/e2e/files/footer-ignore-upload-refresh.cy.js
```diff
@@ -0,0 +1,143 @@
...
+        'fb-beta/fb-inside.txt': 'inside the directory',
+    };
+
+    const IGNORED_DIR = 'cypress-fb/fb-beta';
+
+    before(assertWROLModeOff);
+
+    beforeEach(() => setupFileBrowser(seedFiles, 'fb-notes.txt'));
+
+    after(() => {
```
lrnselfreliance: **[suggestion]** Teardown is not unconditional. Cleanup lives in `after()` (once per spec), not `afterEach()`. If `ignores and un-ignores a directory` fails after `POST /api/files/ignore_directory`, `cypress-fb/fb-beta` stays in `ignored_directories` for the later upload/refresh tests in the same spec; other specs never unignore. Cypress also aborts the rest of a hook when a command fails: `footer-delete.cy.js`'s `after()` runs `deleteTestTag()` then `teardownFileBrowser()` — a failed `GET /api/tag` skips `fb:reset` / `fb:cleanup`. `beforeEach` `fb:reset` covers leftover files for the next test in the same spec, but not leftover config/tags, and not a crashed run that never reaches `after()`.

**Suggestion:** Move sandbox + ignore + tag cleanup to `afterEach`, use `failOnStatusCode: false` on the API teardowns (as the unignore already does), and always `cy.task('fb:cleanup')` afterward so a failed request cannot skip the disk wipe.

app/cypress/support/commands.js
```diff
@@ -2,3 +2,80 @@ import {mount} from 'cypress/react'
...
+/*
+ * Waiting on WROLPi background work.
+ *
+ * Most mutating file endpoints only QUEUE a job for the background file worker (move, refresh),
+ * and /api/files/worker_status reads 'idle' while jobs are still waiting in its queue, so
+ * neither the response nor the status endpoint says anything about completion.  The reliable
+ * signal is the completion event on /api/events/feed.  The pattern: capture the server's clock
+ * BEFORE starting the work, then poll the feed for a matching event newer than that timestamp.
+ *
+ * If the timeout expires the test FAILS -- historically that has meant the file worker was
```
lrnselfreliance: **[suggestion]** The `waitForEvent` prelude embeds architecture history ("historically that has meant the file worker was wedged and the api service needed a restart (fixed since, but fail loudly regardless)") rather than a short WHY. Same pattern in `footer-ignore-upload-refresh.cy.js` around the refresh reload (`useBrowseFiles only fetches on mount… if this later starts auto-updating, the reload can be dropped`).

**Suggestion:** Keep a one-liner on why the feed is the completion signal (worker_status is idle while queued) and why the refresh test reloads (browse does not auto-refetch). Drop the history and the forward-looking product note.

lrnselfreliance (commented):
## Summary

These Cypress specs are a real live-stack FileBrowser exercise: `fb:*` tasks seed/assert a jailed `cypress-fb/` sandbox, UI actions hit the real API, and most assertions check disk as well as the table. The path jail and `CypressFB[A-Za-z0-9]*` tag-dir cleaner are tight; `Cypress.env('CI')` skip matches the existing `caddy-media.cy.js` pattern and will not run on CircleCI (`CI=true npx cypress run`). The main risk is `waitForEvent` treating *any* matching name after `since` as *this* job — that can desync DB from a `fb:reset` seed and trip the directory-rename UNIQUE constraint the suite is trying to avoid. Teardown is also weaker than advertised (`after()` plus a failing hook command).

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 0

app/cypress/support/commands.js
lrnselfreliance: Fixed in 0f95bd7b, with one honest caveat. The rename tests now drain the queued parent-directory refresh before ending (drainQueuedRefresh waits for 'Refreshed: cypress-fb' after each rename), so the known straggler is gone at its source, and moveAndWait now matches the destination in the 'Moved N items to …' message. On the snapshot idea: since 'since' is captured immediately before the POST, a snapshot of prior events is equivalent to the dt filter — it cannot distinguish an in-flight job that completes after the POST. The only airtight correlation is a job id in the event itself, which is a product change; happy to propose that as a follow-up if you want it.

app/cypress/e2e/files/footer-ignore-upload-refresh.cy.js
lrnselfreliance: Fixed in 0f95bd7b. The unignore now runs in afterEach (failOnStatusCode: false), deleteTestTag no longer throws on API errors, and teardownFileBrowser's DB-purge refresh is best-effort with no assertions — so the hooks always reach fb:cleanup and the disk wipe cannot be skipped by a failed request.

app/cypress/support/commands.js
lrnselfreliance: Fixed in 0f95bd7b — the prelude now states why the feed is the signal (no job id; a test must drain jobs it queues) without the wedge history, and the refresh test's comment is down to the two facts it needs (browse does not refetch; ?folders= keeps the sandbox open).

PR #672, opened 2026-08-20
app: the navbar corner wrapped onto a second row on small phones

lrnselfreliance:
## Summary

On screens narrower than about 380px (iPhone SE class), the mobile navigation bar broke onto two rows: the logo on one line and the whole icon corner — share, hotspot, theme, search, hamburger — on a second.

Each status-icon slot (`.wrolpi-navbar-icon`) carried a `margin-left: 1.5em` separator, ~80px across three slots — more than the deficit. That margin existed to separate the indicators from the desktop bar's links, but the corner's `gap` and the slots' own padding already space the icons.

## Changes

- `.wrolpi-navbar-icon` margin tightened from `1.5em` to `0.25em` on **both** bars — one system, no mobile-only override. On desktop the freed room goes back to the links that `useOverflowNav` folds into "More".
- The mobile bar is extracted as an exported `MobileNav` component (mirroring `DesktopNav`) so tests can mount the real markup.
- New Cypress suite mounts `MobileNav` at 414/375/360px and asserts the bar stays one row and the corner is not clipped. It asserts down to 360 rather than 320 because the fixture is wider than the real bar (Cypress detects a keyboard so the ⌘K hint renders, and the text home link outmeasures the logo); the real app at an emulated 320px renders a single-row bar with room to spare.

## Testing

- Jest: 1221 passed (includes the navbar-style and icon-name audits).
- Cypress ui-layout component spec: 313 passed, including the existing 640–1180px desktop no-wrap sweep.
- `npm run build` compiles cleanly.
- Verified in Chrome against the dev app: one row at 1400px desktop and at emulated 375px and 320px phone widths.

### DIFF
--- a/app/src/App.css
+++ b/app/src/App.css
@@ -143,7 +143,13 @@ html .wrolpi-navbar .mantine-ActionIcon-root:hover {
 .wrolpi-navbar-icon {
     display: flex;
     align-items: center;
-    margin-left: 1.5em;
+    /*
+     * Tightened from 1.5em: three slots' worth of that was ~80px, which wrapped the mobile
+     * bar's corner onto a second row on a 375px phone, and on the desktop bar it only took
+     * room from links that useOverflowNav then folded into "More".  The corner's gap and
+     * the slots' own padding still space the icons.
+     */
+    margin-left: 0.25em;
     /* Full bar height, so the anchor inside has somewhere to stretch to. */
     align-self: stretch;
 }
--- a/app/src/App.css
+++ b/app/src/App.css
@@ -143,12 +143,8 @@ html .wrolpi-navbar .mantine-ActionIcon-root:hover {
 .wrolpi-navbar-icon {
     display: flex;
     align-items: center;
-    /*
-     * Tightened from 1.5em: three slots' worth of that was ~80px, which wrapped the mobile
-     * bar's corner onto a second row on a 375px phone, and on the desktop bar it only took
-     * room from links that useOverflowNav then folded into "More".  The corner's gap and
-     * the slots' own padding still space the icons.
-     */
+    /* A sliver only: the corner's gap and the slots' own padding already separate the
+       icons, and anything more has to fit a 320px phone times the number of slots. */
     margin-left: 0.25em;
     /* Full bar height, so the anchor inside has somewhere to stretch to. */
     align-self: stretch;
--- a/app/src/components/ui/ui-layout.cy.js
+++ b/app/src/components/ui/ui-layout.cy.js
@@ -3742,6 +3742,14 @@ describe('the mobile navigation bar fits a phone on one row', () => {
      * slots: more than the deficit.  The slot margin is small now (`.wrolpi-navbar-icon`),
      * and this holds the bar to fitting a phone whatever the margin becomes.
      */
+    /*
+     * The fixture matches what a phone actually renders, so the widths below are the widths
+     * being proven -- not a stand-in with different contents:
+     *   - the home link is the 32px logo, the default (`NAME` is null in production);
+     *   - the theme slot carries the 0.8em right margin the real DarkModeToggle has;
+     *   - the pointer is made coarse (below), which is what makes a phone drop the search
+     *     link's keyboard hint.
+     */
     const indicators = <>
         <NavIconWrapper>
             {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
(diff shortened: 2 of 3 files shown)


app/src/components/ui/ui-layout.cy.js
```diff
@@ -3734,6 +3734,82 @@ describe('the navigation bar never breaks onto a second line', () => {
...
+                .to.be.greaterThan(30);
+        });
+    });
+
+    it('stays one row at small phone widths', () => {
+        /*
+         * 375px is the iPhone SE that wrapped.  The fixture is wider than the real bar --
+         * the search link renders its keyboard hint here (a phone detects no keyboard and
+         * drops it) and the text home link outmeasures the logo -- so passing at 360px
+         * covers real devices meaningfully narrower than that.
```
lrnselfreliance: **[suggestion]** The suite does mount real `MobileNav` (real `SearchIconButton` + hamburger, indicators in `NavIconWrapper`). It never measures 320px, and the comment claims that passing at 360px “covers real devices meaningfully narrower than that” because the fixture is wider than production (⌘K hint on a non-touch Cypress runner; italic `WROLPi` text vs the 32px logo `NAME` is null). That extra is unmeasured, and it is also incomplete: production `DarkModeToggle` carries `style={{marginRight: '0.8em'}}` (~14px at `--ui-scale: 1.1`) which this fixture’s theme `IconButton` does not, so the “fixture is wider” safety margin is smaller than the comment states. Height vs a 700px `singleRow` and `corner.right <= bar.right` are the right checks (border box of `.wrolpi-navbar` itself, not a padded wrapper), matching the desktop suite; they still only prove this markup at 360+.

**Suggestion:** Drop the claim that 360px stands in for narrower devices. Either (a) mount a 32px logo `homeLink` and accept 360 as the proven floor, or (b) include the theme button’s `0.8em` right margin (or the real `DarkModeToggle`) and, if the hint is the blocker, stub `useKeyboardDetected` so 320px is actually measured.

app/src/App.css
```diff
@@ -143,7 +143,13 @@ html .wrolpi-navbar .mantine-ActionIcon-root:hover {
 .wrolpi-navbar-icon {
     display: flex;
     align-items: center;
-    margin-left: 1.5em;
+    /*
+     * Tightened from 1.5em: three slots' worth of that was ~80px, which wrapped the mobile
```
lrnselfreliance: **[suggestion]** The new comment narrates the change (`Tightened from 1.5em`), the 375px wrap, and `useOverflowNav` folding links into “More”. The durable constraint is only that the corner’s `gap` and the slots’ own padding already separate the icons.

**Suggestion:** Replace with one short WHY line, e.g. that `1.5em` per slot was separator the corner `gap` and padding already provide.

lrnselfreliance (commented):
## Summary

The CSS change is the right fix for the reported wrap: three `.wrolpi-navbar-icon` slots at `1.5em` were ~80px of separator, which is the whole deficit below ~380px. Tightening to `0.25em` saves ~66px and still leaves the corner `gap` plus slot/link padding to separate icons; on desktop `margin-left: auto` on `.wrolpi-navbar-right` means the old margin was never the visible gap until the bar was packed, so the freed room correctly goes back to `useOverflowNav`. `MobileNav` is a behavior-preserving extract (same `SearchIconButton`, hamburger, WROL-mode `homeLink`, hotspot/share wrappers). The new Cypress suite mounts real `MobileNav`, not a mock. Dominant risk is the suite asserting 360px on a fixture that is not the production width, then describing that as coverage of narrower phones.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

app/src/components/ui/ui-layout.cy.js
lrnselfreliance: Fixed in f67790df — took option (b), extended: the fixture now mounts the 32px logo home link (the production default), adds the theme toggle's 0.8em right margin, and stubs '(pointer: coarse)' so useKeyboardDetected stays false and the hint drops exactly as on a phone. A guard test asserts the search link renders no hint text, and 320px is measured directly (414/375/320, plus the corner-clipping check at 375/320). The unproven 'covers narrower devices' claim is gone.

app/src/App.css
lrnselfreliance: Fixed in f67790df — the narration is replaced with the constraint: the corner's gap and the slots' own padding already separate the icons, and anything more must fit a 320px phone times the number of slots.

PR #673, opened 2026-08-21
Fix two concurrency races: torn config reads and instant SQLite lock failures on playlist writes

lrnselfreliance:
## Summary

Two race conditions observed in production logs on a multi-worker Docker deployment:

### 1. `InvalidConfig: config/download_manager.yaml` during a background save

`write_config_data()` truncated the config file in place before dumping YAML into it, so any concurrent reader (another Sanic worker's version check, config import) could observe an empty or partial file. The version check in `ConfigFile.save()` also read the file *before* acquiring `config_save_lock`, so it raced the worker currently writing.

**Fix:** config writes now go to a temporary file in the same directory (fsync'd) and are renamed into place with `os.replace` — a reader can only ever see the complete old or complete new contents. The temporary file inherits the config's permissions and is removed on failure. The version check moved inside `config_save_lock`.

### 2. `sqlite3.OperationalError: database is locked` on `INSERT INTO collection_item`

The playlist item endpoints (add/remove/reorder) used the deferred `request.ctx.session`. They read (the Collection, the FileGroup) before their first write, and SQLite fails a deferred transaction's lock upgrade *instantly* — bypassing `busy_timeout` — whenever another connection wrote in between. Under background write load (tag config saves and directory sync), an add-to-playlist request failed with a 500.

**Fix:** these endpoints now open their sessions with `get_db_session(commit=True)` so the transaction begins as a writer (`BEGIN IMMEDIATE`) and `busy_timeout` (30s) absorbs contention — the same fix `post_download` already received for the identical failure shape.

Other collection write endpoints (create, tag, delete) still use the deferred session and are theoretically exposed to the same failure; kept out of scope to keep this fix targeted.

## Tests

- `test_write_config_data_failure_preserves_existing_file` — a failed write leaves the existing config untouched and a successful write leaves no temporary file behind (failed before: the file was truncated).
- `test_save_version_check_runs_under_config_save_lock` — the version-check read happens while holding the save lock (failed before).
- `test_item_write_endpoints_use_write_intent_sessions` — add/remove/reorder each open a `commit=True` session (failed before).
- Full backend suite: 1761 passed, 5 skipped. Frontend Jest: 1221 passed. No migrations, no frontend changes.

### DIFF
--- a/wrolpi/collections/api.py
+++ b/wrolpi/collections/api.py
@@ -10,6 +10,7 @@ from sanic_ext.extensions.openapi import openapi
 
 from wrolpi.api_utils import json_response
 from wrolpi.common import wrol_mode_check
+from wrolpi.db import get_db_session
 from wrolpi.errors import ValidationError
 from wrolpi.schema import JSONErrorResponse
 from . import lib, schema
@@ -89,20 +90,23 @@ async def create_collection_endpoint(request: Request, body: schema.CollectionCr
 @wrol_mode_check
 async def add_collection_item_endpoint(request: Request, collection_id: int, body: schema.AddItemRequest):
     """Append (or insert at ``position``) a file/zim/url item into the collection."""
-    session = request.ctx.session
-    try:
-        item, created = lib.add_collection_item(
-            session, collection_id, item_kind=body.item_kind,
-            file_group_id=body.file_group_id, zim_id=body.zim_id, zim_entry=body.zim_entry,
-            url=body.url, title=body.title, position=body.position)
-    except UnknownCollection as e:
-        return json_response({'error': str(e)}, status=HTTPStatus.NOT_FOUND)
-    except ValidationError as e:
-        return json_response({'error': str(e)}, status=HTTPStatus.BAD_REQUEST)
+    # Not `request.ctx.session`: this reads (the Collection, the FileGroup) before it INSERTs, and a
+    # deferred transaction fails that lock upgrade *instantly* with "database is locked" if anything
+    # else wrote in between (see `get_db_session`).
+    with get_db_session(commit=True) as session:
+        try:
+            item, created = lib.add_collection_item(
+                session, collection_id, item_kind=body.item_kind,
+                file_group_id=body.file_group_id, zim_id=body.zim_id, zim_entry=body.zim_entry,
+                url=body.url, title=body.title, position=body.position)
+        except UnknownCollection as e:
+            return json_response({'error': str(e)}, status=HTTPStatus.NOT_FOUND)
+        except ValidationError as e:
+            return json_response({'error': str(e)}, status=HTTPStatus.BAD_REQUEST)
 
-    # 201 when a new item was created; 200 when the item already existed (idempotent add).
-    status = HTTPStatus.CREATED if created else HTTPStatus.OK
-    return json_response({'item': item.dict()}, status=status)
+        # 201 when a new item was created; 200 when the item already existed (idempotent add).
+        status = HTTPStatus.CREATED if created else HTTPStatus.OK
+        return json_response({'item': item.dict()}, status=status)
 
 
 @collection_bp.delete('/<collection_id:int>/items/<item_id:int>')
@@ -112,11 +116,12 @@ async def add_collection_item_endpoint(request: Request, collection_id: int, bod
 @wrol_mode_check
 async def remove_collection_item_endpoint(request: Request, collection_id: int, item_id: int):
     """Remove a single item by its id and close the ordering gap."""
-    session = request.ctx.session
-    try:
-        removed = lib.remove_collection_item(session, collection_id, item_id)
-    except UnknownCollection as e:
-        return json_response({'error': str(e)}, status=HTTPStatus.NOT_FOUND)
+    # Write intent (see `add_collection_item_endpoint`).
+    with get_db_session(commit=True) as session:
+        try:
+            removed = lib.remove_collection_item(session, collection_id, item_id)
+        except UnknownCollection as e:
+            return json_response({'error': str(e)}, status=HTTPStatus.NOT_FOUND)
 
     if not removed:
         return json_response({'error': f'Item {item_id} not found in collection {collection_id}'},
@@ -137,15 +142,16 @@ async def remove_collection_item_endpoint(request: Request, collection_id: int, 
 async def reorder_collection_items_endpoint(request: Request, collection_id: int,
                                             body: schema.ReorderItemsRequest):
     """Set the item order from a full list of item ids (used by drag-and-drop)."""
-    session = request.ctx.session
-    try:
-        lib.reorder_collection_items(session, collection_id, body.item_ids)
-    except UnknownCollection as e:
-        return json_response({'error': str(e)}, status=HTTPStatus.NOT_FOUND)
-    except ValueError as e:
-        return json_response({'error': str(e)}, status=HTTPStatus.BAD_REQUEST)
+    # Write intent (see `add_collection_item_endpoint`).
+    with get_db_session(commit=True) as session:
+        try:
+            lib.reorder_collection_items(session, collection_id, body.item_ids)
+        except UnknownCollection as e:
+            return json_response({'error': str(e)}, status=HTTPStatus.NOT_FOUND)
+        except ValueError as e:
+            return json_response({'error': str(e)}, status=HTTPStatus.BAD_REQUEST)
 
-    data = lib.get_collection_with_stats(session, collection_id)
+        data = lib.get_collection_with_stats(session, collection_id)
     return json_response({'collection': data})
 
 
--- a/wrolpi/collections/api.py
+++ b/wrolpi/collections/api.py
@@ -90,23 +90,20 @@ async def create_collection_endpoint(request: Request, body: schema.CollectionCr
 @wrol_mode_check
 async def add_collection_item_endpoint(request: Request, collection_id: int, body: schema.AddItemRequest):
     """Append (or insert at ``position``) a file/zim/url item into the collection."""
-    # Not `request.ctx.session`: this reads (the Collection, the FileGroup) before it INSERTs, and a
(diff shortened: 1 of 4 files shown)


lrnselfreliance (commented):
## Summary

The two production races are addressed with the right tools: `write_config_data` now does same-directory `mkstemp` + `fsync` + `os.replace` (so concurrent readers cannot see a truncated YAML), and the playlist item write endpoints open `get_db_session(commit=True)`, which does `BEGIN IMMEDIATE` via `_immediate_txn`. `config_save_lock` is a `multiprocessing.Lock()` created in `main_process_startup` before Sanic forks, so moving the version check inside it does protect multi-worker Docker; atomic replace is still required because readers never take that lock. Dominant leftover risks are committing the write session on caught 4xx returns, and the same SQLite lock-upgrade race still present on other collection write endpoints that this PR left on `request.ctx.session`.

## Issue counts by severity

- bugs: 0
- suggestions: 5
- nits: 1

wrolpi/collections/api.py
```diff
@@ -89,20 +90,23 @@ async def create_collection_endpoint(request: Request, body: schema.CollectionCr
...
+    # Not `request.ctx.session`: this reads (the Collection, the FileGroup) before it INSERTs, and a
+    # deferred transaction fails that lock upgrade *instantly* with "database is locked" if anything
+    # else wrote in between (see `get_db_session`).
+    with get_db_session(commit=True) as session:
+        try:
+            item, created = lib.add_collection_item(
+                session, collection_id, item_kind=body.item_kind,
+                file_group_id=body.file_group_id, zim_id=body.zim_id, zim_entry=body.zim_entry,
+                url=body.url, title=body.title, position=body.position)
+        except UnknownCollection as e:
```
lrnselfreliance: **[suggestion]** These handlers catch `UnknownCollection` / `ValidationError` / `ValueError` inside `with get_db_session(commit=True)` and `return` an error response. `get_db_session` commits whenever the `with` block exits without an exception, so a 4xx now commits the write session. That is the opposite of `cleanup_session`, which rollbacks `request.ctx.session` on status >= 400, and unlike `post_download`, which lets failures propagate out of the `with` so the commit is skipped. Current error paths (`Collection.find_by_id`, validation before `_place_item` / `_shift_positions`) do not dirty the session, so this is not a live data bug — it is a footgun the next flush-then-validate change will trip.

**Suggestion:** Catch outside the `with`, or re-raise after mapping to HTTP, so a 4xx does not commit. Serialize ORM objects (`item.dict()`, `get_collection_with_stats`) inside the session, then return after it closes.

wrolpi/collections/api.py
```diff
@@ -89,20 +90,23 @@ async def create_collection_endpoint(request: Request, body: schema.CollectionCr
...
-    except ValidationError as e:
-        return json_response({'error': str(e)}, status=HTTPStatus.BAD_REQUEST)
-
-    # 201 when a new item was created; 200 when the item already existed (idempotent add).
-    status = HTTPStatus.CREATED if created else HTTPStatus.OK
-    return json_response({'item': item.dict()}, status=status)
+    # Not `request.ctx.session`: this reads (the Collection, the FileGroup) before it INSERTs, and a
+    # deferred transaction fails that lock upgrade *instantly* with "database is locked" if anything
+    # else wrote in between (see `get_db_session`).
+    with get_db_session(commit=True) as session:
```
lrnselfreliance: **[suggestion]** The item endpoints now take the write lock at BEGIN, which matches the production `INSERT INTO collection_item` failure. `create_collection_endpoint`, `put_collection_endpoint`, `tag_collection_endpoint`, and `delete_collection_endpoint` still use the deferred `request.ctx.session` and have the same read-then-write shape (`find_by_id` / existence check, then INSERT/UPDATE/DELETE). Under WAL they will still fail instantly with `database is locked` whenever another connection writes between the read and the first write. Correctly out of scope for this PR, but it is the same defect, not a theoretical one.

**Suggestion:** Apply `get_db_session(commit=True)` to those write handlers in a follow-up (and keep the 4xx-commit pattern from Issue 1 in mind). Track it if this PR is meant to close the class of failure, not just the playlist-item log line.

wrolpi/collections/api.py
```diff
@@ -89,20 +90,23 @@ async def create_collection_endpoint(request: Request, body: schema.CollectionCr
...
-            url=body.url, title=body.title, position=body.position)
-    except UnknownCollection as e:
-        return json_response({'error': str(e)}, status=HTTPStatus.NOT_FOUND)
-    except ValidationError as e:
-        return json_response({'error': str(e)}, status=HTTPStatus.BAD_REQUEST)
-
-    # 201 when a new item was created; 200 when the item already existed (idempotent add).
-    status = HTTPStatus.CREATED if created else HTTPStatus.OK
-    return json_response({'item': item.dict()}, status=status)
+    # Not `request.ctx.session`: this reads (the Collection, the FileGroup) before it INSERTs, and a
```
lrnselfreliance: **[suggestion]** The new comments narrate the change and embed SQLite lock-upgrade design history (`Not request.ctx.session`, `fails that lock upgrade *instantly*`, `Write intent (see add_collection_item_endpoint)`). `get_db_session`'s docstring already explains BEGIN IMMEDIATE; repeating it at three handlers (and copying the `post_download` essay) will rot.

**Suggestion:** Drop the block comments. A one-line pointer at most, or none — the `get_db_session(commit=True)` call is the documentation.

wrolpi/common.py
```diff
@@ -790,6 +793,10 @@ def version(self) -> int:
 def write_config_data(config: dict, config_file: pathlib.Path, width: int = 90):
     """Write a config dict to a YAML file.  Decimals are converted to str; all other Python objects are rejected.
 
+    The write is atomic (temporary file, then rename): another process reading the config can only
```
lrnselfreliance: **[suggestion]** The `write_config_data` docstring and the comments at the version check / `mkstemp` / `fsync` restate what the code does and leak deployment architecture (`Sanic workers are separate processes, so in-process locks cannot provide this guarantee`). That sentence is easy to misread: this process *does* have a cross-worker `multiprocessing.Lock`; atomic replace is needed because *readers* never take it.

**Suggestion:** Keep the function's contract (atomic YAML write, Decimal-safe, fsync'd). Move the worker/lock rationale to the PR or a short comment at `config_save_lock` if it must live in-tree. The `mkstemp` same-directory and 0600-then-chmod notes are the only WHY comments worth keeping, and they can be one line each.

wrolpi/collections/test/test_playlist_items.py
```diff
@@ -156,6 +156,44 @@ async def test_add_file_item(async_client, test_session, test_directory):
...
+        item_id = response.json['item']['id']
+
+        _, response = await async_client.put(f'/api/collections/{cid}/items/order',
+                                             json={'item_ids': [item_id]})
+        assert response.status_code == HTTPStatus.OK, response.json
+
+        _, response = await async_client.delete(f'/api/collections/{cid}/items/{item_id}')
+        assert response.status_code == HTTPStatus.NO_CONTENT
+
+    assert len(commit_kwargs) == 3, 'Each item write endpoint must open its own session'
```
lrnselfreliance, line 190: **[suggestion]** `test_item_write_endpoints_use_write_intent_sessions` only records that `get_db_session` was called with `commit=True`. It does not check `BEGIN IMMEDIATE` or write-lock ownership (`test_write_session_takes_write_lock_up_front` already covers the helper). `test_write_config_data_failure_preserves_existing_file` proves a failed dump does not clobber the dest and that a success leaves no temps; it does not prove a concurrent reader cannot observe a truncated file (the docstring claims that). `test_save_version_check_runs_under_config_save_lock` is a solid in-process assertion that the version read happens while the lock wrapper is held.

**Suggestion:** Narrow the test docstrings to what they actually assert. Optional: after a failed `write_config_data`, also assert no `*.tmp` leftovers (the success path already does).

wrolpi/common.py
```diff
@@ -798,14 +805,25 @@ class _ConfigDumper(yaml.CSafeDumper):
...
+        with os.fdopen(fd, 'wt') as fh:
+            try:
+                yaml.dump(config, fh, Dumper=_ConfigDumper, width=width, sort_keys=True)
+            except yaml.representer.RepresenterError as e:
+                raise ValueError(f'Config contains a Python object that cannot be serialized: {e}') from e
+            # Wait for data to be written before the rename publishes it.
+            fh.flush()
+            os.fsync(fh.fileno())
+        os.replace(temporary_path, config_file)
+    except BaseException:
```
lrnselfreliance, line 819: **[nit]** If `stat`/`chmod` fails before `os.fdopen(fd, ...)` takes ownership, the `except` unlinks the path but never `os.close(fd)`. The leak is only on that rare path (dump/`replace` failures close the fd via the `with`).

**Suggestion:** Close the fd in `except` when `fdopen` has not run yet (set `fd = None` after `fdopen` succeeds so you do not double-close).

lrnselfreliance:
Addressed the review in cf4c1a17:

- **4xx commits the write session** — fixed. Exceptions now propagate out of the `with get_db_session(commit=True)` block (skipping the commit; the context manager rolls back) and are mapped to HTTP outside it. Serialization (`item.dict()`, `get_collection_with_stats`) happens inside the session.
- **Other write endpoints have the same defect** — tracked in #674 (create/put/tag/delete), including the 4xx-commit pattern to follow.
- **Narrative comments at the handlers** — dropped; one one-line pointer remains on the first handler.
- **`write_config_data` docstring/comments** — trimmed to the contract; the misleading in-process-locks sentence is gone; the same-directory and chmod notes are one line each.
- **Test docstrings overclaim** — narrowed to what the assertions cover; the failed-write path now also asserts no `*.tmp` leftovers.
- **fd leak when `chmod` fails before `fdopen`** — fixed; the descriptor is closed in the except path unless the file object took ownership.

Tests: test_common (152 passed), playlist items + collections api suites (40 passed).

PR #675, opened 2026-08-21
videos: an uploaded audio file never appeared in its Channel

lrnselfreliance:
## Summary

An audio file uploaded into a Channel's directory never appeared on the Channel's page, even though FileBrowser could list and play it. The upload flow and the refresh flow disagree about what a Video is:

- The batch `video_modeler` (refresh path) models `video/%` **and** `audio/%` — thousands of audio FileGroups are Channel Videos this way.
- The upload flow (`upsert_file` → `FileGroup.do_model`) asks `Video.can_model`, which only accepted `video/%`. The audio file fell through to the plain-index fallback: `indexed=True`, no Video row, `model=NULL`.

Worse, the state was permanent: the modeler's `indexed != True` gate excluded the file on every later refresh.

## Changes

- `Video.can_model` accepts `audio/%`, matching the batch modeler — an upload now creates the Video (and `Video.validate` claims the Channel by directory) immediately.
- The modeler now gates on the Video row being missing rather than solely on `indexed == False`, mirroring the identical fix already made to `doc_modeler` — so files already stuck in the broken state heal on the next refresh, with a `failed_ids` set so a file that fails to model cannot be re-selected forever within one run.
- Audio playlists (`.m3u` detects as `audio/x-mpegurl`) are excluded from both paths — the modeler's blanket `audio/%` would otherwise turn a text playlist into a "Video" — and `video_cleanup` unmodels any playlist an older modeler already converted.

## Testing

TDD: four new tests in `modules/videos/test/test_audio_modeling.py` written first (three failed on the bug):
- uploaded audio becomes a Video claimed by its Channel,
- an `.m3u` is never modeled by either path,
- an indexed-but-unmodeled audio FileGroup is healed by the modeler and claimed by its Channel,
- `video_cleanup` unmodels a wrongly-modeled playlist.

Full backend suite: 1765 passed, 5 skipped. Frontend Jest: 1221 passed.

### DIFF
--- a/modules/videos/__init__.py
+++ b/modules/videos/__init__.py
@@ -21,6 +21,28 @@ __all__ = ['video_modeler']
 VIDEO_PROCESSING_LIMIT = 20
 
 
+def _model_video(session: Session, file_group: FileGroup, video: Video | None, probed: dict) -> Video:
+    """Create/update and validate the Video row for one FileGroup."""
+    if not video:
+        video = Video(file_group=file_group, file_group_id=file_group.id)
+        session.add(video)
+        session.flush([video])
+    if not Session.object_session(video):
+        session.add(video)
+        video.flush()
+    # Store the ffprobe data gathered outside the write transaction.
+    if (result := probed.get(file_group.id)) is not None:
+        video.ffprobe_json, ffprobe_file = result
+        if ffprobe_file:
+            # Track the .ffprobe.json cache file that was just written.
+            file_group.append_files(ffprobe_file)
+    video.flush(session)
+    # Validate and index subtitles.  (Poster generation happens here when a Channel
+    # asks for it; it is the remaining slow work inside this transaction.)
+    video.validate(session)
+    return video
+
+
 @register_modeler
 async def video_modeler(progress_callback: Callable[[int], None] = None):
     total_processed = 0
@@ -35,9 +57,7 @@ async def video_modeler(progress_callback: Callable[[int], None] = None):
                 .filter(
                 or_(FileGroup.mimetype.like('video/%'), FileGroup.mimetype.like('audio/%')),
                 FileGroup.mimetype.notin_(AUDIO_PLAYLIST_MIMETYPES),
-                # Model anything without a Video row, even if something else (an upload, an
-                # indexer) already set `indexed=True` -- gating solely on `indexed == False`
-                # left such files permanently invisible in their Channels.
+                # Also model indexed FileGroups that never got a Video row.
                 or_(Video.id.is_(None), FileGroup.indexed != True),
             )
             if failed_ids:
@@ -68,32 +88,14 @@ async def video_modeler(progress_callback: Callable[[int], None] = None):
                 .outerjoin(Video, Video.file_group_id == FileGroup.id))
 
             for file_group, video in file_groups:
-                video_id = None
                 try:
-                    if not video:
-                        video = Video(file_group=file_group, file_group_id=file_group.id)
-                        session.add(video)
-                        session.flush([video])
-                    video_id = video.id
-                    if not Session.object_session(video):
-                        session.add(video)
-                        video.flush()
-                    # Store the ffprobe data gathered above.
-                    if (result := probed.get(file_group.id)) is not None:
-                        video.ffprobe_json, ffprobe_file = result
-                        if ffprobe_file:
-                            # Track the .ffprobe.json cache file that was just written.
-                            file_group.append_files(ffprobe_file)
-                    video.flush(session)
-                    # Validate and index subtitles.  (Poster generation happens here when a Channel
-                    # asks for it; it is the remaining slow work inside this transaction.)
-                    video.validate(session)
+                    _model_video(session, file_group, video, probed)
                 except Exception as e:
+                    # Before the re-raise, so no failure path can re-select this id forever.
+                    failed_ids.add(file_group.id)
                     if PYTEST:
                         raise
-                    failed_ids.add(file_group.id)
-                    i = video.file_group.primary_path if video.file_group else video_id
-                    logger.error(f'Unable to model Video: {str(i)}', exc_info=e)
+                    logger.error(f'Unable to model Video: {file_group.primary_path}', exc_info=e)
 
                 file_group.indexed = True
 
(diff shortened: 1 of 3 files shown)


modules/videos/models.py
```diff
@@ -20,7 +20,10 @@
 
 logger = logger.getChild(__name__)
 
-__all__ = ['Video', 'Channel']
+__all__ = ['Video', 'Channel', 'AUDIO_PLAYLIST_MIMETYPES']
+
+# .m3u playlists detect as audio/* but are text lists of URLs, not media.
+AUDIO_PLAYLIST_MIMETYPES = ('audio/x-mpegurl', 'audio/mpegurl')
```
lrnselfreliance: **[suggestion]** `AUDIO_PLAYLIST_MIMETYPES` only lists `audio/x-mpegurl` and `audio/mpegurl`. That covers `.m3u` as detected here, and `application/vnd.apple.mpegurl` / `application/xspf+xml` never match the `video/*`/`audio/*` allowlist so they are already ignored. `.pls` does not: libmagic typically reports `audio/x-scpls`, which starts with `audio/`, so `can_model`, `video_modeler`, and `video_cleanup` will still treat a Winamp/PLS playlist as a Video — the same class of bug this PR is fixing for `.m3u`.

**Suggestion:** Add `audio/x-scpls` (and any other `audio/*` playlist types you care about) to `AUDIO_PLAYLIST_MIMETYPES`, and extend `test_audio_playlist_is_not_a_video` / `test_video_cleanup_unmodels_playlists` with a `.pls`. `application/vnd.apple.mpegurl` is optional given the current allowlist.

modules/videos/__init__.py
```diff
@@ -24,13 +24,25 @@
 @register_modeler
 async def video_modeler(progress_callback: Callable[[int], None] = None):
     total_processed = 0
+    # Ids that failed to model this run.  A failure leaves no Video row, so the `Video.id IS
+    # NULL` gate below would re-select it forever; failures are retried on the next refresh.
+    failed_ids: set = set()
```
lrnselfreliance, line 51: **[suggestion]** `failed_ids` is per-run only (correct; the next refresh retries), and it is the only thing that stops `Video.id IS NULL` from re-selecting a file that failed before a Video row was flushed. There is no test that the loop actually terminates when modeling keeps failing, unlike `test_doc_modeler_terminates_on_persistent_failure`. Also, `failed_ids.add` sits after `if PYTEST: raise` and is skipped entirely for ffprobe failures (those still create a Video). The dangerous case is a persistent exception before `session.flush([video])` with a full batch of such files — without a test, a future edit that adds to `failed_ids` too late, or not on every path, hangs the modeler.

**Suggestion:** Copy the docs modeler's termination test: patch modeling to always raise with `PYTEST` disabled, create more than `VIDEO_PROCESSING_LIMIT` indexed audio/video FileGroups with no Video row, and `asyncio.wait_for(video_modeler(), timeout=...)`. Keep adding to `failed_ids` on every path that leaves no Video row.

modules/videos/__init__.py
```diff
@@ -24,13 +24,25 @@
...
         with get_db_session() as session:
-            batch: List[Tuple[int, pathlib.Path]] = list(session.query(FileGroup.id, FileGroup.primary_path).filter(
-                FileGroup.indexed != True,
+            query = session.query(FileGroup.id, FileGroup.primary_path) \
+                .outerjoin(Video, Video.file_group_id == FileGroup.id) \
+                .filter(
                 or_(FileGroup.mimetype.like('video/%'), FileGroup.mimetype.like('audio/%')),
-            ).limit(VIDEO_PROCESSING_LIMIT).all())
+                FileGroup.mimetype.notin_(AUDIO_PLAYLIST_MIMETYPES),
+                # Model anything without a Video row, even if something else (an upload, an
```
lrnselfreliance: **[suggestion]** This comment (and the matching one on `Video.can_model`) narrates the old bug and the upload-vs-modeler split instead of stating the invariant. Same for the test module docstring, which walks through the previous `indexed != True` gate.

**Suggestion:** Keep a short WHY, drop the history. For example: "Also model indexed FileGroups that never got a Video row." On `can_model`: "Audio in a Channel is Channel content; playlists are not media."

lrnselfreliance (commented):
## Summary

The upload/modeler split is real, and this fix lines the two paths up: `Video.can_model` now accepts `audio/*` (minus playlist mimetypes), the batch modeler heals indexed FileGroups that never got a Video row, and `video_cleanup` unmodels playlists an older modeler already converted. Channel claiming on upload is handled by `Video.validate` → `Channel.get_by_path`, which the new tests actually exercise. Residual risk is small: the playlist denylist is m3u-only, the new anti-join can re-select a failed file forever within one run if `failed_ids` is skipped, and a few comments restate the bug history instead of the invariant.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 1

modules/videos/__init__.py
```diff
@@ -162,14 +175,16 @@ def _claim_videos(claims: List[Tuple[int, int]]):
 @limit_concurrent(1)
 def video_cleanup():
     logger.info('Claiming Videos for their Channels')
-    # Read the FileGroups that are no longer video/audio, then unmodel them in short transactions.
+    # Read the FileGroups that are no longer video/audio (audio playlists never were),
+    # then unmodel them in short transactions.
     with get_db_curs() as curs:
-        curs.execute('''
+        playlists = ', '.join(f"'{mt}'" for mt in AUDIO_PLAYLIST_MIMETYPES)
```
lrnselfreliance: **[nit]** Playlist mimetypes are interpolated into SQL with an f-string (`IN ({playlists})`). The values are a module constant so this is not injection today, but the rest of this function already passes lists through `json_each(:ids)`.

**Suggestion:** Bind `AUDIO_PLAYLIST_MIMETYPES` the same way as `stale_ids` (`json.dumps` + `json_each`) so the denylist cannot break the statement if it is later extended.

modules/videos/models.py
lrnselfreliance: Fixed in 80faa8bc — added audio/x-scpls and audio/scpls to the denylist. Note detection varies: this box's libmagic reports .pls as text/plain, so the tests set the mimetype directly and parametrize over AUDIO_PLAYLIST_MIMETYPES (can_model rejects every entry; video_cleanup unmodels every entry), keeping the .m3u test as the real-detection path.

modules/videos/__init__.py
lrnselfreliance: Fixed in 80faa8bc — per-file modeling is extracted as _model_video (mirroring _model_doc), failed_ids.add now runs before the PYTEST re-raise so every failure path records the id, and test_video_modeler_terminates_on_persistent_failure patches _model_video to always raise with PYTEST disabled, creates VIDEO_PROCESSING_LIMIT+5 indexed audio FileGroups with no Video row, and asserts the loop finishes under asyncio.wait_for.

modules/videos/__init__.py
lrnselfreliance: Fixed in 80faa8bc — the modeler gate comment is now 'Also model indexed FileGroups that never got a Video row', can_model says 'Audio in a Channel is Channel content; playlists are not media', and the test module docstring states the invariant the two paths must share instead of the old gate's history.

modules/videos/__init__.py
lrnselfreliance: Fixed in 80faa8bc — the playlist mimetypes are bound via json.dumps + json_each(:playlists), same as stale_ids.

lrnselfreliance:
CI triage: the .m3u detection failure is fixed as of 77978bfc — `test_audio_modeling` passes on the latest api-tests run. The remaining api-tests failure, `wrolpi/files/test/test_api.py::test_rename_directory_stale_destination_fails_promptly` (TimeoutError), is a pre-existing flake: it also failed on master build 23687 (f0249461), before this branch existed, and it passes locally in isolation. It is unrelated to this PR's diff (video modeling only).

PR #678, opened 2026-08-29
Fix API stalls: non-blocking switch worker, threaded searches, indexed Videos page

lrnselfreliance:
## Summary

Benchmarking every read-only API endpoint hot and cold on real libraries (10 GB SQLite DB on spinning disks, Pi 5 and x86) turned up three causes of slow responses. This PR fixes them and adds the benchmark tooling.

### 1. `switch_worker` blocked the event loop (`wrolpi/switches.py`)
The worker is an asyncio coroutine re-run every 0.1 s, but it called the blocking `multiprocessing.Event.wait(timeout=1)`. With no switch pending — the normal state — each Sanic worker's loop was frozen ~90% of the time, so **about a third of all requests, including `/api/echo` and `/api/settings`, took ~1,000 ms**. It now checks `is_set()` and returns.

### 2. Synchronous queries on the event loop
`videos/search`, `files/search`, `archive/search`, `docs/search`, `zim/search` and `/api/statistics` ran their SQLite work inline, so a single slow search stalled every other request on that worker (a 20 ms `echo` was observed waiting 60 s behind a search). They now use `asyncio.to_thread`, the pattern already used by `modules/flasher/api.py`.

### 3. The unfiltered Videos page scanned every video (`modules/videos/video/lib.py`)
`mimetype LIKE 'video/%' OR mimetype LIKE 'audio/%'` + `COUNT(*) OVER()` forces SQLite to visit every video's index entry before `LIMIT`: 8.5 s cold on a 100k-video library, 10 s hot for page 50. `_search_videos_by_date` now seeks each exact mimetype on the `(mimetype, effective_datetime)` index for `limit + offset` rows and merges them; the total comes from a separate count cached for 60 s (bypassed under pytest). Results are byte-identical to the generic query, including the id tiebreaker and NULL placement. Only the no-filter, date-ordered request takes this path; searches, tags, channels and other orders are unchanged.

### 4. Video description served by its own endpoint (`ea10d68f`)
`Video.__json__` computed `description` as `c_text or get_video_description()`, and the fallback opens and parses the video's `info.json` from disk. List endpoints serialize every result with `__json__`, so any page containing videos without an indexed description (15k of them on one library) read those files on every request, cold, for a field the list never shows. The description is not part of search either (search uses the FTS copy of `c_text`).

`description` is dropped from `__json__` and served by `GET /api/videos/<id>/description`, like comments and captions (the fallback to `info.json` lives only there, for rows indexed before descriptions were stored). The player page fetches it through `useVideoExtras`. Tests assert list/detail responses never touch `info.json`.

**Merge note for the `ai` branch:** `modules/ai/lib.py` reads `video.get('description')` from the Video JSON; it should call `Video.get_description()` (or use `c_text`) after this lands.

### Tooling
`scripts/api_perf.py` catalogs the read-only endpoints (GETs plus the `POST .../search` and `.../estimates` calls the UI issues), discovers real ids, and times them; `--cold` restarts the API over SSH (dropping the OS page cache where possible) and refuses the Caddy 302 → Controller fallback page that would otherwise look like a healthy API. `scripts/api_perf_compare.py` merges per-host CSVs. Host profiles live in gitignored `scripts/api_perf_hosts.json` (example committed).

## Measured (docker host, 10 GB library, before → after)

| | before | after |
|---|---:|---:|
| Median of all hot requests | 1,007 ms | 22 ms |
| Requests stuck at ~1 s | 78 / 258 | 3 / 172 |
| `videos/search` page 1, hot | 287 ms | 108 ms |
| `videos/search` page 50, hot | 1,345 ms | 120 ms |

## Not fixed here (follow-ups)
- Cold first Videos page is still ~13 s on a fragmented HDD library: the video count (1–4 s), `video.channel` lazy loads (~2 s), and general index-page fragmentation. A `VACUUM INTO` copy of the same DB answered the old page query in 0.2 s instead of 8.5 s cold, so a VACUUM policy is the biggest remaining lever.

## Tests
- `test_video_description_endpoint`, `test_video_json_does_not_read_info_json`, and a `useVideoExtras` hook test
- Cypress: 66/66 specs passed against the local dev stack
- `test_switch_worker_does_not_block_event_loop` (fails at 1.0 s on the old code)
- `test_search_videos_does_not_block_event_loop` (fails when `to_thread` is forced inline)
- `test_search_videos_default_page_pagination` (ties, NULL dates, both directions, pagination, fast path taken only for unfiltered date orders)
- `test_cached_total`
- Full backend suite: 1781 passed, 5 skipped. Frontend: 1221 passed (unchanged).


lrnselfreliance (commented):
## Summary

The three stall fixes are well-targeted and look correct: `switch_worker` no longer blocks the loop (`is_set()` plus the existing 0.1s `perpetual_signal` sleep is a poll, not a busy-wait; switches can be delayed by at most one tick, not missed), searches/statistics now run via `asyncio.to_thread` and open their own DB sessions inside the worker thread (same pattern as `flasher/api.py`), and the unfiltered Videos date path is a sound per-mimetype top-N merge on `file_group_mimetype_effective_idx` with an id tiebreaker that matches `VIDEO_ORDERS`. Dominant risks are the one `to_thread` call that passes a request-scoped SQLAlchemy `Session` across threads, a sibling libzim endpoint left on the loop, and a pagination test that never actually compares the fast path to the generic query.

## Issue counts by severity

- bugs: 1
- suggestions: 4
- nits: 1

modules/zim/api.py
```diff
@@ -65,8 +66,9 @@ async def post_set_zim_auto_search(request: Request, zim_id: int, body: schema.Z
 @validate(schema.ZimSearchRequest)
 async def search_all_zims(request: Request, body: schema.ZimSearchRequest):
     session = request.ctx.session
-    results = lib.search_all_zims(session, body.search_str, tag_names=body.tag_names, offset=body.offset,
-                                  limit=body.limit)
+    # libzim searches are synchronous and slow on a Pi; keep them off the event loop.
+    results = await asyncio.to_thread(lib.search_all_zims, session, body.search_str, tag_names=body.tag_names,
```
lrnselfreliance: **[bug]** `search_all_zims` is the only converted endpoint that passes `request.ctx.session` (created on the event-loop thread by `inject_session`, committed/closed there by `cleanup_session`) into `asyncio.to_thread`. SQLAlchemy Sessions are not thread-safe even when used sequentially; this is the exact hazard `post_search_suggestions` already documents, and it is the opposite of `flasher/api.py` / `search_videos` / `search_files` / `search_archives` / `_search_docs`, which create connections inside the threaded function via `get_db_session`/`get_db_curs`. `check_same_thread=False` only covers the sqlite3 connection, not Session identity-map/transaction state. Under load this can raise or return inconsistent rows; under pytest the shared `test_session` is used from two threads.

**Suggestion:** Do not pass `request.ctx.session` into the executor. Open a session inside `lib.search_all_zims` (or wrap the call in `get_db_session()` in the thread) and return plain dicts, matching the other search endpoints.

lrnselfreliance: **[suggestion]** `/api/zim/search` was moved off the loop because "libzim searches are synchronous and slow on a Pi", but the sibling handlers still run libzim on the event loop: `search_zim` (`POST /api/zim/search/<id>`) and `post_search_estimates` (`Zims.estimate` / `entries_with_tags`). A single-zim search or estimate can still stall every other request on that worker. Default `asyncio` pool size (`min(32, cpu_count+4)`, 8 threads on a Pi 5) is adequate for these read-only searches; no change needed there as long as sessions are not shared.

**Suggestion:** Run `headline_zim` / `Zims.estimate` through `asyncio.to_thread` as well, creating the session inside the thread rather than capturing `request.ctx.session`.

wrolpi/switches.py
```diff
@@ -99,13 +99,12 @@ async def switch_worker():
     switch_name = None
     try:
         # Get new switch and its context.  Handle each switch one at a time.
-        # During testing, don't block the event loop - check for pending switches directly.
-        # In production, use blocking wait which is more efficient for long-running processes.
-        if PYTEST and not api_app.shared_ctx.switches:
-            return
-        else:
-            switches_changed.wait(timeout=1)
+        # This coroutine runs on the Sanic worker's event loop and must never block (a blocking
```
lrnselfreliance: **[suggestion]** The replacement comment narrates the old `Event.wait` bug and restates how `perpetual_signal` works. The 0.1s sleep is still there (`@perpetual_signal(sleep=0.1)`), so this is not a busy-poll; `is_set()` plus the dict emptiness check is enough to pick up a switch on the next tick. The early `return` is inside the `try`, so `finally: switches_changed.clear()` still runs on the empty path and can clear an event set between the check and the return — the pending dict entry is still handled 0.1s later, so this is not a miss, just surprising.

**Suggestion:** Drop the history. A short "must not block the event loop; perpetual_signal retries every 0.1s" is enough. Optionally `return` before the `try` so the empty path does not clear the event.

modules/videos/video/api.py
```diff
@@ -57,7 +58,10 @@ async def search_videos(_: Request, body: schema.VideoSearchRequest):
     if body.order_by not in lib.VIDEO_ORDERS:
         raise InvalidOrderBy('Invalid order by')
 
-    file_groups, videos_total = lib.search_videos(
+    # The query is synchronous SQLite; run it off the event loop so a slow search on a large
```
lrnselfreliance: **[suggestion]** The same two-line "The query is synchronous SQLite; run it off the event loop..." comment is copied into videos, files, archive, and docs search. It restates the call and the PR narrative; the `to_thread` is obvious.

**Suggestion:** Delete the copies (or keep a single-line why on the first one only).

modules/videos/test/test_api.py
```diff
@@ -588,3 +591,102 @@ async def test_video_upload_file_tracking(test_session, async_client, video_fact
...
+    assert echo_elapsed < 0.4, f'echo waited {echo_elapsed:.2f}s behind a blocking search'
+
+
+@pytest.mark.asyncio
+async def test_search_videos_default_page_pagination(test_session, video_factory, audio_factory, assert_video_search):
+    """The unfiltered Videos page (no search/tags/channel) pages through every video and audio file in
+    date order, including files sharing a date and files without one, and reports the true total.
+
+    This page takes a fast path (one index range per mimetype instead of scanning every video), so
+    it must produce exactly what the generic query produces."""
```
lrnselfreliance, line 631: **[suggestion]** The docstring says the fast path "must produce exactly what the generic query produces", but the test never invokes the generic path for the same inputs. It only asserts `fast_path.called`, then hard-coded ids. `assert fast_path.called` does fail on the old code; the id/pagination asserts would also pass against the old `COUNT(*) OVER()` query for this 6-row fixture, so they do not lock merge identity (NULL placement, id tiebreaker, multi-mimetype interleaving) against a regression that changes both paths. Empty-page totals already diverge: generic `COUNT(*) OVER()` yields `total=0` when `OFFSET` is past the end (`results else 0`), while the fast path returns the cached count (the `offset=6, assert_total=6` case).

**Suggestion:** For the unfiltered date orders, also call `search_videos` with the fast path disabled (empty `INDEXED_DATE_ORDERS`, or a no-op filter that does not change membership) and assert equal id lists. Treat empty-page `total` as an intentional difference or align the two paths.

modules/videos/video/lib.py
```diff
@@ -109,7 +186,12 @@ def search_videos(
 
     # Explicit JSON nulls bypass the schema defaults; SQLite rejects LIMIT/OFFSET NULL.
     limit = int(limit) if limit else 20
-    params = dict(search_str=search_str, offset=int(offset) if offset else 0)
+    offset = int(offset) if offset else 0
+    if not search_str and not tag_names and not channel_id and not censored and order in INDEXED_DATE_ORDERS:
+        # The Videos page as first opened: no filters, newest first.
```
lrnselfreliance: **[nit]** Comment says "The Videos page as first opened: no filters, newest first", but `INDEXED_DATE_ORDERS` also includes `published_datetime` (oldest first), which the new test covers.

**Suggestion:** Say "no filters, date order" (or drop the comment; the `INDEXED_DATE_ORDERS` check is enough).

modules/zim/api.py
lrnselfreliance: Fixed: `search_all_zims` now opens its own session inside the worker thread (`get_db_session()` in the threaded function) and returns the plain dicts `lib.search_all_zims` already produces; the request session is no longer passed across threads.

lrnselfreliance: Done: `search_zim` (`headline_zim`) and `post_search_estimates` (`Zims.estimate` / `entries_with_tags`) also run via `asyncio.to_thread`, each with a session opened in the thread.

wrolpi/switches.py
lrnselfreliance: Reworded to the invariant only, and the empty-path `return` now happens before the `try`, so it no longer clears the event.

modules/videos/video/api.py
lrnselfreliance: Removed the copies; a single one-line comment remains on `search_videos`.

modules/videos/test/test_api.py
lrnselfreliance: Added: for both `INDEXED_DATE_ORDERS` and several limit/offset pairs the test calls `search_videos` with the fast path disabled (`INDEXED_DATE_ORDERS` patched to `()`) and asserts identical id lists and totals. The empty-page total is kept as an intentional difference (the fast path reports the true total where `COUNT(*) OVER()` has no row to carry it) and documented in the test.

modules/videos/video/lib.py
lrnselfreliance: Fixed the comment: "no filters, in date order".

PR #701, opened 2026-09-04
videos: codec download preferences with transcode and strict-fail options

lrnselfreliance:
## What

Adds download-time control over video/audio encodings to the video, channel, and RSS downloaders, alongside the existing Video Resolutions selector:

* **Video Codecs / Audio Codecs** — ordered preference lists (video: h264, vp9, av1, vp8, hevc; audio: aac, opus, mp3, vorbis). Codec preference is **codec-major**: a preferred codec at a lower resolution is chosen over another codec at a higher resolution.
* **Transcode** — when no preferred codec is available at the source, the downloaded file is re-encoded with ffmpeg after download. Only the mismatching stream is re-encoded (the other is copied). Transcode targets are extensible maps (`TRANSCODE_VIDEO_TARGETS`/`TRANSCODE_AUDIO_TARGETS`), currently h264 (libx264) and aac/opus/mp3 — software VP9/AV1 encoding is impractical on a Raspberry Pi.
* **Fail if codecs unavailable** — when enabled (and transcode is not), a download with no matching format fails permanently instead of falling back; yt-dlp's "Requested format is not available" is treated as unrecoverable. Transcode makes strict moot, so the UI disables the strict toggle while transcode is on and the backend ignores it.

All four settings are inheritable (global config → channel/RSS → video) via `INHERITABLE_VIDEO_SETTINGS`, editable on the download forms (Advanced Settings), and on the Videos settings page.

## Invariants

* Empty codec lists produce a **byte-identical** yt-dlp command to before — existing behavior is unchanged unless codecs are selected.
* Generated format selectors contain no commas (one download per video) and codec patterns contain no `/`.
* The post-download ffprobe check runs before the Video record is created and holds no DB session, so a long ffmpeg run never blocks writers.
* A failed transcode leaves the original file untouched and no temporary file behind; a container change (webm→mp4) keeps the same stem so the FileGroup stays intact.
* Transcoding refuses to start with less than 1.5× the source size free on disk.

## Testing

* Backend: full suite `pytest -nauto` — 1807 passed. New tests cover selector generation, strict/defer outcomes, codec enforcement, transcode mechanics (tmp-then-rename, container force, disk guard), config round-trips, schema validation, and channel inheritance.
* Frontend: Jest 1229 passed (new `CodecSettings.test.js`), Cypress `Download.cy.js` 12 passed, plain `npm run build` clean.
* Verified the `[vcodec~='regex']` selector syntax parses in the pinned yt-dlp (2026.07.04).

### DIFF
--- a/modules/videos/downloader.py
+++ b/modules/videos/downloader.py
@@ -36,7 +36,7 @@ from wrolpi.files.worker import file_worker
 from wrolpi.vars import PYTEST, YTDLP_CACHE_DIR
 from .channel.lib import create_channel, get_channel
 from .common import get_no_channel_directory, update_view_counts_and_censored, \
-    ffmpeg_video_complete
+    ffmpeg_video_complete, get_or_create_ffprobe_json
 from .errors import UnknownChannel, ChannelDirectoryConflict, ChannelNameConflict, ChannelURLConflict, \
     ChannelSourceIdConflict
 from .cookies import cookies_unlocked, cookies_for_download
@@ -45,6 +45,8 @@ from .lib import get_videos_downloader_config, get_effective_video_settings, CHA
 from .models import Video, Channel
 from .normalize_video_url import normalize_video_url
 from .schema import ChannelPostRequest
+from .transcode import codecs_match, get_transcode_target, transcode_video_file, \
+    TRANSCODE_VIDEO_TARGETS, TRANSCODE_AUDIO_TARGETS
 from .video.lib import download_video_info_json
 
 logger = logger.getChild(__name__)
@@ -153,20 +155,118 @@ VIDEO_RESOLUTION_MAP = {
     'maximum': ['bestvideo*+bestaudio/best'],
 }
 
+# yt-dlp `vcodec`/`acodec` values vary by extractor (avc1.64001f, vp09.00.10.08, mp4a.40.2, ...);
+# match them by prefix regex.  These are used inside `-f` bracket filters, so they must never
+# contain a comma or a forward slash.
+VIDEO_CODEC_YT_DLP_PATTERNS = {
+    'h264': r'^(avc|h264)',
+    'hevc': r'^(hev|hvc|h265|hevc)',
+    'vp9': r'^vp0?9',
+    'av1': r'^av0?1',
+    'vp8': r'^vp0?8',
+}
+AUDIO_CODEC_YT_DLP_PATTERNS = {
+    'aac': r'^(aac|mp4a\.40)',
+    'opus': r'^opus',
+    'mp3': r'^(mp3|mp4a\.69|mp4a\.6b)',
+    'vorbis': r'^vorbis',
+}
+RESOLUTION_HEIGHTS = {'360p': 360, '480p': 480, '720p': 720, '1080p': 1080,
+                      '1440p': 1440, '2160p': 2160, 'maximum': None}
+
 # VTT is the only format that a browser can use to display captions on a <video/>.
 DEFAULT_CAPTION_FORMAT = 'vtt'
 DEFAULT_POSTER_FORMAT = 'jpg'
 DEFAULT_CHANNEL_DOWNLOAD_ORDER = 'newest'
 
 
-def format_selector(video_resolutions: List[str]) -> str:
-    """Flatten the requested resolutions into a single yt-dlp `-f` format selector.
+def format_selector(video_resolutions: List[str], video_codecs: List[str] = None,
+                    audio_codecs: List[str] = None, strict_codecs: bool = False) -> str:
+    """Flatten the requested resolutions (and codec preferences) into a single yt-dlp `-f`
+    format selector.
 
     The selectors are joined with "/" so yt-dlp downloads only the first one that matches.
     Joining with "," would instead ask yt-dlp to download *every* matching selector, which
     downloads (and post-processes) the same video once per match.
+
+    Codec preference is codec-major: a preferred codec at a lower resolution beats a
+    non-preferred codec at a higher resolution.  Unless `strict_codecs`, the legacy
+    resolution selectors are appended as a fallback so a video is still downloaded when no
+    preferred codec is available.
     """
-    return '/'.join(j for i in video_resolutions for j in VIDEO_RESOLUTION_MAP[i])
+    legacy = '/'.join(j for i in video_resolutions for j in VIDEO_RESOLUTION_MAP[i])
+    if not video_codecs and not audio_codecs:
+        return legacy
+
+    audio_selectors = [f"bestaudio[acodec~='{AUDIO_CODEC_YT_DLP_PATTERNS[i]}']" for i in audio_codecs or []]
+    if not strict_codecs or not audio_codecs:
+        audio_selectors.append('bestaudio')
+
+    video_filters = [f"[vcodec~='{VIDEO_CODEC_YT_DLP_PATTERNS[i]}']" for i in video_codecs] \
+        if video_codecs else ['']
+    selectors = []
+    for video_filter in video_filters:
+        for resolution in video_resolutions:
+            height = RESOLUTION_HEIGHTS[resolution]
+            dimensions = [f'[height={height}]', f'[width={height}]'] if height else ['']
+            for dimension in dimensions:
+                for audio_selector in audio_selectors:
+                    selectors.append(f'bestvideo{dimension}{video_filter}+{audio_selector}')
+    if not strict_codecs:
+        selectors.append(legacy)
+    return '/'.join(selectors)
+
+
+async def enforce_codecs(video_path: pathlib.Path, video_paths: List[pathlib.Path], effective: dict) \
+        -> Tuple[pathlib.Path, List[pathlib.Path]]:
+    """Verify the downloaded file's codecs against the effective codec preferences.
+
+    The format selector already prefers matching sources, but the legacy fallback selectors (and
+    lying extractor metadata) can still deliver a non-preferred codec.  On mismatch: transcode when
+    enabled, fail permanently when strict, otherwise keep the file (best-effort preference).
+
+    Runs before the Video record is created and holds no DB session, so ffmpeg may run for a long
+    time without blocking writers.
+    """
+    video_codecs = effective.get('video_codecs') or []
+    audio_codecs = effective.get('audio_codecs') or []
+    transcode = effective.get('transcode', False)
+    strict_codecs = effective.get('strict_codecs', False) and not transcode
+
+    data, _ = await get_or_create_ffprobe_json(video_path)
+    video_match, audio_match = codecs_match(data, video_codecs, audio_codecs)
+    if video_match and audio_match:
+        return video_path, video_paths
+
+    if transcode:
+        target_vcodec = None if video_match else get_transcode_target(video_codecs, TRANSCODE_VIDEO_TARGETS)
+        target_acodec = None if audio_match else get_transcode_target(audio_codecs, TRANSCODE_AUDIO_TARGETS)
+        if not target_vcodec and not target_acodec:
+            logger.warning(f'Cannot transcode {video_path}: none of the preferred codecs'
+                           f' (video={video_codecs}, audio={audio_codecs}) is a supported transcode target')
+            return video_path, video_paths
+        try:
+            from wrolpi.events import Events
+            Events.send_user_notify(f'Transcoding {video_path.name}')
+        except Exception:
+            # Events are best-effort; never let them break a download.
+            logger.debug(f'Failed to send transcode event for {video_path}', exc_info=True)
+        new_path = await transcode_video_file(video_path, target_vcodec, target_acodec,
+                                              container=effective.get('video_format') or 'mp4')
+        await get_or_create_ffprobe_json(new_path)  # Refresh the sidecar for the new streams.
+        video_paths = [new_path if i == video_path else i for i in video_paths]
+        return new_path, video_paths
+
+    if strict_codecs:
+        actual_video = [i.get('codec_name') for i in data.get('streams', []) if i.get('codec_type') == 'video']
+        actual_audio = [i.get('codec_name') for i in data.get('streams', []) if i.get('codec_type') == 'audio']
+        raise UnrecoverableDownloadError(
+            f'Downloaded video has codecs (video={actual_video}, audio={actual_audio}) but strict codecs'
+            f' (video={video_codecs}, audio={audio_codecs}) was required and transcoding is disabled')
+
+    logger.warning(f'{video_path} does not match preferred codecs'
+                   f' (video={video_codecs}, audio={audio_codecs}); keeping it as downloaded')
+    return video_path, video_paths
 
 
 def get_youtube_video_id(url: str) -> Optional[str]:
@@ -936,6 +1036,11 @@ class VideoDownloader(Downloader, ABC):
         effective = get_effective_video_settings(settings)
         audio_only = effective.get('audio_only', False)
         audio_format = effective.get('audio_format', 'mp3')
+        video_codecs = effective.get('video_codecs') or []
+        audio_codecs = effective.get('audio_codecs') or []
+        transcode = effective.get('transcode', False)
+        # Transcoding guarantees the codec, so strict is only meaningful without it.
+        strict_codecs = effective.get('strict_codecs', False) and not transcode
 
         # download_for_log carries .id (for progress) and .url (for log messages).
         # Falling back to a stub keeps unit tests viable.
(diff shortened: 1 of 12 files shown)


lrnselfreliance (commented):
## Summary

This PR adds inheritable video/audio codec preferences with yt-dlp format-selector construction, optional post-download ffmpeg transcode, and a strict-fail mode. The format-selector invariants check out (empty lists stay byte-identical to legacy, codec-major ordering is correct, patterns avoid `,`/`/`), inheritance/schema/UI wiring looks solid, and `transcode_video_file` handles tmp cleanup, disk guard, and stem-preserving container changes well. The dominant risk is the transcode/strict interaction: when transcode is enabled but no preferred codec is a supported target (e.g. av1-only), strict is still suppressed and the non-matching file is kept. Missing fsync before replace is the other durability gap worth fixing on Pi deployments.

## Issue counts by severity

- bugs: 1
- suggestions: 4
- nits: 1

modules/videos/downloader.py
```diff
@@ -153,20 +155,118 @@ def audio_output_extension(audio_format: str) -> str:
...
+    if video_match and audio_match:
+        return video_path, video_paths
+
+    if transcode:
+        target_vcodec = None if video_match else get_transcode_target(video_codecs, TRANSCODE_VIDEO_TARGETS)
+        target_acodec = None if audio_match else get_transcode_target(audio_codecs, TRANSCODE_AUDIO_TARGETS)
+        if not target_vcodec and not target_acodec:
+            logger.warning(f'Cannot transcode {video_path}: none of the preferred codecs'
+                           f' (video={video_codecs}, audio={audio_codecs}) is a supported transcode target')
+            return video_path, video_paths
```
lrnselfreliance: **[bug]** When `transcode` is set, `strict_codecs` is forced off (`and not transcode`) and a mismatch with no supported transcode target returns the original file after only a warning. A settings combo like `video_codecs=['av1'], transcode=True, strict_codecs=True` therefore keeps a vp9/h264 download instead of failing permanently. The UI can still submit both flags (enabling transcode only disables the strict toggle; it does not clear the value), and channel/API/YAML can set both explicitly. The comment that “transcoding guarantees the codec” is false whenever no preference is in `TRANSCODE_*_TARGETS`.

**Suggestion:** Only suppress strict when a transcode target actually exists (or after a successful transcode). If preferences mismatch and no target can fix them, fall through to the existing `UnrecoverableDownloadError` when `strict_codecs` was requested. Add a regression test for `transcode=True, strict_codecs=True, video_codecs=['av1']`.

modules/videos/downloader.py
```diff
@@ -153,20 +155,118 @@ def audio_output_extension(audio_format: str) -> str:
...
+    transcode = effective.get('transcode', False)
+    strict_codecs = effective.get('strict_codecs', False) and not transcode
+
+    data, _ = await get_or_create_ffprobe_json(video_path)
+    video_match, audio_match = codecs_match(data, video_codecs, audio_codecs)
+    if video_match and audio_match:
+        return video_path, video_paths
+
+    if transcode:
+        target_vcodec = None if video_match else get_transcode_target(video_codecs, TRANSCODE_VIDEO_TARGETS)
```
lrnselfreliance, line 242: **[suggestion]** Related partial-satisfy case: if video prefers only non-target codecs (av1) and audio prefers aac, and both mismatch, the code transcodes audio only and returns success with the non-preferred video stream still present. “Transcode to preferred codecs” then silently leaves the video codec wrong.

**Suggestion:** Treat “all mismatched preferences are resolvable by a target” as the condition for a successful transcode path; otherwise keep-or-strict as in Issue 1 rather than partially converting.

modules/videos/transcode.py
```diff
@@ -0,0 +1,143 @@
...
+            raise RuntimeError(
+                f'ffmpeg exited with {result.return_code} while transcoding {video_path}:'
+                f'\n{result.stderr.decode()[-2000:]}')
+    except Exception:
+        tmp_path.unlink(missing_ok=True)
+        raise
+
+    # The old ffprobe sidecar describes the old streams.
+    video_path.with_suffix('.ffprobe.json').unlink(missing_ok=True)
+    tmp_path.rename(final_path)
```
lrnselfreliance, line 196: **[suggestion]** The replace path is `tmp_path.rename(final_path)` with no `fsync` of the temp file (or its directory). On a Raspberry Pi power loss after rename, the final file can be truncated/corrupt even though ffmpeg exited 0. Elsewhere in the project (e.g. archive singlefile writes) durable writes use `flush` + `os.fsync`.

**Suggestion:** After a successful ffmpeg run, open the temp file, `flush`/`os.fsync`, then rename; optionally fsync the parent directory afterward.

modules/videos/downloader.py
```diff
@@ -153,20 +155,118 @@ def audio_output_extension(audio_format: str) -> str:
...
+    'h264': r'^(avc|h264)',
+    'hevc': r'^(hev|hvc|h265|hevc)',
+    'vp9': r'^vp0?9',
+    'av1': r'^av0?1',
+    'vp8': r'^vp0?8',
+}
+AUDIO_CODEC_YT_DLP_PATTERNS = {
+    'aac': r'^(aac|mp4a\.40)',
+    'opus': r'^opus',
+    'mp3': r'^(mp3|mp4a\.69|mp4a\.6b)',
```
lrnselfreliance: **[suggestion]** The mp3 yt-dlp pattern is case-sensitive (`mp4a\.6b`). Real `acodec` values are sometimes `mp4a.6B`; that would fail to match under yt-dlp’s `~=` regex and skip preferred mp3 formats.

**Suggestion:** Include both cases, e.g. `r'^(mp3|mp4a\.69|mp4a\.6[bB])'`, or otherwise case-fold in a way yt-dlp accepts.

modules/videos/transcode.py
```diff
@@ -0,0 +1,143 @@
...
+}
+FFPROBE_AUDIO_CODEC_NAMES = {
+    'aac': {'aac'},
+    'opus': {'opus'},
+    'mp3': {'mp3'},
+    'vorbis': {'vorbis'},
+}
+
+# Codecs which can be a transcode *target*, mapped to their ffmpeg encoder args.  Encoding VP9/AV1
+# in software is impractically slow on a Raspberry Pi, so only widely-playable targets are offered;
```
lrnselfreliance: **[suggestion]** Comments around transcode targets and the strict/transcode relationship restate design history (“extension point”, “Raspberry Pi”, “transcode guarantees the codec”) rather than a non-obvious constraint. The “guarantees the codec” wording in `downloader.py` / `Download.js` is also inaccurate given Issues 1–2.

**Suggestion:** Trim to the constraint only (e.g. why VP9/AV1 are omitted as targets), and fix or remove the incorrect “guarantees” claim once Issue 1 is addressed.

modules/videos/transcode.py
```diff
@@ -0,0 +1,143 @@
...
+            f'Not enough free space to transcode {video_path} ({free} bytes free, source is {source_size} bytes)')
+
+    video_args = TRANSCODE_VIDEO_TARGETS[target_vcodec] if target_vcodec else ('-c:v', 'copy')
+    audio_args = TRANSCODE_AUDIO_TARGETS[target_acodec] if target_acodec else ('-c:a', 'copy')
+
+    final_path = video_path.with_suffix(f'.{container}')
+    tmp_path = video_path.with_suffix(f'.transcode.{container}')
+    cmd = (FFMPEG_BIN, '-y',
+           '-i', video_path,
+           '-map', '0:v:0', '-map', '0:a:0',
```
lrnselfreliance: **[nit]** `-map 0:v:0` includes attached-picture streams; `-map 0:a:0` fails hard if audio is absent. WROLPi usually writes thumbnails as sidecars and rejects audio-less videos later, so this is uncommon, but a mismatched attached-pic-first file would transcode the wrong stream.

**Suggestion:** Prefer `-map 0:V:0` (exclude attached pics) and make audio optional (`0:a:0?`) when no audio stream is present, or build maps from the ffprobe data already loaded in `enforce_codecs`.

modules/videos/downloader.py
lrnselfreliance: Fixed in 2eaf6fc6: strict is now suppressed only when transcoding can produce every preferred codec (new `transcode_can_satisfy_codecs`). `video_codecs=['av1'], transcode=True, strict_codecs=True` now fails permanently, both at the format-selector level (fallbacks dropped) and in `enforce_codecs`. Regression tests added for both layers.

modules/videos/downloader.py
lrnselfreliance: Fixed in 2eaf6fc6: transcode now runs only when every mismatched stream has a target; an unfixable video mismatch blocks the audio-only conversion and falls through to strict/keep. Test added for the av1+aac partial case.

modules/videos/transcode.py
lrnselfreliance: Fixed in 2eaf6fc6: the temp file is fsynced before the rename and the parent directory after (directory fsync is best-effort since some filesystems refuse it).

modules/videos/downloader.py
lrnselfreliance: Fixed in 2eaf6fc6: pattern is now `^(mp3|mp4a\.69|mp4a\.6[bB])`.

modules/videos/transcode.py
lrnselfreliance: Fixed in 2eaf6fc6: target comment trimmed to the constraint (why VP9/AV1/HEVC are omitted), and the incorrect "transcoding guarantees the codec" wording replaced in downloader.py and the Download.js popup. The strict toggle is no longer disabled by transcode, since strict now matters when no transcode target exists.

modules/videos/transcode.py
lrnselfreliance: Fixed in 2eaf6fc6: maps are now `0:V:0` (excludes attached pictures) and `0:a:0?` (tolerates missing audio).

lrnselfreliance (commented):
## Summary

Follow-up commit `294dfe1c` (serialize transcodes machine-wide via shared_ctx lock):

This commit correctly replaces an asyncio-only serialization approach with a `multiprocessing.Lock` on `shared_ctx`, matching the established `config_save_lock` / `switches_lock` pattern and created in `main_process_ready` before Sanic workers fork. Non-blocking `acquire(False)` plus `asyncio.sleep` keeps the event loop responsive, and `try`/`finally` releases on success, failure, and task cancellation after the lock is held. Dominant residual risks are an orphaned lock if a worker dies mid-transcode (hold can last hours) and a test that only proves in-process waiting, not cross-process sharing.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 1

wrolpi/contexts.py
```diff
@@ -69,6 +69,10 @@ def attach_shared_contexts(app: Sanic):
     app.shared_ctx.perpetual_tasks_lock = multiprocessing.Lock()
 
     # Switches
+    # Transcoding saturates every core; only one ffmpeg transcode may run machine-wide,
+    # across all Sanic workers.
+    app.shared_ctx.transcode_lock = multiprocessing.Lock()
```
lrnselfreliance, line 75: **[suggestion]** `multiprocessing.Lock` is not released if the owning Sanic worker is SIGKILL'd or crashes while ffmpeg runs. Unlike the short-held `config_save_lock`, `transcode_lock` can be held for hours (`TRANSCODE_TIMEOUT` is 4h), so a single dead worker can block every future transcode machine-wide until a full API restart. The codebase already treats worker death as a real failure mode for other shared_ctx state (see perpetual-tasks owner PID/heartbeat).

**Suggestion:** Document the restart requirement, or track the holder PID alongside the lock and allow waiters to break/steal when the owner is gone (same idea as `_perpetual_owner_is_alive`), or use a file lock (`flock`) that the kernel drops on process exit.

modules/videos/transcode.py
```diff
@@ -0,0 +1,201 @@
...
+    list.  When False, enabling transcode cannot guarantee the preferences (e.g. av1-only), so
+    strict_codecs must still be honored."""
+    return ((not video_codecs or get_transcode_target(video_codecs, TRANSCODE_VIDEO_TARGETS) is not None)
+            and (not audio_codecs or get_transcode_target(audio_codecs, TRANSCODE_AUDIO_TARGETS) is not None))
+
+
+async def _acquire_transcode_lock():
+    """Block (asynchronously) until this process holds the machine-wide transcode lock.
+
+    Sanic runs multiple worker processes, so an asyncio primitive cannot serialize transcodes;
```
lrnselfreliance: **[suggestion]** The `_acquire_transcode_lock` docstring restates architecture/history ("Sanic runs multiple worker processes, so an asyncio primitive cannot…") rather than only the non-obvious WHY for polling.

**Suggestion:** Trim to something like: "Poll with sleep instead of blocking acquire so a multi-hour ffmpeg hold does not stall this worker's event loop."

modules/videos/test/test_transcode.py
```diff
@@ -0,0 +1,214 @@
...
+    with mock.patch('modules.videos.transcode.shutil.disk_usage', return_value=fake_usage), \
+            mock.patch('modules.videos.transcode.run_command') as mock_run:
+        with pytest.raises(RuntimeError, match='free space'):
+            await transcode_video_file(video_path, target_vcodec='h264')
+    mock_run.assert_not_called()
+
+
+@pytest.mark.asyncio
+async def test_transcode_lock_serializes(test_directory, async_client):
+    """Only one transcode runs at a time machine-wide: while the shared lock is held (by any
```
lrnselfreliance: **[suggestion]** The test docstring claims machine-wide serialization "while the shared lock is held (by any Sanic worker)", but the test only holds the lock in-process and asserts another task in the same process waits. It does not exercise a second OS process or prove `shared_ctx` inheritance across workers.

**Suggestion:** Soften the docstring to match what is actually tested (in-process wait/release via `shared_ctx.transcode_lock`), or add a multiprocess smoke test if cross-process behavior needs to be guaranteed in CI.

modules/videos/transcode.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    if not FFMPEG_BIN:
+        raise RuntimeError('ffmpeg was not found')
+    if not target_vcodec and not target_acodec:
+        raise RuntimeError('Refusing to transcode without a target codec')
+    if container not in TRANSCODE_CONTAINERS:
+        logger.info(f'Forcing mp4 container for transcode of {video_path} ({container} is not supported)')
+        container = 'mp4'
+
+    lock = await _acquire_transcode_lock()
```
lrnselfreliance, line 136: **[suggestion]** Cross-module side effect: `enforce_codecs` in `modules/videos/downloader.py` sends `Events.send_user_notify(f'Transcoding {video_path.name}')` before awaiting `transcode_video_file`. This commit makes that await potentially wait hours for the machine-wide lock before ffmpeg starts, so the notify can fire long before any transcode work begins.

**Suggestion:** Move the user notify to after the lock is acquired (e.g. at the start of `_transcode_video_file`), or change the pre-wait message to indicate queuing/waiting for an in-progress transcode.

wrolpi/contexts.py
```diff
@@ -69,6 +69,10 @@ def attach_shared_contexts(app: Sanic):
     app.shared_ctx.perpetual_tasks_lock = multiprocessing.Lock()
 
     # Switches
+    # Transcoding saturates every core; only one ffmpeg transcode may run machine-wide,
```
lrnselfreliance, line 72: **[nit]** `transcode_lock` is inserted under the `# Switches` section comment even though it is unrelated to switches.

**Suggestion:** Give it its own section header (e.g. `# Transcode`) above the lock, and leave `# Switches` immediately above `switches` / `switches_lock`.

wrolpi/contexts.py
lrnselfreliance: Accepted as a known limitation (per project decision): a SIGKILL'd holder leaves the lock held until API restart; transcoding is not critical work. Documented at the lock's declaration in 03cec371 rather than adding owner-PID stealing or flock.

modules/videos/transcode.py
lrnselfreliance: Fixed in 03cec371: docstring trimmed to the polling rationale.

modules/videos/test/test_transcode.py
lrnselfreliance: Fixed in 03cec371: docstring softened to what the test measures (in-process wait/release); cross-process sharing is noted as provided by shared_ctx, not proven by the test.

modules/videos/transcode.py
lrnselfreliance: Fixed in 03cec371: the notify moved from enforce_codecs into _transcode_video_file, after the lock is acquired, so it fires when this file's work actually begins.

wrolpi/contexts.py
lrnselfreliance: Fixed in 03cec371: transcode_lock now has its own '# Transcode' section; '# Switches' sits directly above the switches state again.

PR #706, opened 2026-09-12
zim: recompute the outdated-Zims flag after startup and on Delete

lrnselfreliance:
## Summary

The Dashboard's "Outdated Zim Files" banner is driven by the persisted `outdated_zims` flag, while the modal behind its Delete button lists files from a live `GET /api/zim/outdated`. The two can disagree: the flag is restored from the DB on startup, and the startup recompute ran *before* the DB was up, so the restored (stale) value won. Once the outdated file had been removed by other means, the banner stayed up forever, the modal showed an endless placeholder with nothing under it, and Delete could never clear it because the endpoint only re-flagged when it actually deleted something.

## Invariants

- The `outdated_zims` flag always reflects the Zim directory on disk once startup has finished: the recompute runs after `db_up` and `import_all_db_configs`, i.e. after `init_flags` restored the persisted value.
- `DELETE /api/zim/outdated` always leaves the flag consistent with disk, whether or not it deleted anything.
- The modal always shows a fresh scan when opened, and states plainly when nothing was found.

## Changes

- `main.py`: move the startup `flag_outdated_zim_files()` call to after the DB-dependent config imports.
- `modules/zim/api.py`: always call `flag_outdated_zim_files()` after a delete.
- `app/src/components/Zim.js`, `app/src/hooks/customHooks.js`: the hook exposes its fetch; the modal re-fetches on open and renders "No outdated Zim files were found" for an empty list.

## Tests

- `test_delete_outdated_zims_clears_stale_flag`: a set flag with no outdated files on disk is cleared by the delete endpoint.
- Jest: the modal explains an empty result, and calls the API again when opened.
- Full `pytest -nauto`, full Jest suite (73 suites / 1245 tests) and `npm run build` pass.

## Verification

Deployed to a Raspberry Pi that exhibited the bug (flag `true`, live scan empty). After restart the flag cleared on its own once startup finished, and the Dashboard no longer shows the banner.


main.py
```diff
@@ -297,6 +291,14 @@ async def start_single_tasks(app: Sanic):
         from wrolpi.common import import_all_db_configs
         await import_all_db_configs()
 
+    # Recompute the outdated Zims flag from disk now that the DB is up and configs are imported.  This must run
+    # after `init_flags` restored the persisted value, otherwise a stale value (file was removed manually) wins.
+    from modules.zim.lib import flag_outdated_zim_files
+    try:
+        flag_outdated_zim_files()
```
lrnselfreliance, line 293: **[suggestion]** The comment claims this runs after `init_flags`, but `start_single_tasks` only waits for `flags.db_up`, then optionally awaits `import_all_db_configs()`, then calls `flag_outdated_zim_files()`. It never waits for `flags_initialized` or calls `flags.init_flags()` itself. `init_flags` runs in `perpetual_check_db_is_up_worker`, which can be claimed by a different Sanic worker than `start_single_tasks`. If that worker’s restore of `outdated_zims` lands after this recompute, the stale DB value wins again and sticks until another recompute. Placing the call after `import_all_db_configs()` also does not help correctness (`get_zim_directory()` only needs the already-imported WROLPi config) and prolongs the window where a just-restored stale flag can flash the Dashboard banner during startup.

**Suggestion:** Right after `db_up` is observed, call `flags.init_flags()` (idempotent) and then `flag_outdated_zim_files()` before the slow DB config import. That makes “disk wins after restore” an enforced local sequence instead of a cross-worker hope.

modules/zim/api.py
```diff
@@ -201,8 +201,10 @@ async def get_outdated_zims(_: Request):
 )
 async def delete_outdated_zims(_: Request):
     deleted_count = await lib.remove_outdated_zim_files()
```
lrnselfreliance: **[suggestion]** The supported way to clear a stale banner is now “open modal → see empty list → click Delete”, which always calls `lib.remove_outdated_zim_files()`. That helper activates `restart_kiwix_handler` even when `outdated` is empty (`refresh_sync` is a no-op on `[]`, but the restart switch is not). Clearing a stale flag therefore restarts Kiwix and briefly takes Zim browsing down for no file change.

**Suggestion:** Skip the Kiwix restart when `deleted_count == 0` inside `remove_outdated_zim_files`, or clear the flag without going through the delete/restart path when the live scan is already empty.

modules/zim/api.py
```diff
@@ -201,8 +201,10 @@ async def get_outdated_zims(_: Request):
...
+    # Always recompute: the persisted flag may be stale (file removed manually), and the user expects
+    # the banner to go away after clicking Delete.
+    lib.flag_outdated_zim_files()
```
lrnselfreliance: **[suggestion]** `flag_outdated_zim_files()` runs only if `remove_outdated_zim_files()` returns normally. If unlink/`refresh_sync` fails after some files were already removed, the handler errors out and the persisted flag is not recomputed, leaving banner/disk inconsistent—the invariant this PR is trying to restore.

**Suggestion:** Recompute in a `finally` (or equivalent) so the flag reflects disk whether or not the delete path fully succeeded; keep the event/response behavior as-is on failure.

main.py
```diff
@@ -297,6 +291,14 @@ async def start_single_tasks(app: Sanic):
         from wrolpi.common import import_all_db_configs
         await import_all_db_configs()
 
+    # Recompute the outdated Zims flag from disk now that the DB is up and configs are imported.  This must run
```
lrnselfreliance: **[suggestion]** The two-line comment restates the surrounding code and embeds the bug/design history (“after `init_flags` restored the persisted value… stale value wins”). Same pattern at `modules/zim/api.py:204`, `app/src/components/Zim.test.js:77`, and the new test docstring in `modules/zim/test/test_api.py`. Comments here should state the ordering/constraint briefly, not narrate why the change exists.

**Suggestion:** Replace with a short WHY constraint only, e.g. `# Disk scan must overwrite the DB-restored outdated_zims value`, and drop the architecture-history block comments/docstrings (or keep the behavioral assert message in the test body).

lrnselfreliance (commented):
## Summary

The PR correctly fixes the two user-visible failures: the startup recompute no longer runs before the DB (so `init_flags` cannot overwrite a fresh disk scan with a stale persisted value forever), and `DELETE /api/zim/outdated` always recomputes the flag so a stale banner can be cleared when nothing remains on disk. The frontend empty-state and refetch-on-open changes match that story, and the new API/Jest tests would fail on the old behavior. Dominant residual risks are that startup still does not explicitly synchronize with `init_flags` (and waits on an unrelated slow config import), and the new “click Delete to clear” empty path still goes through `remove_outdated_zim_files()`, which restarts Kiwix even when it deleted nothing.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 0

main.py
lrnselfreliance: Done in 2d4be111: `start_single_tasks` now calls `flags.init_flags()` (idempotent) and then `flag_outdated_zim_files()` immediately after observing `db_up`, before the DB config import.

modules/zim/api.py
lrnselfreliance: Done in 2d4be111: `remove_outdated_zim_files` skips `refresh_sync` and the Kiwix restart switch when `deleted_count == 0`. Added `test_remove_outdated_zim_files_no_restart_when_nothing_deleted`.

modules/zim/api.py
lrnselfreliance: Done in 2d4be111: the recompute now runs in a `finally`, so the flag matches disk even if the delete raises. Event and response behavior unchanged on failure.

main.py
lrnselfreliance: Done in 2d4be111: comments and docstrings trimmed to the constraint only in `main.py`, `modules/zim/api.py`, `Zim.js`, `Zim.test.js`, and the API test docstring.

PR #707, opened 2026-09-13
docs: browse walks the file_group date index instead of probing every doc

lrnselfreliance:
## Problem

The first unfiltered `/api/docs/search` after a reboot takes about a minute on a spinning disk, even after other pages have warmed the drive. Warm requests take 0.07s.

The ORM join `FileGroup JOIN Doc` is planned from `doc` (small covering index), then probes `file_group` by primary key for every doc to read the sort key, then sorts in a temp b-tree. Doc file_groups carry hundreds of MB of inline extracted text and are interleaved by id with every other file_group, so this is ~20k random 4 KB reads across a 10 GB file. Warming archives touches different pages, so it does not help.

## Fix

Same approach the archive browse already uses (see the comment in `modules/archive/lib.py`): for the unfiltered browse ordered by published date or id, fetch the page of ids with

```sql
SELECT fg.id FROM file_group fg CROSS JOIN doc d ON d.file_group_id = fg.id
ORDER BY fg.published_datetime DESC NULLS LAST, fg.id DESC LIMIT :limit OFFSET :offset
```

`CROSS JOIN` pins `file_group` as the outer table so SQLite walks `file_group_published_datetime_idx` and probes doc's unique covering index. Both sides are index-only and the scan stops at the page limit. FileGroups are then loaded by id and reordered. Filtered searches and the `size`/`title` orderings are unchanged.

Measured against a 10 GB production-shaped database with the page cache evicted:

| | Cold | Warm |
|---|---|---|
| Before | 51-64 s | 0.07 s |
| After (same rows) | 0.6 s | <0.01 s |

## Test

`test_search_docs_browse_walks_file_group_date_index` runs `EXPLAIN QUERY PLAN` on the browse SQL and asserts it uses the date index and doc's covering index, with no primary-key probe and no temp b-tree. It also checks NULLS LAST ordering, paging, and that a filtered search is unaffected. It fails before the change.

`pytest -nauto`: 1821 passed, 5 skipped.


modules/docs/lib.py
```diff
@@ -348,6 +348,28 @@ def _get_doc(session, file_group_id: int):
...
+# ORDER BY clauses for the unfiltered docs browse.  Only orderings that walk a `file_group` index
+# qualify; the rest keep the ORM query.
+DOC_BROWSE_ORDERS = {
+    'published_datetime': 'fg.published_datetime DESC NULLS LAST, fg.id DESC',
+    'id': 'fg.id DESC',
+}
+
+
+def _doc_browse_sql(order_by: str) -> Optional[str]:
+    """SQL for a page of FileGroup ids when browsing docs with no filters, or None when `order_by` has no fast path.
```
lrnselfreliance, line 360: **[suggestion]** `_doc_browse_sql`'s docstring restates the SQL mechanics and embeds architecture/performance history (~60s cold, 20k docs, 10 GB DB, "same as archive") rather than a short WHY. The nearby `DOC_BROWSE_ORDERS` comment already covers the constraint; this duplicates the archive browse comment style in a longer form.

**Suggestion:** Keep one concise sentence on why the outer table must be `file_group` (index-only probe of `doc`), and drop the benchmark narrative / cross-module history from the docstring.

lrnselfreliance (commented):
## Summary

This PR correctly ports the archive unfiltered-browse `CROSS JOIN` pattern to docs so the default published-date listing walks `file_group_published_datetime_idx` and probes `doc`'s unique covering index instead of materializing every large `file_group` row to sort. Fast-path gating matches the UI (`published_datetime` yes; `size`/`title` stay on the ORM; unknown/`rank`-without-search falls back to `id` like the ORM `else` branch), `NULLS LAST` + `fg.id DESC` match the existing docs ORM order, and OFFSET/LIMIT/total handling stay aligned with the generic path. Dominant residual risk is comment/test chattiness and EXPLAIN-string brittleness, not correctness; measured behavior and the new plan assertion look sound.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 1

modules/docs/lib.py
```diff
@@ -418,7 +440,17 @@ def _search_docs(search_str=None, author=None, subject=None, language=None, mime
         else:
             query = query.order_by(desc(FileGroup.id))
 
-        file_groups = query.offset(offset).limit(limit).all()
+        # Unfiltered browse: the ORM query above orders by a file_group column (or falls back to id) so the
```
lrnselfreliance: **[suggestion]** The call-site comment narrates the change ("ORM query above orders… so the index-walking SQL returns the same page") instead of stating a non-obvious constraint. A reader already sees the `unfiltered` / `DOC_BROWSE_ORDERS` gate.

**Suggestion:** Remove it, or replace with a one-liner on the non-obvious part only (e.g. why `size`/`title` must not fall through to the `id` fallback).

modules/docs/test/test_lib.py
```diff
@@ -334,3 +337,40 @@ async def test_search_docs_no_hint_when_no_query(test_session, test_directory):
     results, _ = _search_docs(mimetype='application/pdf')
     assert results and results[0]['id'] == fg.id
     assert 'section_hint' not in results[0]
+
+
+@pytest.mark.asyncio
+async def test_search_docs_browse_walks_file_group_date_index(test_session, test_directory, doc_factory):
+    """The unfiltered docs browse drives from file_group's date index and probes doc's covering index.
```
lrnselfreliance: **[suggestion]** The test docstring repeats the same design/performance history as `_doc_browse_sql` (60s cold / 20k docs / 10 GB). That belongs in the PR description, not in a regression test docstring.

**Suggestion:** Describe what the test locks in (date index + covering `doc` probe, no PK probe / temp b-tree; NULLS LAST + paging) and omit the cold-disk story.

modules/docs/test/test_lib.py
```diff
@@ -334,3 +337,40 @@ async def test_search_docs_no_hint_when_no_query(test_session, test_directory):
...
+    docs[0].file_group.published_datetime = None
+    docs[1].file_group.published_datetime = datetime(2020, 1, 1, tzinfo=timezone.utc)
+    docs[2].file_group.published_datetime = datetime(2021, 1, 1, tzinfo=timezone.utc)
+    test_session.commit()
+
+    sql = _doc_browse_sql('published_datetime')
+    plan = ' '.join(row[3] for row in test_session.execute(text(f'EXPLAIN QUERY PLAN {sql}'),
+                                                            dict(limit=20, offset=0)).fetchall())
+    assert 'file_group_published_datetime_idx' in plan, plan
+    assert 'sqlite_autoindex_doc_1' in plan, plan
```
lrnselfreliance: **[nit]** Asserting the exact autoindex name `sqlite_autoindex_doc_1` is brittle. It is correct today because alembic creates an unnamed `UniqueConstraint('file_group_id')` on `doc`, but renaming that constraint or adding another unique before it would change the autoindex name and fail the test without losing the covering-index property the change cares about.

**Suggestion:** Prefer a stabler plan signal (e.g. `COVERING INDEX` + `file_group_id=?` on `doc`, or `autoindex_doc`) alongside the existing negative checks for `INTEGER PRIMARY KEY` and `TEMP B-TREE`.

modules/docs/lib.py
lrnselfreliance: Done in 756fe677: docstring now states only why file_group must be the outer table; benchmark and archive history live in the PR body.

modules/docs/lib.py
lrnselfreliance: Done in 756fe677: replaced with one line on why size/title must not fall back to the id ordering.

modules/docs/test/test_lib.py
lrnselfreliance: Done in 756fe677: docstring now describes what the test locks in (index walk, covering probe, no PK probe or temp b-tree, NULLS LAST, paging).

modules/docs/test/test_lib.py
lrnselfreliance: Done in 756fe677: asserts 'SEARCH d USING COVERING INDEX' plus '(file_group_id=?)' instead of the autoindex name.

PR #708, opened 2026-09-14
Speed Test calculator: ping, download, and upload against the API

lrnselfreliance:
## Summary

An offline speed test under **Calculators › Tools & Reference**. It answers "is the hotspot fast enough?" and "how many people can stream at once?" with no third-party service.

* **Backend** `wrolpi/speedtest.py`, blueprint at `/api/speedtest`: `ping`, `download` (random bytes streamed for 0.1–15 s), `upload` (streamed body discarded, byte count returned, 8 MiB cap), `info` (client address + in-flight test count).
* **No lock, on purpose.** Concurrent tests are how a user sees how the link shares between streamers. The server only counts in-flight tests in a `shared_ctx` Value (created pre-fork so all workers agree) and the UI says how many others ran alongside.
* **Sanic `PROXIES_COUNT = 1`** so `request.remote_addr` is the browser's address from Caddy's `X-Forwarded-For`. Nothing else in the codebase read that field.
* **Caddy**: both Caddyfiles route `/api/speedtest/*` with `flush_interval -1`, like the SSE chat route.
* **Frontend**: `speedtest.js` is React-free logic (250 ms buckets, steady-state average dropping the first second, ping min/median/jitter, stream-capacity tiers, hotspot address check). `SpeedTestCalculator` runs 20 pings → 10 s download → 10 s upload with a live figure and rolling trace, then shows Mbps/MB·s⁻¹, ping verdict, client address with a **via WROLPi hotspot** badge (10.42.x.x and hotspot status connected), and a 480p→4K "streams at once" table. Upload uses sequential 4 MiB POSTs because Safari cannot stream a request body.
* **`Sparkline`**: new inline-SVG `ui` component, in `/theme-sample`, checked in all four themes. A lost sample is a gap, not a plunge.
* The page states it measures the network, not the drive that serves video.

Help doc and screenshot (`docs/modules/calculators/speed-test.md`) are written in the wrolpi-help repo, not yet committed there.

## Test plan

- [x] `pytest -nauto`: 1832 passed, 5 skipped (11 new in `wrolpi/test/test_speedtest.py`: clamping, chunking, each endpoint, 413 cap, forwarded client IP, counter released after completion and after a broken stream, counter in shared context)
- [x] Jest full suite: 1267 passed (36 new across `speedtest.test.js` and `SpeedTestCalculator.test.js`)
- [x] `npm run build` clean for new files
- [x] Cypress smoke `cypress/e2e/calculators/speedtest.cy.js` passes
- [x] Live run on the Docker dev stack: all three phases complete, no console errors, counter back to 0 afterward
- [x] Sparkline checked in light, dark, night, amber
- [ ] Hotspot QA on a Pi: confirm the hotspot badge with a real 10.42.x.x client (Docker reports the host gateway)

## Follow-up (separate PR)

Continuous "walk test" mode for hotspot range checks. v1 is designed so it needs no server change: chain 10 s downloads and reuse `Sparkline` and the `runDownload`/`runPing` helpers.


app/src/components/calculators/SpeedTestCalculator.js
```diff
@@ -0,0 +1,277 @@
...
+        controllerRef.current = controller;
+        const {signal} = controller;
+        setResults(emptyResults());
+        setError(null);
+        setActiveDuring(0);
+        setLive({mbps: 0, trace: [], progress: 0, rtts: []});
+
+        try {
+            try {
+                setInfo(await fetchInfo());
```
lrnselfreliance, line 92: **[bug]** `start()` awaits `fetchInfo()` before `setPhase('ping')`, so `running` stays false and the Start/Run again button remains clickable. A second click (easy while `fetchInfo` is slow on the hotspot this tool targets) creates another `AbortController`, overwrites `controllerRef`, and runs a second concurrent test; Cancel then only aborts the later run.

**Suggestion:** Synchronously enter a non-idle phase (or disable via a ref) before any `await`, and abort any existing controller when starting. e.g. set `phase` to `ping`/`starting` immediately after creating the controller, or ignore clicks while `controllerRef.current` is set.

app/src/components/calculators/speedtest.js
```diff
@@ -0,0 +1,315 @@
...
+ * POST the same random body over and over for `seconds`, tallying the bytes each request
+ * delivered.  Sequential requests rather than a streamed body because Safari cannot stream a
+ * request body.  Resolves to `{mbps, peak, bytes, samples}`.
+ */
+export async function runUpload({signal, onProgress, seconds = PHASE_SECONDS, body, deps} = {}) {
+    const {fetch, now} = {...defaultDeps(), ...deps};
+    const blob = body || randomBlob();
+    const buckets = new Buckets(BUCKET_MS, now);
+    const deadline = buckets.start + seconds * 1000;
+    while (now() < deadline && !signal?.aborted) {
```
lrnselfreliance: **[bug]** `runUpload` only checks the deadline between POSTs. Each request body is a fixed 4 MiB (`UPLOAD_CHUNK_BYTES`), so on a slow link one in-flight POST can run for tens of seconds past `PHASE_SECONDS` with no abort-at-deadline. That contradicts the UI copy of a 10 s upload and holds a Sanic worker for the whole transfer; the slow-hotspot case is exactly what this calculator is for.

**Suggestion:** Race each upload `fetch` against a remaining-time deadline (abort when `PHASE_SECONDS` elapses), and/or use much smaller sequential chunks so the loop can stop near the deadline. Mirror a server-side max upload duration (see Issue 3).

wrolpi/speedtest.py
```diff
@@ -0,0 +1,161 @@
...
+async def upload(request: Request):
+    declared = request.headers.get('content-length')
+    if declared and declared.isdigit() and int(declared) > MAX_UPLOAD_CHUNK:
+        raise UploadTooLarge()
+
+    _adjust_active(1)
+    try:
+        start = time.monotonic()
+        total = 0
+        while True:
```
lrnselfreliance: **[bug]** Download is hard-capped at `MAX_DOWNLOAD_SECONDS`, but upload is only size-capped (`MAX_UPLOAD_CHUNK`). A slow or paused body can hold a worker until 8 MiB eventually arrives (or the socket times out). The module comment says the cap exists so uploads cannot hold a worker; size alone does not achieve that.

**Suggestion:** Add a monotonic deadline (e.g. same 15 s ceiling as download, or slightly above the client phase) around the `request.stream.read()` loop and abort/413/408 when exceeded; optionally reject further reads after the deadline even if under 8 MiB.

wrolpi/test/test_speedtest.py
```diff
@@ -0,0 +1,146 @@
...
+async def test_download_invalid_seconds(async_client):
+    request, response = await async_client.get('/api/speedtest/download?seconds=abc')
+    assert response.status_code == HTTPStatus.BAD_REQUEST
+    request, response = await async_client.get('/api/speedtest/download?seconds=-1')
+    assert response.status_code == HTTPStatus.BAD_REQUEST
+    assert _active() == 0
+
+
+@pytest.mark.asyncio
+async def test_download_client_disconnect_releases_counter(async_client):
```
lrnselfreliance, line 86: **[suggestion]** Download client-disconnect releasing `speedtest_active` is tested; the analogous upload path (broken/cancelled stream after `_adjust_active(1)`) is not. Upload is the easier place for a stuck counter if an exception escapes outside the `try`/`finally` shape later.

**Suggestion:** Add a test that patches `request.stream.read` (or closes the client mid-body) and asserts `_active() == 0` afterward, including when `UploadTooLarge` is raised mid-stream.

app/src/components/ThemeSamplePage.js
```diff
@@ -567,6 +568,23 @@ export function ThemeSamplePage() {
             </Panel>
         </Section>
 
+        <Section label='Sparkline'>
+            <Panel>
+                {/*
```
lrnselfreliance: **[suggestion]** The new Sparkline gallery comment restates theme-token behavior and product rationale already visible in the component and speed-test UI (“null is a gap… needs nothing per theme”).

**Suggestion:** Drop or shorten to a single WHY line if anything non-obvious remains; the demo values already show the gap behavior.

wrolpi/speedtest.py
```diff
@@ -0,0 +1,161 @@
+"""Speed test endpoints: the browser measures its network path to this WROLPi.
```
lrnselfreliance: **[suggestion]** The module docstring narrates design history (no lock on purpose, network vs disk, UI wording) rather than a short module purpose. Same pattern in `Sparkline.tsx`’s component docstring.

**Suggestion:** Keep a brief purpose line; move intentional tradeoffs to the PR/plan doc if they are not needed to safely edit the file.

lrnselfreliance (commented):
## Summary

Solid end-to-end feature: bounded download streaming, Safari-safe sequential uploads, shared pre-fork active counter, Caddy `flush_interval` on both deploy Caddyfiles, and thorough unit/UI coverage for the happy path, cancel, and clamp/413/XFF cases. Auth posture matches the rest of the unauthenticated LAN API; `PROXIES_COUNT = 1` is low-risk because nothing else reads `remote_addr`. Dominant risks are a Start-button race before `setPhase('ping')` (especially on a slow hotspot) and upload phases that can far overrun the advertised 10 s because 4 MiB POSTs are not deadline-aborted and the server upload loop has no time bound.

## Issue counts by severity

- bugs: 3
- suggestions: 3
- nits: 1

wrolpi/speedtest.py
```diff
@@ -0,0 +1,161 @@
...
+def clamp_seconds(value: Optional[str]) -> float:
+    """Parse the requested download duration and keep it within the server's bounds."""
+    if value is None or value == '':
+        return DEFAULT_DOWNLOAD_SECONDS
+    try:
+        seconds = float(value)
+    except (TypeError, ValueError):
+        raise ValidationError(f'seconds must be a number, got {value!r}')
+    if seconds < 0 or seconds != seconds:  # negative or NaN
+        raise ValidationError('seconds must not be negative')
```
lrnselfreliance, line 69: **[nit]** `float('nan')` is rejected with the message `seconds must not be negative`, which is misleading for NaN.

**Suggestion:** Split the check or use a message like `seconds must be a non-negative number`.

app/src/components/calculators/SpeedTestCalculator.js
lrnselfreliance: Fixed in b5b31992. start() returns early when a controller already exists and enters a synchronous `starting` phase ("Connecting…") before the first await, so the button is already Cancel. Test added: the Start/Run again button is gone and the phase reads Connecting while fetchInfo is still pending, and only one run happens.

app/src/components/calculators/speedtest.js
lrnselfreliance: Fixed in b5b31992. runUpload() now races each POST against the phase deadline with its own AbortController: an in-flight request is aborted when PHASE_SECONDS elapses and only completed requests count. Bodies are 1 MiB instead of 4 MiB so a few-Mbps hotspot still completes several inside the phase. Test added with fake timers: a request that never completes is aborted at the deadline and the phase ends on time.

wrolpi/speedtest.py
lrnselfreliance: Fixed in b5b31992. drain_upload() enforces MAX_UPLOAD_SECONDS (15 s, same as download) around the read loop and raises UploadTooSlow (408) when exceeded; info reports the new cap. The constant's comment now says why size alone is not enough.

wrolpi/test/test_speedtest.py
lrnselfreliance: Added in b5b31992: three upload tests assert the counter is back to 0 after a mid-stream 413 (chunked body that only reveals its size while streaming), after the new 408, and after drain_upload blows up with a connection error.

app/src/components/ThemeSamplePage.js
lrnselfreliance: Trimmed in b5b31992 to one line about the gap behavior.

wrolpi/speedtest.py
lrnselfreliance: Trimmed in b5b31992. The module docstring is two lines of purpose; the Sparkline docstring is one. The no-lock rationale lives in the PR body and the plan doc.

wrolpi/speedtest.py
lrnselfreliance: Fixed in b5b31992: NaN is checked first with its own message, negative keeps the old one. Test covers both.

PR #709, opened 2026-09-14
Hotspot captive portal: welcome page with the WROLPi's address

lrnselfreliance:
## Summary

People who join the WROLPi hotspot constantly ask "what address do I open?".  This makes the phone answer it for them.

When a device joins the hotspot it probes a fixed plain-HTTP URL (`captive.apple.com/hotspot-detect.html`, `connectivitycheck.gstatic.com/generate_204`, ...) to check for Internet.  NetworkManager's shared-mode dnsmasq now resolves those probe hostnames to the hotspot's address, the probe lands on the Controller (port 80), and is redirected to a landing page:

> **This is your WROLPi.** The address of this WROLPi is **10.42.0.1**. [Open WROLPi] [Open the Controller]

`http://wrolpi/` also reaches the WROLPi from a hotspot client.

## How it works

* `controller/lib/captive_portal.py` renders a dnsmasq include (`address=/<probe host>/<ip>` lines) into `/etc/NetworkManager/dnsmasq-shared.d/`, which NetworkManager's dnsmasq already reads via `--conf-dir` when the Hotspot connection activates.
* `start_hotspot()` writes the include before `nmcli device wifi hotspot`, then reads the real address with `ip -j` and, only if NetworkManager picked a non-default subnet, rewrites the include and re-activates the connection once.
* New router `controller/api/captive_portal.py`: `GET /portal` (the page) and `GET|HEAD` on each probe path → `302 /portal`, all `Cache-Control: no-store`.  Bare paths on purpose: phones request fixed URLs.
* The page is plain HTTP and self-contained because the phone's sign-in sheet is a limited browser with no trusted CA.  It tells the user to open a real browser if the buttons do nothing.

## Invariants

* **Only probe hostnames are hijacked, never a wildcard.**  Shared mode forwards other DNS/traffic to the WROLPi's uplink and a wildcard would kill that.
* **Writing the include is best-effort.**  A failure logs a warning and the hotspot still starts.
* **The include exists only while the setting is on.**  repair.sh creates the directory but never copies a file into it, so a disabled setting survives upgrades.
* **`/portal` is 404 when the hotspot is down or in Docker.**  A page with the wrong address is worse than none.
* **Docker never redirects probes or writes the include.**

## Setting

`hotspot.captive_portal` in controller.yaml, default `true`, exposed on `/api/hotspot/settings` and as a checkbox in both Controller UIs.  Like the other hotspot settings it applies on the next hotspot start (dnsmasq only reads its include at activation), so saving while the hotspot is up returns `restart_required` and both UIs say "Restart the hotspot to apply".  `/api/hotspot/status` gains `ip`, `captive_portal`, `portal_url`; both UIs show the address while the hotspot is running.

## Tests

* `controller/test/test_captive_portal.py`: include rendering (every host, no wildcard), write/remove/best-effort, `ip -j` parsing, probe redirects (on/off/Docker), portal page (address, links, 404 cases).
* `test_admin.py`: include written before the hotspot call, removed when disabled, rewritten + re-activated on a non-default subnet, `restart_required` only while connected, status dict address.
* `ControllerPage.test.js`: checkbox round-trip, restart toast, address line.

Controller 924 passed · `pytest -nauto` 1835 passed · Jest 1272 passed · `npm run build` clean.

## Acknowledgement (added after on-device QA)

Phones only offer "Done" once a re-probe after a full-page navigation returns the OS's success reply. The welcome page never acknowledges a client (iOS fetches it in the background); **Continue** navigates to `/portal/continue`, which acknowledges the client IP, and later probes from it get the expected reply (Apple "Success" HTML, Android 204, Windows "Microsoft Connect Test", ...). Cleared on hotspot start. Probe hosts use `local=` as well as `address=` so HTTPS-type queries are never forwarded. See the QA comment below.

## Not in this PR

* Android and Windows on-device QA (iOS done).
* Help doc update lives in the wrolpi-help repo.


app/src/components/admin/ControllerPage.js
```diff
@@ -1482,6 +1487,28 @@ function HotspotSettingsForm() {
         fetchSettings();
     }, []);
 
+    // The address is only meaningful while the hotspot is up; it is optional information.
+    React.useEffect(() => {
```
lrnselfreliance, line 1491: **[bug]** `HotspotSettingsForm` fetches `getHotspotStatus()` once on mount (deps `[dockerized]` only). Starting or stopping the hotspot via the sibling `HotspotToggle` does not refresh this state, so the new "Hotspot address / Welcome page" line can keep showing an address after the hotspot is stopped, or stay hidden after it is started. The fallback UI correctly refreshes via `onStatus: showHotspotAddress` on every status poll; the React form does not.

**Suggestion:** Re-fetch when hotspot powered state changes (e.g. depend on `useHotspot().on`), or clear/refetch after the toggle’s start/stop completes. Mirror the fallback UI’s refresh-on-status behavior.

lrnselfreliance (commented):
## Summary

This PR adds a well-scoped hotspot captive portal: NetworkManager dnsmasq hijacks only specific OS probe hostnames (no wildcard), the Controller on :80 redirects those probe paths to a self-contained `/portal` page, and Docker/hotspot-down paths 404 as claimed. Caddy is HTTPS-only (`:443`) while the Controller binds `0.0.0.0:80`, so plain-HTTP probes correctly reach the new routes; Jinja2 autoescape covers SSID/IP XSS; settings, defaults, both UIs, and `repair.sh` (directory only) look consistent. Dominant risks are React status staleness after hotspot toggle, and the rare non-default-subnet path where `_fix_captive_portal_ip` can skip or silently fail to re-activate.

## Issue counts by severity

- bugs: 1
- suggestions: 4
- nits: 1

controller/lib/admin.py
```diff
@@ -515,6 +540,8 @@ def start_hotspot() -> dict:
             # WPA2 is what nmcli created the hotspot with anyway; not fatal.
             logger.warning("Failed to explicitly set WPA2 on hotspot: %s", error)
 
+        _fix_captive_portal_ip(device)
```
lrnselfreliance, line 547: **[suggestion]** `_fix_captive_portal_ip(device)` runs only on the success path after protocol modify/up. The earlier WPA3 failure path that successfully reverts to WPA2 returns `success: False` while leaving the hotspot up, and never reaches this call. In the rare case NetworkManager assigned a non-default shared subnet, the dnsmasq include can keep the wrong IP even though clients are connected.

**Suggestion:** Call `_fix_captive_portal_ip(device)` before those early returns when the hotspot is actually up (or share a single post-activation cleanup path for all “hotspot is running” exits).

controller/lib/admin.py
```diff
@@ -529,6 +556,27 @@ def start_hotspot() -> dict:
...
+    if not captive_portal_lib.is_captive_portal_enabled():
+        return
+    ip = captive_portal_lib.get_hotspot_ip(device)
+    if not ip or ip == captive_portal_lib.dnsmasq_config_ip():
+        return
+    logger.info("Hotspot address is %s, rewriting captive portal config and re-activating", ip)
+    if not captive_portal_lib.apply_dnsmasq_config(ip):
+        return
+    try:
+        subprocess.run(["nmcli", "connection", "up", "Hotspot"], capture_output=True, text=True, timeout=30)
```
lrnselfreliance: **[suggestion]** The corrective `nmcli connection up Hotspot` ignores a non-zero exit status; only exceptions are logged. If re-activation fails, dnsmasq keeps serving the previous include IP with no warning, which defeats the non-default-subnet fix.

**Suggestion:** Check `returncode` (and stderr) and log a warning on failure, same as the surrounding hotspot start paths.

controller/lib/captive_portal.py
```diff
@@ -0,0 +1,135 @@
+"""
```
lrnselfreliance, line 1: **[suggestion]** The module docstring is a long design/architecture narrative (probe flow, why no wildcard, Docker constraints) rather than a short why-oriented summary. Similar design-leaking commentary appears elsewhere in this change (e.g. the React “optional information” / “phones need this address” comments).

**Suggestion:** Trim to one or two sentences on responsibility; leave rationale to the plan/PR description. Keep only comments that encode non-obvious constraints (no wildcard, include-before-activate, best-effort write).

controller/test/test_admin.py
```diff
@@ -938,15 +983,68 @@ def mock_get_config(key, default=None):
...
+
+    def test_writes_captive_portal_config_before_creating_hotspot(self, reset_runtime_config):
+        """dnsmasq reads the include at activation, so it must exist before `nmcli ... hotspot`."""
+        include = captive_portal.INCLUDE_FILE
+        result, run = self._start_with_nmcli()
+        assert result["success"] is True
+        assert include.exists()
+        assert 'address=/captive.apple.com/10.42.0.1' in include.read_text()
+        # The address is only read after the hotspot exists; the include was written before that.
+        hotspot_call_index = next(i for i, c in enumerate(run.call_args_list) if 'hotspot' in c[0][0])
```
lrnselfreliance: **[suggestion]** `test_writes_captive_portal_config_before_creating_hotspot` claims to prove the include exists before `nmcli ... hotspot`, but the assertions only check that the file exists after the run and that the hotspot subprocess call precedes the later `ip` read. That does not observe whether the include was present at hotspot-create time (production code does write first; the test does not lock that in).

**Suggestion:** Assert include existence inside a `subprocess.run` side_effect/wrapper on the hotspot call (or record ordered events: `apply` then `hotspot`).

controller/api/schemas.py
```diff
@@ -156,6 +156,9 @@ class HotspotStatusResponse(BaseModel):
     reason: Optional[str] = Field(default=None, description="Reason if unavailable")
     ssid: Optional[str] = Field(default=None, description="Hotspot SSID when enabled")
     device: Optional[str] = Field(default=None, description="WiFi device name")
+    ip: Optional[str] = Field(default=None, description="Hotspot IPv4 address when enabled")
+    captive_portal: bool = Field(default=False, description="Whether the captive portal is enabled")
```
lrnselfreliance: **[nit]** `HotspotStatusResponse.captive_portal` defaults to `False`, while `HotspotSettingsResponse.captive_portal` defaults to `True` and `defaults.py` / `is_captive_portal_enabled()` also default true. Runtime responses always set the field, so this is OpenAPI/docs inconsistency rather than a functional bug.

**Suggestion:** Align the status schema default with the real default (`True`), or omit a misleading default and require the field.

app/src/components/admin/ControllerPage.js
lrnselfreliance: Fixed in 85c27b91: the form now takes `on` from `useHotspot()` and re-reads the Controller status whenever it flips, so the address line follows the toggle. Test added: `hotspot address is re-read when the hotspot is toggled`.

controller/lib/admin.py
lrnselfreliance: Fixed in 85c27b91: `_fix_captive_portal_ip(device)` now also runs on the WPA3-failed-but-reverted-to-WPA2 path, which is the only early return that leaves the hotspot up. Test: `test_wpa3_reverted_hotspot_still_gets_captive_portal_ip`.

controller/lib/admin.py
lrnselfreliance: Fixed in 85c27b91: a non-zero exit from the corrective `nmcli connection up` now logs a warning with nmcli's stderr, saying dnsmasq still serves the old address. Test: `test_reactivation_failure_is_logged`.

controller/lib/captive_portal.py
lrnselfreliance: Trimmed in 85c27b91: the module docstring is two sentences, the no-wildcard constraint lives on the `PROBE_HOSTS` comment where it is enforced, and the React narrative comments were removed. The rationale stays in the PR description.

controller/test/test_admin.py
lrnselfreliance: Fixed in 85c27b91: the test now wraps `subprocess.run` and reads the include at the moment of the `nmcli ... hotspot` call, asserting it already contains the address entries. Ordering is observed, not inferred.

controller/api/schemas.py
lrnselfreliance: Fixed in 85c27b91: `HotspotStatusResponse.captive_portal` defaults to `True` like the settings schema and `defaults.py`.

lrnselfreliance:
## On-device QA (iPhone, 10.0.0.8 with an eth0 uplink)

Result: welcome page → **Continue** → "You are connected" page, links open in Safari, the sheet offers **Done**, and the hotspot is freely usable afterwards.

Two bugs only a real phone could show, both fixed on the branch:

1. **HTTPS-type DNS queries were forwarded** (66151df6). `address=` only synthesizes A/AAAA. iOS also asks for the HTTPS record type; dnsmasq forwarded it upstream, Apple's answer aliased to `captive.g.aaplimg.com`, and the sheet got the genuine "Success" from Apple. Every probe host now also gets `local=/host/` so no record type is forwarded, and the CDN alias is a probe host.
2. **Acknowledging on view broke the sheet** (f182e64e → c72a7f7b). iOS fetches the portal page in the background before opening the sheet, and only re-probes (and offers Done) after a full-page navigation ([WBA captive portal behavior reference](https://captivebehavior.wballiance.com/)). The portal is now two steps: the welcome page never acknowledges; **Continue** is a real navigation to `/portal/continue`, the only place a client IP is acknowledged. Later probes from that client get the reply its OS expects (Apple "Success" HTML, Android 204, Windows "Microsoft Connect Test", Firefox, GNOME). The set clears on hotspot start.

Dropped RFC 8910 / DHCP option 114 as a follow-up: iOS only honors it on reconnect and calls the RFC 8908 API over HTTPS, which needs a trusted certificate.

Still to QA on a device: Android, Windows.

lrnselfreliance (commented):
## Summary

Follow-up review of commits after `6bea096` (`85c27b9` … `c72a7f7`):

The four new commits look correct and complete: the address-review commit genuinely fixes all six prior findings, `local=`+`address=` matches dnsmasq 2.86+ guidance for HTTPS/AAAA non-forwarding, and Continue-gated acknowledgement with OS-specific success bodies is coherent and well tested. Dominant residual risks are soft: IP-keyed acknowledgement can skip the portal if DHCP reuses an address during a long hotspot session, and a couple of comments still describe the pre-Continue acknowledge-on-view behavior. No correctness bugs found in the new probe/Continue path, XSS stays covered by Jinja autoescape, and the Controller’s single-worker uvicorn makes the in-process set safe.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 0

controller/lib/captive_portal.py
```diff
@@ -0,0 +1,158 @@
...
+    'clients3.google.com',  # Android (older)
+    'www.msftconnecttest.com',  # Windows 10+
+    'www.msftncsi.com',  # Windows 7/8
+    'detectportal.firefox.com',  # Firefox
+    'connectivity-check.ubuntu.com',  # Ubuntu / NetworkManager (probes `/`, served by the dashboard)
+    'nmcheck.gnome.org',  # GNOME
+)
+
+# Paths those probes request, and the reply each OS takes as "online".  The Controller redirects a
+# probe to the portal until the client has loaded the portal page, then returns the expected
```
lrnselfreliance: **[suggestion]** The `PROBE_SUCCESS_RESPONSES` block comment still says probes succeed once “the client has loaded the portal page.” After the Continue commit, viewing `/portal` does not acknowledge; only `/portal/continue` does. The same stale wording remains in `probe()`’s docstring.

**Suggestion:** Update both comments to say the client must tap Continue (or hit `PORTAL_CONTINUE_PATH`), matching `test_viewing_the_portal_does_not_acknowledge`.

controller/api/captive_portal.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    response = _render_portal(request, connected=True)  # 404s first if the hotspot is down
+    acknowledge_client(request.client.host if request.client else None)
+    return response
+
+
+async def probe(request: Request):
+    """
+    A connectivity probe.  Anything but the expected reply makes the phone show the portal, so
+    redirect until this client has loaded the portal page, then give the reply it expects.
```
lrnselfreliance: **[suggestion]** `probe()`’s docstring says “redirect until this client has loaded the portal page,” which is the acknowledge-on-view behavior that commit `c72a7f7` intentionally removed. Future readers may “fix” Continue back to view based on this.

**Suggestion:** Rephrase to “until this client has tapped Continue on the portal.”

controller/lib/captive_portal.py
```diff
@@ -0,0 +1,158 @@
...
+            if line.startswith('address=/'):
+                return line.rsplit('/', 1)[-1].strip() or None
+    except OSError:
+        pass
+    return None
+
+
+# Clients (by hotspot IP) which tapped Continue on the portal page.  Their later probes get the
+# success reply.  Reset whenever the hotspot starts so a new session shows the page again.
+_acknowledged_clients: set[str] = set()
```
lrnselfreliance, line 143: **[suggestion]** Acknowledgements are keyed only by hotspot client IP and cleared only on `start_hotspot()`. That is fine for NAT (clients are on the LAN with distinct DHCP leases) and for a single uvicorn worker, but a later joiner that inherits a previously-acknowledged lease can get probe success immediately and never see the welcome page for the rest of that hotspot session.

**Suggestion:** Acceptable for a personal hotspot; if it bites in practice, expire entries (TTL ≈ DHCP lease) or clear on `stop_hotspot()` as well as start.

controller/lib/captive_portal.py
```diff
@@ -0,0 +1,158 @@
...
+# only when the Hotspot connection is activated.
+DNSMASQ_SHARED_DIR = Path('/etc/NetworkManager/dnsmasq-shared.d')
+INCLUDE_FILE = DNSMASQ_SHARED_DIR / '50-wrolpi-captive-portal.conf'
+
+# NetworkManager's default shared-mode subnet.  The real address is read once the hotspot is up.
+DEFAULT_HOTSPOT_IP = '10.42.0.1'
+
+PORTAL_PATH = '/portal'
+# A full-page navigation the user makes from the portal.  Phones only re-probe (and offer
+# "Done") after a navigation, so the client is acknowledged here, never on viewing the page.
```
lrnselfreliance: **[suggestion]** The `PORTAL_CONTINUE_PATH` comment narrates the old design (“never on viewing the page”) more than it states why the path exists. Same flavor as the Jinja `{# A real navigation... #}` note in `portal.html`.

**Suggestion:** Keep a one-liner WHY (“phones only re-probe after a full navigation”) and drop the view-vs-Continue history.

controller/lib/captive_portal.py
lrnselfreliance: Fixed in 94c87fae: the block comment now says probes succeed once the client taps Continue.

controller/api/captive_portal.py
lrnselfreliance: Fixed in 94c87fae: the docstring reads "redirect until this client has tapped Continue on the portal".

controller/lib/captive_portal.py
lrnselfreliance: Took the cheap mitigation in 94c87fae: acknowledgements are now cleared on `stop_hotspot()` as well as start, so a reused lease in a later session sees the page again. Test: `test_stop_forgets_acknowledged_clients`. Left a TTL out; within one session a reused lease means the earlier device left, which on this hotspot is rare enough not to justify the timer.

controller/lib/captive_portal.py
lrnselfreliance: Trimmed in 94c87fae: the constant's comment is the one-line why (phones only re-probe after a full-page navigation), and the Jinja note says the same without the history.

PR #711, opened 2026-09-15
Jobs queue, video Edit dropdown: transcode/remux, edit title & description

lrnselfreliance:
## Summary

Adds an in-memory background **Job** queue and uses it for a per-video **Transcode** action.  The video page's action row becomes Download / **Edit ▾** / Add to Playlist, where the Edit dropdown holds Edit, Refresh, Transcode and Delete.

Motivation: a very long single-file MP4 played badly on a phone, and there was no way to re-encode or remux an already-downloaded video from the UI.  Codec preferences (#701) only apply at download time.

## Jobs (`wrolpi/jobs.py`)

- Functions register by name (`@register_job('name')`); callers enqueue by name with JSON kwargs.  Names rather than callables because the call crosses Sanic worker processes.
- State lives in `shared_ctx` (`jobs`, `jobs_queue`, `jobs_lock`).  One consumer loop (`job_worker`, a `perpetual_signal`) runs in the perpetual-tasks owner, so Jobs are strictly FIFO and one at a time.
- Output capture: a contextvar marks the running Job and a root-logger handler copies its log records; `JobContext.run_command` tees subprocess stdout live and the stderr tail on exit.  200-line ring buffer, flushed to shared state at most twice a second.
- Progress (`set_progress`), result, error, timestamps.  **Cancel**: pending Jobs are cancelled immediately; a running Job's handler task is cancelled and its subprocess killed.
- Not persisted; an API restart forgets them (by design).
- API: `GET /api/jobs`, `GET /api/jobs/<id>`, `DELETE /api/jobs/<id>` (cancel; allowed in WROL Mode since it only stops work).

Download-time transcodes keep their existing `transcode_lock`; moving them onto the queue is a later change.

## Transcode

- `POST /api/videos/<file_group_id>/transcode` with `video_codec`, `audio_codec`, `container` (validated against the existing target maps) queues a `transcode_video` Job.
- ffmpeg runs with `-nostats -progress pipe:1`; `out_time_us` is parsed into percent progress.
- No codec for either stream is a **remux**: both streams copied into the chosen container with `+faststart`.
- After ffmpeg, the **existing FileGroup is pointed at the new file before the refresh** (files, primary_path, mimetype, `indexed=False`, `ffprobe_json=None`).  Without this a suffix change made the refresh delete the FileGroup and create a new one, losing tags and history.
- The modal pre-fills codecs from the user's preferences (same first-producible rule as the downloader), starts the container at the file's own, shows the current codecs, then polls the Job with a progress bar, status, log tail and Cancel.

## Audio

- `video_codec: "none"` removes the video stream; the result is an audio file.  Audio-only output uses an audio container (`m4a`/`ogg`/`mp3`), checked against the codec it will hold (copied or target); a video container asked for audio output is replaced by the codec's natural one.
- Audio files can be transcoded too (audio codec + audio container); a video target on one is a 400.  `-movflags +faststart` only for mp4/m4a.  Verification knows the audio-only shape and ffprobe's container names (mkv reports `matroska`).
- The modal offers "Remove video (audio only)", switches to audio containers and follows the audio codec, reads **Extract Audio**, and opens for audio files without a video codec select.

## Fragmented mp4 (long videos)

A whole-file mp4 index grows with the frame count; an 11-hour file had a 40 MB `moov`, which a phone must fetch and parse before the first frame, and every seek walks it.  The **Fragmented** toggle (mp4/m4a only) remuxes with `+frag_keyframe+empty_moov+default_base_moof+global_sidx` and 10 s fragments: `moov` 40 MB → 1.4 KB plus a 2 × 120 KB seek index, duration/codecs unchanged, 25 s for 1.1 GB.  Tested on an iPhone: playback starts and seeks quickly.  HLS was considered and rejected for now (segment directories do not fit the FileGroup model and need hls.js outside Safari).

## Output verification

Before the original is replaced: the output is probed and compared with the source (container, streams present, codec per stream, duration within max(1s, 1%)), and its last 30 seconds are decoded to a null sink.  The tail decode is now the download-time completeness check too (`ffmpeg_video_complete`, one implementation, sync + async), replacing the old single-frame-near-the-end check; it is stricter for downloads as well.

## Edit

- `PUT /api/videos/<file_group_id>` with optional `title` / `description`; only the given fields change, an empty description clears it.  Edits go into the info json's **`wrolpi` section** (`custom_title`, `custom_description`, created if the video has no info json), never over yt-dlp's keys, so the original title and description stay in the file.  The modeler prefers the custom values, so a refresh keeps the edit.  Both re-download paths carry the `wrolpi` section over (the metadata-only refresh merges it into the fresh info json; a full re-download snapshots it before yt-dlp overwrites the file and restores it in `finalize_download`).

## WROL Mode

Routes return 403 (already), and now also: a transcode Job queued before WROL Mode was enabled fails before ffmpeg starts, `update_video` carries `wrol_mode_check`, and every dropdown item plus the modals' primary buttons are disabled.  Written test-first.

## Other

- The dropdown-button pattern was added to `/theme-sample` (Buttons section).
- Embedded subtitle streams are dropped by the transcode mapping (`0:V:0` + `0:a:0?`), same as the download-time transcode from #701.

## Testing

- `pytest -nauto`: full suite green (new: `wrolpi/test/test_jobs.py`, `modules/videos/test/test_transcode_job.py`, `modules/videos/test/test_video_edit.py`).
- Jest: full suite green (new: `VideoEditMenu.test.js`); `npm run build` clean.
- Live on the Docker dev stack: real transcode (h264/aac), remux mp4→mkv→mp4, extract audio mp4→mp3, audio transcode mp3→m4a, title and description edits, Job log/progress/events; FileGroup id preserved throughout.  File restored from backup afterwards.
- Cypress not run: it targets the running dev stack, not the branch under test.  Not yet QA'd on a Raspberry Pi.


lrnselfreliance (commented):
## Summary

This PR adds a solid in-memory Jobs queue (shared_ctx + single perpetual owner consumer), wires video transcode/remux through it with progress/log/cancel, and folds Refresh/Transcode/Delete/Edit into an Edit dropdown with WROL Mode gating. The core design (register-by-name, FileGroup retarget before refresh, info.json as source of truth for edits) looks correct and is well covered by backend and frontend tests. Dominant risks are a PENDING→RUNNING cancel race in the job worker, the Transcode modal’s Close path abandoning completion polling (so the video page can keep a stale path after an extension-changing remux), and a crash window where the old media file is deleted before the FileGroup row is retargeted.

## Issue counts by severity

- bugs: 4
- suggestions: 3
- nits: 1

wrolpi/jobs.py
```diff
@@ -0,0 +1,414 @@
...
+        json.dumps(value)
+        return value
+    except (TypeError, ValueError):
+        return str(value)
+
+
+async def _run_job(job_id: str):
+    """Run one Job to completion (or failure/cancellation) in this process."""
+    record = _jobs().get(job_id)
+    if not record or record['status'] != PENDING:
```
lrnselfreliance: **[bug]** `_run_job` reads the job record and checks `status == PENDING` without holding `jobs_lock`, then later calls `_update_job(..., status=RUNNING)` which unconditionally overwrites the record. If `cancel_job` marks a pending job `CANCELLED` in that window, the worker still transitions it to `RUNNING` and executes it. Cancel of a not-yet-started job is therefore racy.

**Suggestion:** Claim the job under the lock in one step: if status is still `PENDING`, write `RUNNING` (and pid/started_at) before releasing; otherwise return. Do not promote a record that is no longer pending.

app/src/components/VideoEditMenu.js
```diff
@@ -0,0 +1,471 @@
...
+                        {sameContainer
+                            ? 'Both streams are kept: the file is rewritten as-is with fast start (moves the index to the front for quicker playback start).'
+                            : `Both streams are kept: only the container changes to ${container} (a quick remux, no quality loss).`}
+                    </Text>}
+                </>}
+                {jobId && <JobProgress jobId={jobId} onFinished={handleFinished}/>}
+            </Stack>
+        </Modal.Content>
+        <Modal.Actions>
+            <Button role='cancel' onClick={onClose}>{jobId ? 'Close' : 'Cancel'}</Button>
```
lrnselfreliance: **[bug]** While a Job is running the modal offers Close, which only sets `open=false`. Mantine Modal defaults to `keepMounted: false`, so `JobProgress` unmounts, polling stops, and `onFinished` / `onComplete` (`fetchVideo`) never run. After a container/extension change the page keeps the old `video_path`/codecs until a manual reload, even though the server job succeeded (and may have deleted the old file). Reopening also resets `jobId` and allows queueing a second transcode against a path the first job may already have replaced.

**Suggestion:** Keep completion watching outside the modal (e.g. poll at `VideoEditMenu`/`VideoPage` while `jobId` is set), or use `keepMounted` / render `JobProgress` outside `Modal`, and disable starting another transcode for the same file until the active job finishes. Still call `onComplete` when the job completes even if the modal is closed.

app/src/components/VideoEditMenu.js
```diff
@@ -0,0 +1,471 @@
...
+        finishedRef.current = false;
+        let cancelled = false;
+
+        const poll = async () => {
+            let latest;
+            try {
+                latest = await getJob(jobId);
+            } catch (e) {
+                console.error(e);
+                return;
```
lrnselfreliance: **[bug]** On any `getJob` failure the poll loop `return`s without scheduling another attempt. A single transient network/API error leaves the UI stuck on the last frame (or forever on “Starting...” if the first poll fails) with no further updates or Cancel affordance recovery.

**Suggestion:** Log the error, then `setTimeout(poll, JOB_POLL_MS)` again (optionally with backoff), and only stop when unmounted or a finished status is observed.

modules/videos/transcode.py
```diff
@@ -207,3 +247,61 @@ async def _transcode_video_file(video_path: pathlib.Path, target_vcodec: Optiona
...
+    validate_transcode_request(video_codec, audio_codec, container)
+
+    with get_db_session() as session:
+        video = Video.find_by_file_group_id(session, file_group_id)
+        video_path = video.video_path
+        duration = video.file_group.length
+        if not video_path or not video_path.is_file():
+            raise RuntimeError(f'Video file does not exist: {video_path}')
+
+    final_path = await transcode_video_file(video_path, video_codec, audio_codec, container, duration=duration)
```
lrnselfreliance: **[bug]** `transcode_video_file` renames the temp output and, when the extension changes, deletes the original before `transcode_video_job` retargets `FileGroup.primary_path` / `files` (lines 289–302). A crash or power loss in that window leaves the DB pointing at a deleted path while the new file exists unlinked from the Video/FileGroup (tags/history at risk on a later refresh). Concurrent readers on the open video page can also 404 against the old path during the same gap.

**Suggestion:** After a successful rename to `final_path`, update the FileGroup (and commit) before unlinking the old path when `final_path != video_path`. Prefer deleting the old file only once the DB points at the new one; keep the careful fsync/rename behavior for the new file.

wrolpi/jobs.py
```diff
@@ -0,0 +1,414 @@
...
+        result=None,
+        log=[],
+        pid=None,
+        cancel_requested=False,
+    )
+    with _lock():
+        _jobs()[job_id] = record
+        _prune_finished_jobs()
+    _queue().put_nowait(job_id)
+    logger.info(f'enqueue_job: {job_id} ({description or name})')
```
lrnselfreliance, line 143: **[suggestion]** The jobs queue and `jobs` dict accept unlimited pending entries; only finished jobs are pruned (`JOB_HISTORY_LIMIT`) on enqueue. A runaway client (or UI bug that re-queues) can grow shared memory without bound. Cancelled pending ids also remain in the `multiprocessing.Queue` until drained as no-ops.

**Suggestion:** Enforce a max pending/queued count in `enqueue_job` (reject or coalesce), and/or skip queueing ids already cancelled; document the limit next to `JOB_HISTORY_LIMIT`.

wrolpi/jobs.py
```diff
@@ -0,0 +1,414 @@
...
+        """`wrolpi.cmd.run_command` whose stdout lines (live) and stderr tail (on exit) are
+        captured into this Job's log.  `stdout_callback` still receives each stdout line."""
+        from wrolpi.cmd import run_command
+
+        def _callback(line: str):
+            self.log(line)
+            if stdout_callback:
+                stdout_callback(line)
+
+        result = await run_command(cmd, stdout_callback=_callback, **kwargs)
```
lrnselfreliance, line 267: **[suggestion]** `JobContext.run_command` does not pass `start_new_session=True` into `wrolpi.cmd.run_command`. Cancel relies on `task.cancel()` → `CancelledError` → `proc.kill()` on the ffmpeg pid only. That is usually enough for ffmpeg, but it does not kill a process group the way other long-lived tools in this codebase do, and `run_command` swallows `CancelledError` (returning `cancelled=True`) so handlers must not assume the outer task stays cancelled.

**Suggestion:** Pass `start_new_session=True` for job subprocesses (or at least for ffmpeg), and/or have `JobContext.run_command` re-raise when `result.cancelled` so cancel semantics stay consistent for handlers that check return codes.

app/src/components/ThemeSamplePage.js
```diff
@@ -382,6 +383,25 @@ export function ThemeSamplePage() {
                         </div>)}
                     </Row>
                 </div>
+                <div style={{marginTop: 12}}>
+                    <Header as='h5'>A dropdown of actions</Header>
```
lrnselfreliance, line 387: **[suggestion]** The new sample comment embeds product/design rationale (“trailing chevron is what says menu…”, “disabled item stays listed so the user learns it exists”) rather than a short non-obvious constraint. Similar design-narration comments appear in the new `VideoEditMenu` / jobs module docs.

**Suggestion:** Trim to a one-line note if anything (e.g. that this mirrors the video Edit control), or delete; keep comments for non-obvious WHY only.

app/src/api.js
```diff
@@ -236,6 +236,58 @@ export async function getVideo(fileGroupId) {
...
+        const message = await getErrorMessage(response, 'Unable to queue the transcode.  See server logs.');
+        toast({type: 'error', title: 'Failed to transcode', description: message, time: 5000});
+        throw new Error(message);
+    }
+    const data = await response.json();
+    return data['job_id'];
+}
+
+export async function getJobs() {
+    const response = await apiGet(`${API_URI}/jobs`);
```
lrnselfreliance, line 266: **[nit]** `getJobs()` does not check `response.ok` (unlike `getJob` / `cancelJob` / `transcodeVideo`), so a non-2xx body is parsed as success. Nothing in the PR UI calls it yet, but the helper is inconsistent and easy to misuse later.

**Suggestion:** Mirror `getJob`: if `!response.ok`, throw with `getErrorMessage`.

wrolpi/jobs.py
lrnselfreliance: Fixed in 42b17ea5: `_claim_job` moves PENDING→RUNNING (with pid/started_at) in one step under `jobs_lock` and returns None when the record is no longer pending; `_run_job` runs only a claimed record. Test: `test_jobs_claim_is_atomic`.

app/src/components/VideoEditMenu.js
lrnselfreliance: Fixed in bd6b0012: the Job id and polling (`useJob`) now live in `VideoEditMenu`, so Close only hides the modal; `onTranscodeComplete` (→ `fetchVideo`) still fires. While a Job runs the item reads Transcoding..., reopening shows the progress instead of the form (no second queue), and Delete is disabled. Tests: 'closing the modal keeps watching the Job…', 'while a Job runs, Transcode shows its progress and Delete waits'.

app/src/components/VideoEditMenu.js
lrnselfreliance: Fixed in bd6b0012: a failed `getJob` is logged and the next poll is scheduled; polling stops only on unmount, a new job id, or a finished status. Test: 'a failed request is retried instead of ending the polling'.

modules/videos/transcode.py
lrnselfreliance: Fixed in 42b17ea5: `transcode_video_file(keep_original=True)` renames the output into place but leaves the original; the Job then retargets the FileGroup (committed) and only afterwards unlinks the old path. The downloader path is unchanged. Test: `test_transcode_video_job_retargets_before_deleting` asserts the old file still exists when the DB is updated.

wrolpi/jobs.py
lrnselfreliance: Fixed in 42b17ea5: `enqueue_job` refuses past `JOB_PENDING_LIMIT` (100) pending Jobs, documented next to `JOB_HISTORY_LIMIT`. Cancelled ids already in the multiprocessing queue are drained as no-ops (the queue holds ids only). Test: `test_jobs_pending_limit`.

wrolpi/jobs.py
lrnselfreliance: Fixed in 42b17ea5: `JobContext.run_command` defaults `start_new_session=True` and raises `CancelledError` when the result reports `cancelled`, so handlers cannot mistake a killed command for a normal exit. Test: `test_jobs_run_command_cancel_semantics`.

app/src/components/ThemeSamplePage.js
lrnselfreliance: Trimmed in bd6b0012: the gallery comment is now one line, and the narration in the VideoEditMenu docs was cut back to the non-obvious constraints (why the Job is owned by the menu, the jsdom menu quirk in tests).

app/src/api.js
lrnselfreliance: Fixed in bd6b0012: `getJobs` throws via `getErrorMessage` on a non-2xx response, like `getJob`.

app/src/components/VideoEditMenu.js
```diff
@@ -0,0 +1,573 @@
...
+    const wrolModeEnabled = useWROLMode();
+    const [editOpen, setEditOpen] = useState(false);
+    const [transcodeOpen, setTranscodeOpen] = useState(false);
+    const [refreshing, setRefreshing] = useState(false);
+    const [deleteOpen, setDeleteOpen] = useState(false);
+    const [deleting, setDeleting] = useState(false);
+    // `initialJobId`: resume watching a Job already queued for this video (tests use it too).
+    const [jobId, setJobId] = useState(initialJobId);
+
+    const handleJobFinished = (finished) => {
```
lrnselfreliance, line 476: **[bug]** `handleJobFinished` only toasts on `complete`. Closing the modal while a Job runs is the new intended UX, but if that Job then `failed` (or `cancelled` from elsewhere), the handler clears `jobId` when `!transcodeOpen` and shows nothing. The user who closed the modal never learns the original was kept because ffmpeg/verify failed.

**Suggestion:** Toast on `failed` (and optionally `cancelled`) with `finished.error` / status before clearing `jobId`, matching the success path.

lrnselfreliance (commented):
## Summary

Follow-up review of commits after `d591774` (`f2d15c2` … `ec53418`):

The prior eight review findings are addressed: atomic job claim, pending cap, sessioned subprocesses with CancelledError on kill, retarget-before-delete via `keep_original`, `useJob` owned by `VideoEditMenu` (survives Close and retries poll errors), `getJobs` checks `ok`, and the ThemeSamplePage narration comment is trimmed. Output verification (probe + duration window + tail decode) correctly refuses bad output and leaves the original; audio-only / Remove video (`none`), container coercion, FileGroup mimetype retarget, player audio UI, and WROL gating look sound. Two real gaps remain from the fix commits: silent failure toast when the modal is closed, and cancel leaving `.transcode.*` temps because `CancelledError` bypasses `except Exception`.

## Issue counts by severity

- bugs: 2
- suggestions: 3
- nits: 1

modules/videos/transcode.py
```diff
@@ -158,27 +333,55 @@ async def _transcode_video_file(video_path: pathlib.Path, target_vcodec: Optiona
...
                 f'ffmpeg exited with {result.return_code} while transcoding {video_path}:'
                 f'\n{result.stderr.decode()[-2000:]}')
+
+        # The original is only replaced by a complete output: compare probes, then prove the tail
+        # decodes.  A failure here leaves the original untouched.
+        logger.info(f'Verifying transcode output {tmp_path}')
+        output_probe = await probe_for_verify(tmp_path)
+        verify_transcode_output(source_probe, output_probe, target_vcodec, target_acodec, container)
+        await _verify_tail_decodes(tmp_path, runner)
     except Exception:
```
lrnselfreliance: **[bug]** Cleanup of the temporary output is only in `except Exception`. `JobContext.run_command` now raises `asyncio.CancelledError` when the command was killed (`wrolpi/jobs.py:276`), and on Python 3.9 that is a `BaseException`, not an `Exception`. Cancel during encode or during the new tail-decode verify therefore leaves `*.transcode.{container}` beside the original. Before the CancelledError change, a cancelled command typically returned a non-zero `CommandResult` and hit this cleanup path. A partial `.transcode.mp4` can also be picked up by a later file refresh.

**Suggestion:** Unlink `tmp_path` on any exit before a successful rename (e.g. `except BaseException:` or a `success` flag + `finally`), so cancel and verify failure both remove the temp.

app/src/components/VideoEditMenu.js
```diff
@@ -0,0 +1,573 @@
...
+        }
+        let stopped = false;
+        let timer = null;
+
+        const poll = async () => {
+            let latest = null;
+            try {
+                latest = await getJob(jobId);
+            } catch (e) {
+                console.error(e);
```
lrnselfreliance, line 105: **[suggestion]** Failed `getJob` polls are retried forever. That fixes transient errors, but a permanent miss (API restart wipes in-memory Jobs; 404) leaves `jobId` set and `job` null, so `transcoding` stays true (`!!jobId && !isJobFinished(job)`), Delete stays disabled, and the menu reads "Transcoding..." until navigation remounts the menu.

**Suggestion:** After a 404 / N consecutive failures, clear `jobId` (or mark the Job missing) and toast so the UI unlocks.

app/src/components/VideoEditMenu.js
```diff
@@ -0,0 +1,573 @@
...
+
+const statusKind = (status) => {
+    if (status === 'complete') return 'complete';
+    if (status === 'running') return 'active';
+    if (status === 'failed' || status === 'cancelled') return 'failed';
+    return 'pending';
+}
+
+/**
+ * Poll a Job until it finishes.  Lives in whichever component must outlive the UI showing the Job:
```
lrnselfreliance: **[suggestion]** Several new comments embed architecture / change rationale rather than a short WHY (`useJob` "Lives in whichever component…", TranscodeModal "owned by the parent", VideoEditMenu "Owns the active transcode Job…", and `showJob` "until the parent forgets it"). Same class of comment flagged earlier on ThemeSamplePage.

**Suggestion:** Keep one-line behavioral contracts; drop the ownership/history narration (or move it to the PR description).

wrolpi/jobs.py
```diff
@@ -0,0 +1,441 @@
...
+            logger.info(f'cancel_job: requested cancel of running {job_id}')
+        _jobs()[job_id] = record
+    return record
+
+
+def _claim_job(job_id: str) -> Optional[dict]:
+    """Move a Job from PENDING to RUNNING in one step under the lock; returns the running record,
+    or None when the Job is no longer pending (cancelled while queued, or already claimed).
+
+    Reading the status and writing RUNNING separately let a cancel land in between and be
```
lrnselfreliance: **[suggestion]** `_claim_job`'s docstring second paragraph narrates the old PENDING→RUNNING race ("Reading the status and writing RUNNING separately…") instead of stating the current invariant.

**Suggestion:** Document that claim is atomic under the lock and returns None when not pending; drop the history of the bug.

app/src/components/VideoEditMenu.js
```diff
@@ -0,0 +1,573 @@
...
+    // A queued or finished Job replaces the form until the parent forgets it.
+    const showJob = !!jobId;
+
+    let hint = null;
+    if (defaultsLoaded && removeVideo) {
+        hint = `The video stream is removed; the result is an audio file (.${container}).  `
+            + (audioCodec === TRANSCODE_COPY ? 'The audio is copied without re-encoding.' : `The audio is converted to ${audioCodec}.`);
+    } else if (defaultsLoaded && remuxOnly) {
+        hint = sameContainer
+            ? 'Both streams are kept: the file is rewritten as-is with fast start (moves the index to the front for quicker playback start).'
```
lrnselfreliance: **[nit]** For an audio-only remux (`audioOnly`), the remux hint still says "Both streams are kept", which is wrong for a file with only an audio stream.

**Suggestion:** Use stream-neutral copy for audio (e.g. "The audio is kept…") when `audioOnly` / `audioOutput`.

app/src/components/VideoEditMenu.js
lrnselfreliance: Fixed in 9f8da18c: `handleJobFinished` toasts an error on `failed` (with `finished.error` and a note that the original was kept) and an info toast on `cancelled`, before clearing `jobId`. Test: 'a Job that fails after the modal was closed still tells the user'.

modules/videos/transcode.py
lrnselfreliance: Fixed in 9f8da18c: the cleanup is `except BaseException`, so a cancelled encode or tail-decode removes the temp output. Test: `test_transcode_video_file_cancel_removes_tmp` (fake run_command raises CancelledError; original intact, no `.transcode.mp4`).

app/src/components/VideoEditMenu.js
lrnselfreliance: Fixed in 9f8da18c: `getJob` attaches the HTTP status to its error; `useJob` gives up on a 404 at once or after 5 consecutive failures and calls `onLost`, which toasts a warning and clears `jobId` (Transcode and Delete unlock). Tests: 'a 404 means the API forgot the Job', 'repeated failures give up after a few polls', 'a Job the API no longer knows unlocks the menu with a warning'.

app/src/components/VideoEditMenu.js
lrnselfreliance: Trimmed in 9f8da18c: `useJob`, `TranscodeModal`, `VideoEditMenu` and `showJob` comments are now one-line contracts.

wrolpi/jobs.py
lrnselfreliance: Fixed in 09111b9b: the docstring states the atomic claim and the None return; the race history is gone.

app/src/components/VideoEditMenu.js
lrnselfreliance: Fixed in 9f8da18c: the hint reads 'The audio is kept' for audio-only output, and only promises fast start for mp4/m4a. Test: 'an audio file offers audio codecs and containers only' asserts no 'Both streams' text.

PR #713, opened 2026-09-17
files: SQL FileGroup compare, stem column, drop unused discovery path

lrnselfreliance:
## Summary

File refresh compared the filesystem to the DB by loading every group into Python and `stat()`-ing every unchanged file. A two-directory refresh scanned the entire media directory. The old `refresh_discover_paths` path and `/api/files/refresh_progress` were still in the tree but unused.

- **Compare in SQL.** `find` records mtime in a TEMP table (`-printf` on GNU find, one `stat` per file on macOS). The set-diff runs on that connection. Unchanged groups are not `stat()`'d again.
- **Multi-root.** `compare_file_groups([A, B])` only considers those trees. Refreshing A and B no longer deletes a stale row in C.
- **Stem column.** `file_group.stem` is a cache of `split_path_stem_and_suffix`, versioned in `wrolpi_kv`. `ensure_file_group_stems()` runs before every compare so a suffix-rule change cannot delete-and-reinsert tagged files. Index is `(directory, stem)` non-unique; UNIQUE waits until after a full refresh on a real library.
- **Dead path.** Deletes `refresh_discover_paths`, `/api/files/refresh_progress`, and `shared_ctx.refresh`. Discovery tests now go through FileWorker. `file_group.idempotency` is no longer written (column kept: `file_group` is the FTS content table). `directory.idempotency` is unchanged.
- **Modelers.** Video, doc, and zim use `(domain_row IS NULL OR indexed != True)`. Archive still uses `indexed != True` only because it shares `text/html` with generic HTML. `find_indexer` uses longest prefix. The indexing state machine is documented on `apply_indexers`.

## Tests

- `test_compare_file_groups_multiple_roots_does_not_touch_outside`
- `test_refresh_two_directories_does_not_delete_third`
- `test_ensure_file_group_stems_recomputes_on_version_change`
- `test_find_indexer_longest_prefix`
- Discovery tests ported from `refresh_discover_paths` onto FileWorker

`main.py db check` is clean. Related suites (files worker/lib/api, docs modeler, zim modeler, events, OpenAPI fixtures) passed.

## Performance

`test_refresh_many_files` (10k `.txt` files, this Mac):

| | Before | After |
|---|---:|---:|
| First refresh (all new) | 2.40 s | 2.35 s |
| Second refresh (all unchanged) | 0.47 s | 0.36 s |
| Compare phase of second pass | 0.25 s | 0.13 s |

First pass is upsert+index dominated and stayed flat. Second pass is the no-op compare.

## Follow-ups (not this PR)

- UNIQUE `(directory, stem)` after a full refresh on a real library
- Drop `file_group.idempotency` (needs a careful FTS-safe table rebuild)
- Modeler protocol / FileWorker split


wrolpi/files/models.py
```diff
@@ -96,8 +98,14 @@ class FileGroup(ModelHelper, Base):
...
-    idempotency = Column(TZDateTime)  # used to track which files need to be deleted during refresh
-    indexed = Column(Boolean, default=lambda: False)  # wrolpi.files.lib.apply_indexers
+    # Leftover from the idempotency-stamp refresh.  FileWorker no longer writes this.
+    # directory.idempotency is still used.  Do not drop: file_group is the FTS content table.
+    idempotency = Column(TZDateTime)
+    indexed = Column(Boolean, default=lambda: False)  # see indexing state machine on apply_indexers
+    # Cached output of split_path_stem_and_suffix(primary_path).  Identity of a group is
+    # (directory, stem); primary_path is the preferred file inside the group.  Recomputed on
+    # upsert and when STEM_ALGORITHM_VERSION changes (see ensure_file_group_stems).
+    stem = Column(String)
```
lrnselfreliance, line 108: **Reorganize leaves `stem` stale (HIGH)**

`_bulk_update_file_groups_reorganize` in `wrolpi/files/lib.py` (the `UPDATE file_group SET directory, primary_path, files, data` around line 1632) renames files to a new destination stem but never writes this new `stem` column. Before this PR the compare derived the stem from the paths at scan time, so it could not go stale. Now the compare joins on the cached column.

Scenario: reorganize a channel so `old title.mp4` becomes `2024-01-01 New Title.mp4`, then refresh that directory. The DB row has `(dir, 'old title')` with no filesystem match, so it is classified deleted. The filesystem has `(dir, '2024-01-01 New Title')` with no DB match, so it is classified new. The FileGroup and its Video row are dropped and recreated, losing tags, `viewed`, and data. The same stale value also feeds `_refresh_files_directly` via `fg.stem or ...`.

Fix: add `stem = ?` to that UPDATE, computed with `split_path_stem_and_suffix(update['primary_path'])`. Please add a test that reorganizes a channel, refreshes, and asserts the FileGroup id and tags survive. The transcode retarget path and the worker move keep the same stem and look fine.

wrolpi/files/worker.py
```diff
@@ -260,18 +368,56 @@ async def find_directories(directories: list[Path]) -> set[Path]:
...
+        proc = await asyncio.create_subprocess_exec(
+            'find', *find_args, '-printf', '%h\t%f\t%T@\n',
+            stdout=asyncio.subprocess.PIPE,
+        )
+        try:
+            async for line in proc.stdout:
+                decoded = line.decode().rstrip('\n')
+                if not decoded:
+                    continue
+                directory, filename, mtime_s = decoded.split('\t', 2)
```
lrnselfreliance: **Tab or newline in a filename aborts the whole refresh (MEDIUM)**

`find -printf '%h\t%f\t%T@\n'` is parsed by splitting on tab. A filename containing a tab makes the mtime field fail `float()`. A newline produces a record with fewer than three fields and the unpack raises. Either exception escapes the generator and fails every global refresh on Debian, Pi, and Docker until that one file is renamed. The old `-print` path only produced a bogus row for such files.

Suggest NUL separators (`-printf '%h\0%f\0%T@\0'`, split on `\0`), or at minimum catch the parse error, log a warning with the raw record, and skip it.

wrolpi/files/lib.py
```diff
@@ -810,105 +818,88 @@ def remove_files_in_ignored_directories(files: List[pathlib.Path]) -> List[pathl
...
+    so an algorithm change without this backfill reports the same files as deleted+new and
+    drops tags, viewed timestamps, and domain rows.  Must run before any compare.
+    """
+    stored = get_stem_algorithm_version()
+    force = stored != STEM_ALGORITHM_VERSION
+    with get_db_curs() as curs:
+        if force:
+            curs.execute('SELECT id, primary_path FROM file_group')
+        else:
+            curs.execute("SELECT id, primary_path FROM file_group WHERE stem IS NULL OR stem = ''")
```
lrnselfreliance: **`ensure_file_group_stems()` is a blocking full-table scan on every refresh (LOW)**

This is called synchronously from inside the async `compare_file_groups` (worker.py ~506) on every refresh, including single-directory ones. The `WHERE stem IS NULL OR stem = ''` filter cannot use the new `(directory, stem)` index, so it is a full `file_group` scan each time. The force path on first refresh after upgrade loads every `(id, primary_path)` into Python and runs the UPDATEs while blocking the worker's event loop, which on a large Pi library could be a long stall.

Suggest running it via `asyncio.to_thread`, and gating the NULL scan behind a kv flag once a full backfill has completed.

wrolpi/files/worker.py
```diff
@@ -141,6 +142,113 @@ def _deduplicate_db_groups(
...
+
+def _normalize_roots(roots: Path | list[Path] | None) -> list[Path]:
+    if roots is None:
+        return [get_media_directory()]
+    if isinstance(roots, (str, Path)):
+        return [Path(roots)]
+    return [Path(r) for r in roots]
+
+
+def _directory_in_roots_sql(roots: list[Path], column: str = 'fg.directory') -> Tuple[str, dict]:
```
lrnselfreliance, line 242: **Empty `roots` list yields invalid SQL (LOW)**

With `roots=[]` this returns `()`, which is a syntax error in the `db_canonical` query. `_normalize_roots([])` returns `[]` rather than falling back to the media directory the way `None` does, so `compare_file_groups(roots=[])` raises OperationalError. The FileWorker caller is guarded by `if dir_paths:`, so this is only reachable from direct callers and tests, but the `None` fallback silently not applying to `[]` is a footgun. Suggest treating an empty list the same as `None`.

lrnselfreliance (commented):
Review findings from a pass over the SQL compare and stem column. One blocker, one medium, two low.

Checked and found OK: all four registered modelers accept `progress_callback` so dropping the signature check is safe; longest-prefix `find_indexer` matches current registrations; no remaining users of `/api/files/refresh_progress` or `shared_ctx.refresh`; alembic chain has a single head; `FileGroup.move`, `from_paths`, and the video edit-title path all set `stem`; SQL dedupe ordering matches the old Python dedupe; zim modeler loop terminates via `failed_ids`.

wrolpi/files/models.py
lrnselfreliance: Fixed. `_bulk_update_file_groups_reorganize` now writes `stem` from `split_path_stem_and_suffix(primary_path)`. Tests: `test_bulk_update_file_groups_reorganize_updates_stem` and `test_reorganize_then_refresh_preserves_filegroup_id_and_tags`.

wrolpi/files/worker.py
lrnselfreliance: Fixed. GNU find now uses `-printf '%h\0%f\0%T@\0'` and records are split on NUL, so a tab or newline in a filename is not a field break. Unreadable records are logged and skipped. `test_split_nul_field_triples_allows_tab_and_newline_in_filename`.

wrolpi/files/lib.py
lrnselfreliance: Fixed. After a successful backfill the kv version matches `STEM_ALGORITHM_VERSION` and later refreshes skip the table scan. `ensure_file_group_stems` runs in `asyncio.to_thread` so it does not block the worker loop. Upsert, from_paths, move, and reorganize keep stem current.

wrolpi/files/worker.py
lrnselfreliance: Fixed. `_normalize_roots([])` falls back to the media directory, same as `None`. `test_compare_file_groups_empty_roots_is_global`.

PR #715, opened 2026-09-18
files: unique (directory, stem) after collapsing duplicate FileGroups

lrnselfreliance:
## Summary

Follow-up to #713. That PR stored `file_group.stem` with a non-unique index so compare could join on it. This release makes `(directory, stem)` the real unique identity.

- **Migration** `2026_09_18_1200`: backfill NULL stems, collapse leftover duplicates (prefer a modeled row, then tags, then non-empty files, then highest id), move tags and collection membership onto the winner, then `CREATE UNIQUE INDEX file_group_directory_stem_idx`.
- **Upsert / refresh order.** Sidecar FileGroups are deleted **before** INSERT so a new primary cannot collide. Refresh deletes marked-deleted groups before upserting new/modified ones.
- **`from_paths`** looks up by `(directory, stem)` as well as `primary_path`, so adding a sidecar attaches to the existing group.
- **NoPrimaryFile groups** (e.g. `foo.txt` + `foo.zip`) are one FileGroup with a stable primary, matching compare.

## Tests

- `test_file_group_directory_stem_is_unique`
- `test_dedupe_file_groups_merges_tags_onto_winner`
- `test_dedupe_file_groups_prefers_tagged_row`
- `test_unique_stem_migration_collapses_duplicates` (alembic 713 → head)
- `test_refresh_files_no_groups` now expects one group

`main.py db check` is clean.

## Follow-ups (not this PR)

- Drop unused `file_group.idempotency` (FTS content table rebuild)
- Modeler protocol / FileWorker split


lrnselfreliance (commented):
Review findings for the unique `(directory, stem)` index. One blocker, one medium, one low. The first two were reproduced with throwaway tests.

Checked and found OK: the migration from a pre-#713 revision with duplicates and NULL stems passes in one alembic run; `tag_file` has a `(tag_id, file_group_id)` UNIQUE so `INSERT OR IGNORE` is correct; FK cascades are on; the `from_paths` stem lookup handles its own recursion via `unique_by_predicate` and the merge self-skip; `_await_db` and `_stem_backfill_lock` look right.

wrolpi/files/lib.py
```diff
@@ -720,17 +725,17 @@ def _upsert_files(files: List[pathlib.Path],
...
-                        mimetype = get_mimetype(primary_path)
-                        modification_datetime = from_timestamp(primary_path.stat().st_mtime)
-                        size = primary_path.stat().st_size
-                        files_json = json.dumps(_paths_to_files_dict([primary_path]))
-                        stem, suffix = split_path_stem_and_suffix(primary_path)
-                        suffix = (suffix or '').lower() or None
-                        values[primary_path] = (modification_datetime, mimetype, size, files_json, suffix, stem)
+                    # Same stem still means one FileGroup.  Pick a stable primary so
+                    # (directory, stem) stays unique even when nothing is modelable.
+                    primary_path = sorted(group)[0]
```
lrnselfreliance: **Primary-path choice disagrees with `_upsert_file_groups`; tags and collection membership are lost (HIGH)**

This now picks `sorted(group)[0]`. The pre-rewrite in `FileWorker._upsert_file_groups` (`worker.py` ~1623) still uses `new_paths[0]`, where `new_paths` is built from `diff.fs_files`, a `set`. When the two disagree, the existing row with its tags and `collection_item`s lands in `non_primary_files` and is DELETEd before the INSERT, then a fresh untagged row is inserted.

Reproduced: 20 tagged `nN.txt` groups, add `nN.csv` beside each, refresh. 5 of 20 groups came back with no tag. On master the same scenario kept the tag because NoPrimaryFile made one group per file.

Fix: use the same deterministic choice in both places, ideally one shared helper that both call. A test for "add a sidecar to a tagged unmodeled group, refresh, tag survives" would lock this in.

wrolpi/files/models.py
```diff
@@ -83,8 +83,9 @@ class FileGroup(ModelHelper, Base):
         Index('file_group_size_ix', 'size'),
         Index('file_group_url_idx', 'url'),
         Index('file_group_viewed_idx', 'viewed'),
-        # Speeds compare joins.  Not UNIQUE: duplicates are collapsed at refresh time.
-        Index('file_group_directory_stem_idx', 'directory', 'stem'),
+        # Identity of a FileGroup.  Duplicates are collapsed by
+        # dedupe_file_groups_by_directory_stem before this unique index is created.
+        Index('file_group_directory_stem_idx', 'directory', 'stem', unique=True),
```
lrnselfreliance, line 88: **`FileGroup.move` into an existing `(directory, stem)` raises `IntegrityError` after files are renamed on disk (MEDIUM)**

`move()` only checks `new_path.exists()` per suffix, then `shutil.move`s the files, then sets `directory` and `stem`. With this unique index the commit fails with `UNIQUE constraint failed: file_group.directory, file_group.stem`, the session rolls back, and disk and DB disagree.

Reproduced via `lib.rename_file(bar.mp4 -> foo.mp4)` with `foo.jpg` already its own FileGroup. Disk ended up as `foo.jpg, foo.mp4, foo.ffprobe.json`; DB still had `bar.mp4` and `foo.jpg` as separate rows. The same `move()` is used by `rename_file` and by the video download dedupe in `modules/videos/models.py`.

Fix: before any `shutil.move`, look up the `(new_directory, new_stem)` group and merge into it, or raise `FileExistsError` before touching disk.

wrolpi/files/worker.py
```diff
@@ -1487,10 +1506,11 @@ async def handle_refresh(self, task: FileTask):
                 logger.info(f'Refreshing {len(file_paths)} files directly')
                 file_result = await self._refresh_files_directly(file_paths, expand_stems=task.expand_stems)
 
-                # Process file results
+                # Delete first so a new primary_path cannot collide on (directory, stem)
+                # with a duplicate FileGroup that compare already marked deleted.
                 self._cleanup_modified_models(file_result.modified)
-                await self._upsert_file_groups(file_result.new + file_result.modified)
                 await self._delete_file_groups(file_result.deleted)
+                await self._upsert_file_groups(file_result.new + file_result.modified)
```
lrnselfreliance, line 1513: **Third caller of `_refresh_files_directly` still upserts before deleting (LOW)**

This reorder (delete, then upsert) is applied here and in the directory branch below, so a deleted duplicate cannot collide with a new primary on the unique index. The synchronous `refresh_files_directly` (~line 1276), used by `lib.delete` and upload, keeps the old upsert-then-delete order. Any row the compare marks deleted that shares `(directory, stem)` with a new row will raise there. Same three-line reorder needed.

wrolpi/files/lib.py
lrnselfreliance: Fixed. `choose_primary_file` is the shared helper: `get_primary_file` when there is a modeled primary, otherwise `sorted(paths)[0]`. Both `_upsert_files` and `_upsert_file_groups` use it. Test: `test_refresh_keeps_tag_when_sidecar_added`.

wrolpi/files/models.py
lrnselfreliance: Fixed. `FileGroup.move` looks up the destination `(directory, stem)` group, merges it (tags included), and flushes the DELETE before we take that stem so the unique index is not violated after files are already renamed. Same-path files after the merge are not `shutil.move`d onto themselves. Test: `test_rename_into_existing_stem_merges`.

wrolpi/files/worker.py
lrnselfreliance: Fixed. `refresh_files_directly` now deletes then upserts, same as the two `handle_refresh` branches.

